PR #29, opened 2021-11-25
Quality improvements

### LINKED ISSUE #6: Cache datasets across experiments and language pairs
Right now datasets are redownloaded for every new experiment.

The data should be cached in a common folder to be reused across multiple experiments and reversed language pairs.

### LINKED ISSUE #9: Replace absolute paths with relative ones
Feedback:

>Given the name WORKDIR , I'd expect it means the directory where the scripts are doing work like training or whatever.  But it appears to mean I have to tell the scripts where they are.  Why not the best practice of scripts finding their dependencies relative to themselves?  If you want it rooted at the top of the scripts repo, then just tell each script at the top how deep it is.

It seems it is not the best practice to expect all scripts to know their location in terms of an absolute path, even though it is easier to debug, so we can replace this approach with relative paths.
Also `WORKDIR` name is confusing and can be renamed to something else.

### LINKED ISSUE #15: Add dataset specific fixes
Check https://github.com/ZJaume/clean

There should be a possibility to add scripts that fix a dataset based on its name. It should be easily extensible to new datasets.

### LINKED ISSUE #18: Try other teacher hyperparameters
An example of teacher hyperparameters from Kenneth (requires the latest Marian):

```
devices:
 - 0
 - 1
 - 2
 - 3

workspace: 12000
log: train.log
model: model.npz

train-sets:
  - data/train.en-de.en.gz
  - data/train.en-de.de.gz
seed: 1131
vocabs:
  - data/vocab.ende.spm
  - data/vocab.ende.spm
task: transformer-big
shuffle-in-ram: true
# Validation set options
valid-sets:
  - data/valid.en-de.en
  - data/valid.en-de.de
valid-freq: 3000
disp-freq: 100
valid-metrics:
  - ce-mean-words
  - bleu-detok
valid-mini-batch: 8
valid-max-length: 300
valid-translation-output: wmt19.valid.output
keep-best: true
valid-log: valid.log
overwrite: true
learn-rate: 0.0003 # Turn this down if you get a diverged model, maybe 0.0001
optimizer-delay: 2 # Roughly GPU devices * optimizer-delay = 8, but keep as an integer
```

If they work better, we should update the teacher config

### LINKED ISSUE #19: Add more languages to corpus cleaning script
we can use a list from wiki to add valid characters for other languages: https://en.wikipedia.org/wiki/Wikipedia:Language_recognition_chart

### LINKED ISSUE #22: Add dataset specific bicleaner thresholds
There should be a default value for bicleaner threshold and the ability to override it for specific datasets.

### LINKED ISSUE #23: Fine tune teacher on parallel data only
When using backtranslations + parallel data for teacher training, continue training on parallel data only.

### LINKED ISSUE #25: Remove punctuation normalizaiton
Ulrich:
> https://github.com/mozilla/firefox-translations-training/blob/main/pipeline/clean/clean-corpus.sh uses the normalize-punctuation perl script from Moses. I would advise against using that, because using it forces us to do the same hacky preprocessing for actual translation (otherwise the translator may encounter characters it has never seen). On the target side I'm fine with standardization of quotation marks as part of data preparation (but these should match language-specific standards, not use Ascii quotes for everything). I'm not sure though, if it is worth the effort in practice. Post-translation punctuation normalization is one of the usual tricks for improving BLEU scores in shared tasks / competitions, while most humans probably aren't bothered too much if quotation marks aren't quite right.

Kenenth:
>Just remove
perl ${CLEAN_TOOLS}/normalize-punctuation.perl -l ${lng}" |

### LINKED ISSUE #27: Evaluate ensemble of teachers

### LINKED ISSUE #28: Add chrF evaluation metric

eu9ene:
1. Update teacher hyperparameters ( provided by Kenneth). fixes  #18
2. Use chrf metric for the best model (https://discourse.translatelocally.com/t/marian-configuration-to-use/24)
3. Update SacreBLEU and add chrF metric to evaluation. fixes #28 
4. Add evaluation of each model and the whole ensemble of teachers. fixes  #27
5.  Early stop based on ce-mean-words instead of bleu-detok Q: Not sure about this one, is it ok to stop basd on ce-mean-words but then use best chrf model?
6. Continue teacher training on parallel data only (train on augmented data for 5 epochs first). fixes #23 Q: how many epochs should we use? What about low resource languages, should we finetune on a limited corpus?
5. Do cleaning per dataset
6. Add per-dataset fixes from https://github.com/ZJaume/clean/tree/master/fixes. fixes  #15
7. Use bicleaner per dataset with customizable thresholds. fixes  #22 Q: I added some random thresholds based on Kenneth's comments, how to tune them?
8. Remove punctuation normalization based on Ulrich's input. fixes  #25
9. Add alphabets for more languages in the cleaning scripts. fixes  #19
10. Replace absolute paths with relative ones. fixes  #9
11. Add Snakemake cross-workflow caching. Caching works, but apparently, there is a bug in Snakemake, it doesn't recognize symlinks after caching. Disabled for now. fixes  #6


XapaJIaMnu:
Re, Q5: The model optimises towards CE, as opposed to towards BLEU. Stopping training based on the CE stalling criteria is going to mean that you stop training when the model starts to diverge. BLEU/CHFR shouldn't be used as a stopping criteria because they don't correspond to the current state of training.

Re 1). The transformer-big configuration should not be used as a hard and fast rule. In general if you have less than 5M sentence pairs you are unlikely to see any benefit from transformer-big. Use transformer base instead.

configs/config.test.yml
```diff
@@ -18,12 +19,27 @@ experiment:
...
+  best-model: chrf
+
+  bicleaner:
+    default-threshold: 0.5
+    dataset-thresholds:
+      mtdata_neulab_tedtalksv1_train: 0.6
+
 training:
-  after-epochs: 1
+  s2s:
```
XapaJIaMnu, line 33: Why are things stopping to train after just one epoch? This is inadequate for vast majority of language pairs. Is this just for testing purposes, as opposed to actually configuration used for training?

configs/config.prod.yml
```diff
@@ -24,6 +25,27 @@ experiment:
...
+      mtdata_WikiMatrix_v1: 0.7
+      mtdata_wiki_titles_v2: 0.7
+      mtdata_wmt13_commoncrawl: 0.7
+
+training:
+  s2s:
+    after-epochs: 10
+  # limit number of epochs on augmented dataset (high resource languages)
+  teacher-all:
+    after-epochs: 5
```
XapaJIaMnu, line 48: This is not adequate for lower resource languages, that need quite a few more epochs to converge.

configs/config.test.yml
eu9ene: This is a test config. Its purpose is to run it all end to end as quickly as possible to catch bugs. I'll add comments to make it clear.

configs/config.prod.yml
eu9ene: Yes, that's why I moved it to the main config, to be able to adjust based on low resource/high resource. For low resource, I would just remove this and train until early stopping. I can add comments about this.

Does fine-tuning on very limited parallel data even make sense for low resource?

Is 5 epochs a good number for high resource languages?

eu9ene:
> Re 1). The transformer-big configuration should not be used as a hard and fast rule. In general if you have less than 5M sentence pairs you are unlikely to see any benefit from transformer-big. Use transformer base instead.

This is a very valuable insight. I haven't started with low resource languages yet. Quite a few things must be different there. I'll add transformer base configuration and create another main config for low resource languages as an example.

configs/config.prod.yml
XapaJIaMnu: I would say that there isn't a hard and fast rule about the number of epochs. It really depends on how varied your data is. Eg if your data is from the same domain and very similar, maybe it will be done in 5 epochs. If the data is diverse, then maybe not. I prefer for humans to determine the patience. Is a researcher able to login and `ctrl+c` this step and continue to the next step?

XapaJIaMnu:
> > Re 1). The transformer-big configuration should not be used as a hard and fast rule. In general if you have less than 5M sentence pairs you are unlikely to see any benefit from transformer-big. Use transformer base instead.
> 
> This is a very valuable insight. I haven't started with low resource languages yet. Quite a few things must be different there. I'll add transformer base configuration and create another main config for low resource languages as an example.

just to be clear that is `task: transformer-base`  https://github.com/marian-nmt/marian-dev/blob/master/src/common/aliases.cpp#L75

configs/config.prod.yml
eu9ene: The logic here is that we preconfigure a number of epochs for training on a dataset augmented with back translations then continue training on parallel data only with early stopping (advice from Kenneth: "train on backtranslated first for several epochs, then continue on parallel".).  If we want to pretrain on augmented dataset to convergence, it also should use early stopping.

I don't think it's easy to skip things with snakemake. If a step fails it deletes the expected output to prevent corrupted data. It is possible to stop the workflow, fix some things and continue from where it stopped. (harder on Slurm because it requires cancelling Slurm jobs manually). It is also possible to run snakemake with a specific target to execute only a part of the pipeline.

kpu: The pretraining on backtranslation is controlled by epochs and the continued training on parallel is controlled by patience then?

eu9ene: Yes and this is controlled in the config, so if we remove `after-epochs: 5`, it will be also trained until early stopped (10 epochs by default). I assumed a user can remove it for low resource but keep fixed for high resource language. I can remove it if you think it makes more sense to train on augmented data until convergence.

ugermann: As Nick points out, reasonable settings of epochs depend on 1. the total amount of training data available, and 2. the diversity of the training data. May I suggest that instead of `after-epochs` (which is a deprecated parameter anyway; the preferred way of specifying a stopping point is `after`, which allows us to specify the number of updates or the amount of training data in terms of target labels rather than the number of iterations over the available data. See https://marian-nmt.github.io/docs/cmd/marian/). And no, I don't know what a reasonable value is. This would require experimentation (if such experimentation has been done already, I'm not aware of it, but we have plenty of training logs where we could look at when the learning curve starts to flatten out.

configs/config.test.yml
ugermann: As mentioned in an earlier comment. I'd recommend using `after` instead of `after-epoch` for testing. Going through an entire epoch can take very long if you have lots of data to train on.

pipeline/cefilter/ce-filter.sh
```diff
@@ -29,7 +31,7 @@ test -s "${tmp}/corpus.${SRC}" || pigz -dc "${corpus_prefix}.${SRC}.gz" >"${tmp}
 echo "### Normalizing scores"
 test -s "${tmp}/scores.nrm.txt" ||
   paste "${scores}" "${tmp}/corpus.${TRG}" |
-  parallel --no-notice --pipe -k -j "${threads}" --block 50M "python pipeline/cefilter/normalize-scores.py" |
+  parallel --no-notice --pipe -k -j "${threads}" --block 50M "python normalize-scores.py" |
   cut -f1 >"${tmp}/scores.nrm.txt"
```
ugermann, line 35: Why is this step necessary? `marian-scorer` can do normalisation (run with `-n`). Also, using `str.split()` in `normalize-scores.py` will not properly normalize the sentence-level scores produced by marian-scorer, as that score is based on subword tokenizations.

pipeline/bicleaner/bicleaner.sh
```diff
@@ -15,36 +15,35 @@ corpus_prefix=$1
...
+  ${cmd} --scol 1 --tcol 1 --processes "${threads}"  - - "${pack_dir}"/*.yaml |
   awk -v threshold=${bicleaner_threshold} '{if ($3>threshold) {print $0}}' |
-  pigz >"${tmp_dir}/best.gz"
+  pigz >"${output_prefix}.best.gz"
 
 echo "### Writing output corpus"
-pigz -dc "${tmp_dir}/best.gz" | cut -f1 | pigz >"${output_prefix}.${SRC}.gz"
-pigz -dc "${tmp_dir}/best.gz" | cut -f2 | pigz >"${output_prefix}.${TRG}.gz"
+pigz -dc "${output_prefix}.best.gz" | cut -f1 | pigz >"${output_prefix}.${SRC}.gz"
+pigz -dc "${output_prefix}.best.gz" | cut -f2 | pigz >"${output_prefix}.${TRG}.gz"
```
ugermann, line 44: This can be accomplished in half the time up by using `tee`:
```
pigz -dc "${output_prefix}.best.gz" \
| tee >(cut -f1 | pigz >"${output_prefix}.${SRC}.gz") \
| cut -f2 | pigz >"${output_prefix}.${TRG}.gz"
```

pipeline/cefilter/ce-filter.sh
```diff
@@ -29,7 +31,7 @@ test -s "${tmp}/corpus.${SRC}" || pigz -dc "${corpus_prefix}.${SRC}.gz" >"${tmp}
 echo "### Normalizing scores"
 test -s "${tmp}/scores.nrm.txt" ||
   paste "${scores}" "${tmp}/corpus.${TRG}" |
-  parallel --no-notice --pipe -k -j "${threads}" --block 50M "python pipeline/cefilter/normalize-scores.py" |
+  parallel --no-notice --pipe -k -j "${threads}" --block 50M "python normalize-scores.py" |
```
ugermann, line 34: Why is this step necessary? marian-scorer can do normalisation (run with -n). Also, using str.split() in normalize-scores.py will not properly normalize the sentence-level scores produced by marian-scorer, as that score is based on subword tokenizations.

eu9ene: It was taken from https://github.com/browsermt/students/tree/master/train-student/ce-filter, so we should ask its author. I can replace it with normalization by marian.

XapaJIaMnu: Likely an issue at the time of writing those scripts. Modern marian has picked up quite a few features.

configs/config.test.yml
eu9ene: updated configs to use `after`

configs/config.prod.yml
eu9ene: I moved these settings to `pipeline/train/configs` and updated them to use `after`. It will be easier to modify everything in one place.

pipeline/bicleaner/bicleaner.sh
eu9ene: fixed

pipeline/cefilter/ce-filter.sh
eu9ene: fixed

eu9ene:
> > > Re 1). The transformer-big configuration should not be used as a hard and fast rule. In general if you have less than 5M sentence pairs you are unlikely to see any benefit from transformer-big. Use transformer base instead.
> > 
> > 
> > This is a very valuable insight. I haven't started with low resource languages yet. Quite a few things must be different there. I'll add transformer base configuration and create another main config for low resource languages as an example.
> 
> just to be clear that is `task: transformer-base` https://github.com/marian-nmt/marian-dev/blob/master/src/common/aliases.cpp#L75

I added comments in `pipeline/train/configs/model/teacher.yml`. So far it's pretty manual to reconfigure everything for another model. When all this is stabilized and we trained a bunch of models, I can think of better ways how to provide ready-to-use recipes for low-resource languages.

andrenatal:
This is looking great. Thanks @eu9ene !

andrenatal (approved):

PR #41, opened 2021-12-22
Bugfix and optimization

### LINKED ISSUE #30: Version of marian used for training
We should be training with `marian-nmt/marian-dev` HEAD (current head is safe), not with `browsermt/marian-dev`. `browsermt/marian-dev` should only be used for binarizing the intgemm models that run into the browser.

### LINKED ISSUE #31: Change sharding mode in training to improve speed
We should marian-nmt/marian-dev for our training (current HEAD), as it has a bunch of fixes and performance improvements. One of those is a change in the sharding mode: `https://github.com/marian-nmt/marian-dev/blob/master/src/common/config_parser.cpp#L828` achieved by setting: `sharding: local`, as it noticeably improves training speed.

Downside of using this is slightly increased memory consumption. If we have the memory available in the GPU, we should use it.

### LINKED ISSUE #32: find-corpus.py is outdated?
`find-corpus.py` is now located at `./utils/` as opposed to `./pipeline/utils/`, the readme should be updated.

It also can't find a module named `mtdata`:
```bash
$ python ./utils/find-corpus.py en ru mtdata
Traceback (most recent call last):
  File "/home/nikolay/firefox-translations-training/./utils/find-corpus.py", line 36, in <module>
    from mtdata.main import LangPair
ModuleNotFoundError: No module named 'mtdata'
```

Works fine with OPUS though.

### LINKED ISSUE #33: spm vocabulary training could fail on high res languages
We don't necessarily want to train the vocabulary on all corpora, because for high res language the vocabulary step would fail eg:

```bash
tcmalloc: large alloc 1342177280 bytes == 0x559ec7632000 @ 
trainer_interface.cc(137) LOG(INFO) Loaded 17000000 lines
trainer_interface.cc(137) LOG(INFO) Loaded 18000000 lines
trainer_interface.cc(137) LOG(INFO) Loaded 19000000 lines
trainer_interface.cc(137) LOG(INFO) Loaded 20000000 lines
trainer_interface.cc(112) LOG(WARNING) Too many sentences are loaded! (20000000), which may slow down training.
trainer_interface.cc(114) LOG(WARNING) Consider using --input_sentence_size=<size> and --shuffle_input_sentence=true.
trainer_interface.cc(117) LOG(WARNING) They allow to randomly sample <size> sentences from the entire corpus.
trainer_interface.cc(376) LOG(INFO) Loaded all 20000000 sentences
trainer_interface.cc(391) LOG(INFO) Adding meta_piece: </s>
trainer_interface.cc(391) LOG(INFO) Adding meta_piece: <unk>
trainer_interface.cc(396) LOG(INFO) Normalizing sentences...
trainer_interface.cc(457) LOG(INFO) all chars count=3847586460
trainer_interface.cc(468) LOG(INFO) Done: 99.954% characters are covered.
trainer_interface.cc(478) LOG(INFO) Alphabet size=87
trainer_interface.cc(479) LOG(INFO) Final character coverage=0.99954
trainer_interface.cc(511) LOG(INFO) Done! preprocessed 20000000 sentences.
tcmalloc: large alloc 2147483648 bytes == 0x55a075720000 @ 
tcmalloc: large alloc 4294967296 bytes == 0x55a0f5720000 @ 
tcmalloc: large alloc 8589934592 bytes == 0x55a1f5faa000 @ 
tcmalloc: large alloc 17179869184 bytes == 0x55a3f67aa000 @ 
unigram_model_trainer.cc(124) [(array.size()) <= (static_cast<size_t>(std::numeric_limits<node_int_type>::max()))] Input corpus too large, try with train_extremely_large_corpus=true
Program terminated with an unrecoverable error.
```

Furthermore, we don't usually want to train on large dirty corpora (small outlier inputs would be filtered out by noise anyways, but it slows down training unnecessary).

### LINKED ISSUE #34: bicleaner should be optional for some corpora
Currently the bicleaner step in the pipeline is mandatory, but it's not necessarily appropriate for all cases
1) Some corpora such as paracrawl have already been cleaned by bicleaner, so there is no point of going through them again.
2) Some corpora are already quite clean and do not need the full bicleaner treatment.

bicleaner should be an optional step for every corpus.

### LINKED ISSUE #35: deescape-special-chars.perl should be part of the clean step
Some of the corpora on the internet are there from the moses era where we couldn't have any special characters in the phrase tables and exist in a preprocessed manner on Opus and other sources. To fix that we should use https://github.com/moses-smt/mosesdecoder/blob/master/scripts/tokenizer/deescape-special-chars.perl in our pipeline.

### LINKED ISSUE #36: bicleaner-ai crashes with out of GPU memory, when running with more than one jobs per node
bicleaner-ai uses tensorflow which tends to reserve all the memory on a single GPU (even if it doesn't need it). As a result, cleaning several different parallel corpora will result in a crash due to running out of GPU memory (or memory corruption). 

Recommendation: use CUDA_VISIBLE_DEVICES to mask the GPUs and run only number of GPUs jobs, with each job taking a designated GPU.

### LINKED ISSUE #37: Make it possible to assign  non-consecutive GPUs depending on what we have available on the machine
Currently GPUs are assigned by this:

https://github.com/mozilla/firefox-translations-training/blob/a09b0ac7ac80bea8fe3b0da99ad4c7bd39416545/Snakefile#L118

On some non-cluster shared machines, we may want to use a subset of the GPUs on the system for training, which is not possible with the current system. Instead we should be able to pass something like ```GPUS={2, 3, 4, 5}``` if this is what we have free on our system.

### LINKED ISSUE #38: Custom corpus works fine as a dev set but not as a test set.
I was trying to train an en-bg model and I got a failure in the `eval backward` model step.

Since Bulgarian doesn't have an official dataset, I used a custom one. (Here's a snipped of my config):
```yml
...
datasets:
  # parallel corpus
  train:
    - opus_Ubuntu/v14.10
    ...
    - opus_DGT/v2019
    - opus_ELRC_2682/v1
  devtest:
    - custom-corpus_/mnt/hrist0/nbogoych/government.bg
  test:
    - custom-corpus_/mnt/hrist0/nbogoych/government.bg
  # monolingual datasets (ex. paracrawl-mono_paracrawl8, commoncrawl_wmt16, news-crawl_news.2020)
  # to be translated by the teacher model
  mono-src:
    - news-crawl_news.2020
    - news-crawl_news.2019
    ...
```

Snakemake fails at:

```bash
[Wed Dec  1 20:41:15 2021]                                                                                                                                                                                                                                                                           
Error in rule eval_backward:                                                                                                                                                                                                                                                                         
    jobid: 169                                                                                                                                                                                                                                                                                       
    output: /mnt/hrist0/nbogoych/data/models/en-bg/snakemake-en-bg/evaluation/s2s                                                                                                                                                                                                                    
    log: /mnt/hrist0/nbogoych/data/logs/en-bg/snakemake-en-bg/eval_backward.log (check log file(s) for error message)                                                                                                                                                                                
    conda-env: /mnt/hrist0/nbogoych/firefox-translations-training/.snakemake/conda/ba5ec19af3d16034e3cbf600bb7b34dd                                                                                                                                                                                  
    shell:                                                                                                                                                                                                                                                                                           
        bash pipeline/train/eval.sh "/mnt/hrist0/nbogoych/data/models/en-bg/snakemake-en-bg/evaluation/s2s" "/mnt/hrist0/nbogoych/data/data/en-bg/snakemake-en-bg/original/eval" bg en /mnt/hrist0/nbogoych/data/models/en-bg/snakemake-en-bg/s2s/model.npz.best-chrf.npz >> /mnt/hrist0/nbogoych/data/logs/en-bg/snakemake-en-bg/eval_backward.log 2>&1                                                                                                                                                                                                                                                  
        (one of the commands exited with non-zero exit code; note that snakemake uses bash strict mode!)                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                     
Removing output files of failed job eval_backward since they might be corrupted:
```

The log contains:
```bash
+ set -euo pipefail
+ echo '###### Evaluation of a model'
###### Evaluation of a model
+ test -v GPUS
+ test -v MARIAN
+ test -v WORKSPACE
+ eval_dir=/mnt/hrist0/nbogoych/data/models/en-bg/snakemake-en-bg/evaluation/s2s
+ datasets_dir=/mnt/hrist0/nbogoych/data/data/en-bg/snakemake-en-bg/original/eval
+ src=bg
+ trg=en
+ models=("${@:5}")
+ mkdir -p /mnt/hrist0/nbogoych/data/models/en-bg/snakemake-en-bg/evaluation/s2s
+ echo '### Evaluating the model'
### Evaluating the model
+ for src_path in "${datasets_dir}"/*."${src}.gz"
++ basename '/mnt/hrist0/nbogoych/data/data/en-bg/snakemake-en-bg/original/eval/*.bg.gz' .bg.gz
+ prefix='*'
+ echo '### Evaluating * bg-en'
### Evaluating * bg-en
+ pigz -dc '/mnt/hrist0/nbogoych/data/data/en-bg/snakemake-en-bg/original/eval/*.bg.gz'
pigz: skipping: /mnt/hrist0/nbogoych/data/data/en-bg/snakemake-en-bg/original/eval/*.bg.gz does not exist
```

The directory structure:
```bash
/mnt/hrist0/nbogoych/data/data/en-bg/snakemake-en-bg/original/eval$ tree
.
`-- custom-corpus_
    `-- mnt
        `-- hrist0
            `-- nbogoych
                |-- government.bg.bg.gz
                `-- government.bg.en.gz

4 directories, 2 files
```

The devset on the other hand looks differently:

```bash
/mnt/hrist0/nbogoych/data/data/en-bg/snakemake-en-bg/original/devset$ tree
.
|-- custom-corpus_
|   `-- mnt
|       `-- hrist0
|           `-- nbogoych
|               |-- government.bg.bg.gz
|               `-- government.bg.en.gz
`-- merge.enbg.gz

4 directories, 3 files
```

Since i use the same entries for dev and test sets, I can't imagine it's a config file issue?

### LINKED ISSUE #39: Forward (and back) translation should be done in fp16 mode if the GPU supports it.
When producing the necessary data for training the student and if decoding on a recent GPU (2080ti or newer, **NOT a CPU)** we should use float16 mode. Float16 mode gives about 30% improvement in translation speed at basically no loss of BLEU and also allows for much larger mini-batches. This will save significant amount of time and money.

### LINKED ISSUE #40: Allow expert user to configure mini-batch size when decoding
The default mini batch size setting:
https://github.com/eu9ene/bergamot-training/blob/quality/pipeline/translate/decoder.yml
```
normalize: 1.0
word-penalty: 0
mini-batch: 16
mini-batch-words: 2000
maxi-batch: 1000
maxi-batch-sort: src
max-length: 200
max-length-crop: true
beam-size: 8
quiet-translation: True
```
Is not adequate for all models involved. The s2s model used for backtranslation could easily have 512 mini-batch size on a 3090ti and much larger maxi-batch and mini-batch words. That would yield up to **4X** improvement in translation speed, as shown on table 4: http://statmt.org/wmt21/pdf/2021.wmt-1.74.pdf

It is good to have safe default values for non-expert users, but for getting the most of our limited hardware budget, those need to be easily adjustable configuration options. Also the settings for the teacher models and for the backtranslation models would have to be different.

### LINKED ISSUE #42: TRG lang should also be checked whether it obeys the ratio metric.
https://github.com/mozilla/firefox-translations-training/blob/3b3f33bf2581238d325f05015123fc0a026c394e/pipeline/clean/tools/clean_parallel.py#L103

Missing TRG_LANG case here.

### LINKED ISSUE #44: Bicleaner path is hardcoded even if bicleaner is not used, resulting in crashes
https://github.com/mozilla/firefox-translations-training/blob/3b3f33bf2581238d325f05015123fc0a026c394e/Snakefile#L387

If we try to use a language pair without bicleaner, the training will crash because the corpus location is invalid.

eu9ene:
- bugfix
- training and decoding optimization
- evaluation refactoring
- small usability improvements
- moved marian configurations overriding back to configs/

Bicleaner-ai shouldn't fail on a local machine now, but it's pretty hard to parallelize it in local mode (https://github.com/snakemake/snakemake/issues/281). I reserve all provided GPUs now. Can it use multiple GPUs itself?

Decoding with an ensemble of 4 teachers seems to be too slow, so I'm reducing it to 2 models. Does anybody have a quality comparison for ensembles of 4 vs 2 transformer-big models?

fixes #44 
fixes #42
fixes #40 
fixes #39 
fixes #38 
fixes #37
fixes #36 
fixes #35 
fixes #34 
fixes #33 
fixes #32 
fixes #31 
fixes #30


pipeline/clean/tools/clean_parallel.py
```diff
@@ -104,11 +104,21 @@ def clean_parallel(src, trg, src_lang, trg_lang):
...
         if num_alpha / float(src_len) < RATIO_ALPHA_WORDS:
-            return "RATIO_ALPHA"
+            return "RATIO_ALPHA_SRC"
 
         char_alpha = len(re.findall(CHARS[src_lang], src, re.IGNORECASE))
         if char_alpha / float(len(src.replace(' ', ''))) < RATIO_ALPHA_CHARS:
-            return "RATIO_CHARS"
+            return "RATIO_CHARS_SRC"
+
+    if trg_lang in CHARS:
```
XapaJIaMnu, line 113: Why is it this filtering only applied on `trg_lang`  and not on `src_lang` as well?

pipeline/train/spm-vocab.sh
```diff
@@ -11,14 +11,19 @@ test -v MARIAN
...
 vocab_dir=$(dirname "${vocab_output}")
 mkdir -p "${vocab_dir}"
 
-pigz -dc "${corpus_src}" | shuf -n 10000000 >"${vocab_dir}/data.txt"
-pigz -dc "${corpus_trg}" | shuf -n 10000000 >>"${vocab_dir}/data.txt"
+pigz -dc "${corpus_src}" >"${vocab_dir}/data.txt"
+pigz -dc "${corpus_trg}" >>"${vocab_dir}/data.txt"
 
 "${MARIAN}/spm_train" --bos_id=-1 --eos_id=0 --unk_id=1 --user_defined_symbols="" \
-  --input="${vocab_dir}/data.txt" --model_prefix="${vocab_dir}/vocab" --vocab_size=32000
```
XapaJIaMnu, line 22: --input takes multiple files, no need to concat them first.

XapaJIaMnu (approved):
Looks good to me. I haven't tested it but good job! Thank you for your continued work.

I have just left some minor comments and suggestions.

pipeline/train/spm-vocab.sh
XapaJIaMnu: From github:
```--input: one-sentence-per-line raw corpus file. No need to run tokenizer, normalizer or preprocessor. By default, SentencePiece normalizes the input with Unicode NFKC. You can pass a comma-separated list of files.```
Not sure it works with gzipped corpora as well.

pipeline/clean/tools/clean_parallel.py
eu9ene: I'm not sure what you mean, all filters are applied to both src and trg languages

XapaJIaMnu: My bad, I was confused.

pipeline/train/spm-vocab.sh
eu9ene: ok, fixed, but it doesn't work with compressed files

PR #50, opened 2022-01-14
Make patience metric be chrf

kpu:
For high resource languages, ce-mean-words will keep going down for a long time, even if quality isn't improving.  The first-listed metric determines which one is used for early stopping criteria.  This changes the stopping criteria metric to chrf.


eu9ene (approved):
Thank you for this correction! It was changed based on a new teacher training recipe, but I also noticed that my teachers are training forever now because of this.

PR #51, opened 2022-01-14
Fine-tuning bug fix

### LINKED ISSUE #49: Teacher does not continue training if training on augmented data was early stopped
If `ce-mean-words` was stalled 10 times on `teacher_all` step , `continue_teacher` step fails with error, it stops without training. It's fine to early stop the pertaining, however, continuation should load only weights of the pre-trained model but not optimizer parameters.

### LINKED ISSUE #56: Wrong tcol when cleaning with Bicleaner
Shouldn't [this](https://github.com/mozilla/firefox-translations-training/blob/ef8928b45417a49138e2a4aad97a82282ea448f9/pipeline/bicleaner/bicleaner.sh#L39) line say `--scol 1 --tcol 2`?

eu9ene:
- It causes fewer problems when teacher training on a parallel dataset happens in a separate directory. Model weights are initialized using `--pretrained-model` Marian parameter.
fixes #49 

- Fixes a new bug with student fine-tuning, there was no weights initialization (it was lost during refactoring).
- Fixed usage of a pretrained backward model + vocab
- fixes #56

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -92,8 +92,8 @@ align_dir = f"{data_dir}/alignment"
 
 # models
 models_dir = f"{data_root_dir}/models/{src}-{trg}/{experiment}"
-teacher_all_dir = f"{models_dir}/teacher-all"
-teacher_parallel_dir = f"{models_dir}/teacher-parallel"
+teacher_all_dir = f"{models_dir}/teacher"
+teacher_parallel_dir = f"{models_dir}/teacher-finetuned"
 student_dir = f"{models_dir}/student"
 student_finetuned_dir = f"{models_dir}/student-finetuned"
 speed_dir = f"{models_dir}/speed"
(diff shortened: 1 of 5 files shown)


Snakefile
```diff
@@ -91,14 +92,16 @@ align_dir = f"{data_dir}/alignment"
 
 # models
 models_dir = f"{data_root_dir}/models/{src}-{trg}/{experiment}"
-teacher_dir = f"{models_dir}/teacher"
+teacher_all_dir = f"{models_dir}/teacher-all"
+teacher_parallel_dir = f"{models_dir}/teacher-parallel"
```
XapaJIaMnu, line 96: From reading the source, I don't understand what `teacher_parallel_dir` should contain. What is a parallel teacher model?

XapaJIaMnu (approved):
As far as I understand this wouldn't affect the behaviour when training crashed and was resumed, correct? That would still continue with loading the optimiser parameters. (TBH I haven't tested if this case even works).

eu9ene:
> As far as I understand this wouldn't affect the behaviour when training crashed and was resumed, correct? That would still continue with loading the optimiser parameters. (TBH I haven't tested if this case even works).

This will work because I removed protection from the output file `model.npz.best-chrf.npz`. It means snakemake will delete it when the job is stopped or crushed, so it will rerun training next time to get this file. Since it will happen in the same directory and `model.npz` and optimizer progress files are there, it will continue training. 

However, this is an irregular situation and not desirable. The pipeline is designed to work end to end without interruptions.

Snakefile
eu9ene: Teacher all - the model is trained on all available data.
Teacher parallel - optional model to fine-tune on parallel data only if the data was augmented with back translations.

Will it be easier to understand if I rename them to `teacher` and `teacher-finetuned`?

XapaJIaMnu: Yes that would be easier to understand, but this is a very minor point.

PR #60, opened 2022-01-26
Fix to use with the latest mtdata version

### LINKED ISSUE #43: Update mtdata version
It allows using paracrawl v9

XapaJIaMnu:
The latest mtdata changes corpus names, the funcitonality of the function and rearranges a number of internal functions. I hope this will work now. I also filed a few bug reports @mtdata
This should fix #43 though I have not tested whether the corpus fetcher actually works.

### DIFF
--- a/configs/config.prod.yml
+++ b/configs/config.prod.yml
@@ -41,14 +41,12 @@ experiment:
     default-threshold: 0.5
     dataset-thresholds:
       opus_CCAligned/v1: 0.7
-      opus_WikiMatrix/v1: 0.7
       opus_OpenSubtitles/v2018: 0.8
       opus_bible-uedin/v1: 0.7
-      mtdata_cc_aligned: 0.7
-      mtdata_wiki_titles_v1: 0.7
-      mtdata_WikiMatrix_v1: 0.7
-      mtdata_wiki_titles_v2: 0.7
-      mtdata_wmt13_commoncrawl: 0.7
+      mtdata_Statmt-wiki_titles-1-rus-eng: 0.7
+      mtdata_Facebook-wikimatrix-1-eng-rus: 0.7
+      mtdata_Statmt-wiki_titles-2-rus-eng: 0.7
+      mtdata_Statmt-commoncrawl_wmt13-1-rus-eng: 0.7
       # 0 = skip filtering
       opus_ParaCrawl/v8: 0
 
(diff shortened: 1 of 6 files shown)


eu9ene:
I also tried to update it but encountered the problem with new naming that includes language pair now:
```  
    - mtdata_Statmt-wiki_titles-2-rus-eng
    - mtdata_Tilde-airbaltic-1-eng-rus
    - mtdata_Tilde-czechtourism-1-eng-rus
    - mtdata_Tilde-worldbank-1-eng-rus
    - mtdata_UN-un_dev-1-eng-rus
    - mtdata_UN-un_test-1-eng-rus
   ```
our per-corpus fixes won't work for mtdata because the name of every fix corresponds to snakemake wildcard - full dataset name (it will be `mtdata_UN-un_test-1-eng-rus`, language pair specific)

If we're ok with it, the following components should be updated:

1. conda environments `base` and `corpus` should be updated with the latest version of mtdata.
2.  fetcher. Here is what I changed in `data/importers/corpus/mtdata.sh`:
```
mtdata get -l "${src}-${trg}" -tr "${dataset}" -o "${tmp}" --compress

mv "${tmp}/train-parts/${dataset}.${src_iso}.gz" "${output_prefix}.${src}.gz"
mv "${tmp}/train-parts/${dataset}.${trg_iso}.gz" "${output_prefix}.${trg}.gz"
```

3. config files in `configs`

It's a good idea to run `make test` after smth is fixed or at least `make test TARGET=...` to test it up to a specific step.

eu9ene:
please update the configs too

XapaJIaMnu:
@eu9ene I updated the configs and the mtdata fetcher. At the moment there is the issue that we too many duplicated datasets from the Opus fetcher and the MTdata fetcher. Not sure if mtdata is a complete superset of Opus, or not. One way to solve that problem is to checksum all the downloaded files after they are downloaded and remove the duplicates maybe?

eu9ene:
Thanks! I would not overcomplicate this and leave deduplication to a user. I usually take whatever is available on OPUS as a more reliable resource and try to complement it with datasets from mtdata that are missing on OPUS. 

If there is some duplication, in the worst case a bit more resources will be spent on cleaning, but it will be deduplicated eventually after merging.

I had an idea to automate this completely though - make the `find-corpus` script a part of the pipeline, then this checksum deduplication might be useful. You can create another issue if you think it's worth the effort.

eu9ene (approved):

PR #67, opened 2022-02-04
bicleaner-ai parallelism fixes

XapaJIaMnu:
How does this look to you @eu9ene 
I kept the CUDA_VISIBLE_DEVICES bit unchanged, but even if it is not exported, we should still be able to get the variable from inside the script. I want to test it before merging though.

EDIT:
Seems both cases are working now.

### DIFF
--- a/pipeline/bicleaner/bicleaner.sh
+++ b/pipeline/bicleaner/bicleaner.sh
@@ -44,10 +44,32 @@ else
     tcol=1
   fi
 
+  #Export cuda visible devices if not set
+  if [ ${#CUDA_VISIBLE_DEVICES} == 0 ]; then   export CUDA_VISIBLE_DEVICES=$(nvidia-smi --query-gpu=index --format=csv,noheader); fi
+
   echo "### Classifying"
-  paste <(pigz -dc "${corpus_prefix}.${SRC}.gz") <(pigz -dc "${corpus_prefix}.${TRG}.gz") |
-    ${cmd} --scol ${scol} --tcol ${tcol} --processes "${threads}"  - - "${pack_dir}"/*.yaml |
-    pigz >"${output_prefix}.scored.gz"
+  if [ "${type}" == 'bicleaner-ai' && ${#CUDA_VISIBLE_DEVICES} > 1 ]; then # Use gnu-parallel'd bicleaner-ai if we have more than 1 GPU
+       #Convert CUDA_VISIBLE_DEVICES to an array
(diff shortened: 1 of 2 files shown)


pipeline/bicleaner/bicleaner.sh
```diff
@@ -44,10 +44,32 @@ else
...
+       #Convert CUDA_VISIBLE_DEVICES to an array
+       export CUDA_VISIBLE_ARRAY=($CUDA_VISIBLE_DEVICES)
+       #Turn on tensorflow logging in bicleaner-ai
+       export TF_CPP_MIN_LOG_LEVEL=0
+       #This function expects a bicleaner yaml and a 1-based index into the CUDA_VISIBLE_ARRAY
+       #Example: /mnt/nanna0/nbogoych/data/data/fr-en/fr-en-prod/biclean/pack/metadata.yaml index_in_CUDA_VISIBLE_ARRAY+1
+       biclean() {
+               export CUDA_VISIBLE_ARRAY=($CUDA_VISIBLE_DEVICES)
+               export CUDA_VISIBLE_DEVICES=${CUDA_VISIBLE_ARRAY[$(($2-1))]}
+               bicleaner-ai-classify --scol 2 --tcol 1 - - $1
```
eu9ene, line 61: should be `--scol ${scol} --tcol ${tcol}`

eu9ene:
`CUDA_VISIBLE_DEVICES={gpus}` will have to go away anyway, otherwise, it will override what slurm sets. I can fix it separately.

XapaJIaMnu:
I think it is reasonable to expect that a slurm user won't specify any other parameter other than num_gpus?

eu9ene:
> I think it is reasonable to expect that a slurm user won't specify any other parameter other than num_gpus?

yes but the actual slurm request depends on `resources: gpus:N`. Maybe we don't have use cases to use less than `num_gpu` now, but it might be useful for evaluation for example. It's fast and it's probably easier to allocate 1 GPU on a cluster than say 8.

eu9ene (changes requested):

pipeline/bicleaner/bicleaner.sh
XapaJIaMnu: Fixed.

eu9ene (approved):

eu9ene:
ok, merging then. If it works for you, I can fix GPU settings in a separate PR.

Issue #21, opened 2021-08-26
Out of memory on shuffling huge datasets


eu9ene:
300M dataset, 128 GB RAM

the workaround is to shuffle dataset after the merge step, disable `--shuffle-in-ram` and use `--shuffle batches`

eu9ene:
This might be a bug of Marian. Memory shouldn't grow after `--shuffle-in-ram` is removed and we should use `--shuffle data` mode. It was discussed in https://github.com/mozilla/firefox-translations-training/pull/70#discussion_r800975032

eu9ene:
Training teachers with `--shuffle batches` leads to such training curves. Maybe other factors are at play here.
<img width="351" alt="Screen Shot 2022-06-10 at 12 40 14 PM" src="https://user-images.githubusercontent.com/2486505/173138550-f474498f-c95a-405a-be20-0ce1bc4edf35.png">

eu9ene:
Related Marian issue: https://github.com/marian-nmt/marian-dev/issues/148

XapaJIaMnu:
`--sqlite` should help but I've found it slow in practise.

jelmervdl:
I suspect the running out of memory, even when --shuffle-in-ram is _not_ used, comes from here:

https://github.com/marian-nmt/marian-dev/blob/042ed8f2e23557d0cdb956aea7d79be8c817e0b0/src/data/corpus.cpp#L227-L241

Assuming that's actually the cause, we could replace it with a two-pass shuffle:
1. Read the unshuffled dataset, and write each line to one of N temp files, chosen randomly for each line. 
  How large N needs to be can probably be determined by looking at how large the input file is, and how much memory is available for shuffling. Might be trickier to estimate if the input file is gzipped.
2. Shuffle each of the temp files as is done now:  
  read it into memory, do `std::shuffle`.
3. Concatenate temp files into the shuffled temp files.  
  Or implement some reader class that takes ownership and reads from the bunch of temp files consecutively as if it were one.

eu9ene:
I didn't see this for some time and I assume it's fixed by using OpusTrainer.

PR #70, opened 2022-02-07
Integrate deduplication in the pipeline

XapaJIaMnu:
Following recent conversations, this pull request integrates a deduplication step that deduplicates the full parallel, as well as the monolingual corpora. I have also removed shuffling, since marian shuffles the data on start anyways, it's never read in order, unless a specific parameter is provided. Tagging @kpu


pipeline/translate/merge-corpus.sh
```diff
@@ -21,9 +23,9 @@ mkdir -p "${tmp_dir}"
 cat <(pigz -dc "${src1}") <(pigz -dc "${src2}") | pigz >"${tmp_dir}/original.src.gz"
 cat <(pigz -dc "${trg1}") <(pigz -dc "${trg2}") | pigz >"${tmp_dir}/original.trg.gz"
 
-echo "#### Shuffling"
+echo "#### Deduplicating"
 paste <(pigz -dc "${tmp_dir}/original.src.gz") <(pigz -dc "${tmp_dir}/original.trg.gz") |
-  shuf |
```
eu9ene, line 26: I would still leave shuffling because sometimes marian is out of memory on big datasets and I have to shuffle batches. See https://github.com/mozilla/firefox-translations-training/issues/21. Just removing `--shuffle-in-ram` doesn't help.

Snakefile
```diff
@@ -76,7 +76,9 @@ bmt_converter = f'{bmt_marian_dir}/marian-conv'
 kenlm = f'{cwd}/3rd_party/kenlm'
```
eu9ene, line 76: let's replace those to `{third_party_dir}/` too

Snakefile
```diff
@@ -421,7 +432,7 @@ if augment_corpus:
         log: f"{log_dir}/split_mono_trg.log"
         conda: "envs/base.yml"
         threads: 1
-        input: f"{clean}/mono.{trg}.gz"
+        input: f"{clean}/mono.{trg}.gz", bin=deduper
```
eu9ene, line 435: extract_lex doesn't depend on deduper, right? why did you add this?

eu9ene:
`merge_corpus `step also does deduplication, should we use deduper there too?

pipeline/translate/merge-corpus.sh
XapaJIaMnu: marian ALWAYS shuffles its input, even if the corpora is not in ram. Shuffling is done before the start of the training and then at every epoch. The only way to prevent shuffling is to use the `shuffle: none` option or `no-shuffle: true`:
https://github.com/marian-nmt/marian-dev/blob/a365bb5ce99135eab29ffe378e0c6c9fb9bf0c1b/src/common/config_parser.cpp#L426

External shuffle is completely redundant.

Snakefile
XapaJIaMnu: I think this is just an issue of the way the diff is displayed in the preview:
```python
if augment_corpus:
    checkpoint split_mono_trg:
        message: "Splitting monolingual trg dataset"
        log: f"{log_dir}/split_mono_trg.log"
        conda: "envs/base.yml"
        threads: 1
-        input: f"{clean}/mono.{trg}.gz"
+        input: f"{clean}/mono.{trg}.gz", bin=deduper
```
It's actually in `checkpoint split_mono_trg:`

eu9ene: right

Snakefile
XapaJIaMnu: Done

pipeline/translate/merge-corpus.sh
eu9ene:
```
--shuffle TEXT=data                   How to shuffle input data (data: shuffles data and sorted 
                                      batches; batches: data is read in order into batches, but 
                                      batches are shuffled; none: no shuffling).
                                      ```
so if I have to use `--shuffle batches` the data will be read in order and shuffled inside the batch if I get it right. wouldn't training quality benefit from preshuffled data in this case?

**Comment by eu9ene** [comment]
> > `merge_corpus `step also does deduplication, should we use deduper there too?
> 
> I thought I am already doing that? I have added my changes to `merge-corpus.sh`. One thing that I am missing is that deduplication should happen before building the spm vocabulary so that we do not adversely affect the word counts. I don't see a way to do it nicely there, so I guess I should just duplicate the work?

There are two `merge_corpus.sh` scripts, in `clean` and `translate` folders. The first one merges data after cleaning, the second one does it after back and forward translation.

**Comment by XapaJIaMnu** [comment]
> > > `merge_corpus `step also does deduplication, should we use deduper there too?
> > 
> > 
> > I thought I am already doing that? I have added my changes to `merge-corpus.sh`. One thing that I am missing is that deduplication should happen before building the spm vocabulary so that we do not adversely affect the word counts. I don't see a way to do it nicely there, so I guess I should just duplicate the work?
> 
> There are two `merge_corpus.sh` scripts, in `clean` and `translate` folders. The first one merges data after cleaning, the second one does it after back and forward translation.

Ohhh, right. Ok, that one is the most important

File: pipeline/translate/merge-corpus.sh (hunk 1 above)
Comment (XapaJIaMnu) [reply pipeline/translate/merge-corpus.sh:26]: again, I doesn't matter because the batches would be shuffled anyways. The important bit for training is to randomise the sequence in which the data is processed. Even if you use `--shuffle batches` the data is still shuffled, because input sentences are sorted by order before being fed into mini-batches. At the moment, the default behaviour `data` that is used makes shuffling redundant.

I am not sure what use case `shuffle: batches` covers, maybe for multilingual model training where  the order in which the data is seen matters a lot more.

Comment (eu9ene) [reply pipeline/translate/merge-corpus.sh:26]: It covers issue #21. I have to use it on 128 GB GPUs for Italian and Portuguese for example. So, the data is shuffled only inside the batch, right? But data is read in order and the sequence of batches is not shuffled. It means it will be trained on corpus1, corpus2, corpus3 etc sequentially which is bad because the data inside every batch will be less diverse and it's easier to stuck in local minima. Do I misunderstand how it works?

**Comment by eu9ene** [comment]
> merge-corpus uses an expand rule to capture all arguments. How can i provide the deduplicator as a dependent step without including it in the input?
> 
> ```
> input:  expand(f"{clean_corpus_prefix}/{{dataset}}.{{lang}}.gz", dataset=train_datasets, lang=[src, trg]), bin=deduper
> ```

expand just creates an array here. you can look at `translate_corpus` step as an example. You can do 
```
input: bin=deduper, datasets=expand(f"{clean_corpus_prefix}/{{dataset}}.{{lang}}.gz", dataset=train_datasets, lang=[src, trg]) 
```

**Comment by XapaJIaMnu** [comment]
Sorry my problem is actually in merge mono:
```snakemake
rule merge_mono:
    message: "Merging clean monolingual datasets"
    log: f"{log_dir}/merge_mono_{{lang}}.log"
    conda: "envs/base.yml"
    threads: workflow.cores
    group: "clean_mono{lang}"
    input:
        lambda wildcards: expand(f"{clean}/mono/{{dataset}}.{{lang}}.gz",
            dataset=mono_datasets[wildcards.lang], lang=wildcards.lang),
            bin=deduper
    output: f"{clean}/mono.{{lang}}.gz"
    params: max_sent=lambda wildcards: mono_max_sent[wildcards.lang]
    shell: '''bash pipeline/clean/merge-mono.sh "{output}" {params.max_sent} {input} >> {log} 2>&1'''
```
The `merge-mono.sh` script takes all inputs via `input`. How can I take a subset of it? I didn't find a way to make the dependent step `deduper` exist outside of the input...

XapaJIaMnu:
Ok, I think I got it... Seems to work now.

eu9ene:
> Sorry my problem is actually in merge mono:
> 
> ```
> rule merge_mono:
>     message: "Merging clean monolingual datasets"
>     log: f"{log_dir}/merge_mono_{{lang}}.log"
>     conda: "envs/base.yml"
>     threads: workflow.cores
>     group: "clean_mono{lang}"
>     input:
>         lambda wildcards: expand(f"{clean}/mono/{{dataset}}.{{lang}}.gz",
>             dataset=mono_datasets[wildcards.lang], lang=wildcards.lang),
>             bin=deduper
>     output: f"{clean}/mono.{{lang}}.gz"
>     params: max_sent=lambda wildcards: mono_max_sent[wildcards.lang]
>     shell: '''bash pipeline/clean/merge-mono.sh "{output}" {params.max_sent} {input} >> {log} 2>&1'''
> ```
> 
> The `merge-mono.sh` script takes all inputs via `input`. How can I take a subset of it? I didn't find a way to make the dependent step `deduper` exist outside of the input...

Just assign a name to the lambda, see `rule evaluate` as an example.

XapaJIaMnu:
i think it works now.

pipeline/translate/merge-corpus.sh
XapaJIaMnu: The data is stored externally in the `tmp` directory (`-T tmp`) and shuffled there in case it can't be in ram. `--shuffle-in-ram` does not modify the behaviour, it only speeds up the shuffling if you have enough RAM.
The way training works it:
1) Shuffle all input data You can see how this happens here: https://github.com/marian-nmt/marian-dev/blob/master/src/data/corpus.cpp#L205
2) Construct batches from the shuffled input, by reading `--maxi-batch` sentences and sorting them by length.
3) Training
The behaviour which you are describing is similar (but not exactly) to `shuffle: batches`, but the sequences in batches is always shuffled. Does this answer the question?

eu9ene: Ok, I got it about the shuffle in ram option. I'm trying to understand how batching works. I'm looking at https://github.com/marian-nmt/marian-dev/blob/master/src/data/batch_generator.h

So here it says not to shuffle the corpus if it's not in `data` mode.
https://github.com/marian-nmt/marian-dev/blob/master/src/data/batch_generator.h#L315 

Then it reads data in order, forming maxi batches. Sentences are ordered inside a maxi batch in a priority queue. Then mini batches are formed from a maxi batch. We read data into maxi batch until we reach `maxi-batch=1000`. The size of a mini-batch is dynamic based on `mini-batch-fit: True` but not bigger than `mini-batch: 1000`. 

Then, when maxi-batch is full we shuffle mini-batches here https://github.com/marian-nmt/marian-dev/blob/aafe8fb5ca8f613f52da7589a72e1a647d51f820/src/data/batch_generator.h#L231.

Does all this mean that maxi batches (1M sentences max) are still read in order and not shuffled or I'm missing something?

XapaJIaMnu: I'm not familiar with `batch` mode, I have not used it.
> The size of a mini-batch is dynamic based on mini-batch-fit: True but not bigger than mini-batch: 1000.

Not entirely correct, as there is a secondary logic in the training scheduling that may decrease or increase the effective mini-batch size (this is controlled by the `optimiser-delay: n` parameter.)

> Does all this mean that maxi batches (1M sentences max) are still read in order and not shuffled or I'm missing something? 

As far as I understand (and I may be wrong here), if running in `batch` mode (which we aren't), ALL batches will be shuffled at the start of each epoch, not just the first 1M, but they will contain the same data. They should be saved to temporary files and read in in different order at the start of every training. The 1M sentence limit is the RAM usage limit (eg the amount you are currently holding in memory/processing). Shuffling is not limited by it.

We should be using `data` mode of batching (and we are).

eu9ene: Well, I have to use batch mode sometimes as a workaround as I pointed out in #21. Nothing else helped me with OOM  on student training. The pipeline has to be scalable to dataset size and I even considered switching it to batch mode everywhere to prevent this situation on huge corpora for high-resource languages. 

So unless we are 100% sure that removing the pre-shuffling won't affect model quality we should leave it as is. This operation is harmless and doesn't take that much time or resources compared to training or decoding, so I consider this removing a micro-optimization that doesn't help much anyway.

XapaJIaMnu: Do you have this particular setup somewhere? Could I try to reproduce it? Marian shouldn't ever consume that much memory if it is not shuffling in ram. If it does it is a bug that we need to report and fix it. The whole reason files are written to a temporary directory is precisely for these scenarios, where memory is limited.

As for the data shuffling, without shuffling batches between epochs, shuffling does very little for translation quality.  The really important bit is to shuffle between epochs, and thinking about it, if the model sees always the same mini-batches, but in different order, that is also suboptimal for the training. This is especially true for student model training which should go for multiple epochs.

My problem with `shuf` is that it makes the pipeline non-deterministic meaning two runs on the same machine will result in slightly different models. I guess it's not a problem in production where you don't care for deterministic results. If you really want shuf there,  I'll put it back, but I am concerned that #21 happens at all.

eu9ene: > Marian shouldn't ever consume that much memory if it is not shuffling in ram

I agree, this looks like a marian bug. I can try to reproduce it again when I reach student training. If it doesn't reproduce, we can remove shuffling. If it does, we can report a marian bug but leave this as a workaround until it's fixed. 

The setup is to train a student model with our current settings on 330M+ sentences and 128 GB GPU. It is pretty easy to get for high-resource languages + back translations + forward translation of 10 mono news-crawl datasets.

> My problem with shuf is that it makes the pipeline non-deterministic

I see, determinist is a different issue and it can be solved with `shuf --random-source=some_file`.

XapaJIaMnu: I have some semblance of deterministic shuf in, let's see if it works

eu9ene (approved):
lgtm, please make sure `make test` passes, at least on those steps you touched

XapaJIaMnu:
I'm currently running the test config on one of our servers, steps seem to pass but I'll wait for a complete pipeline run to be done to check if the intermittent files look correct

eu9ene:
@XapaJIaMnu did you finish testing?

eu9ene:
> The test run got stuck on translating the corpora (degenerate cases with lots of repeated fullstops), but it seems the step before that worked. The corpora that go into the backward and forward model training look reasonable.
> 
> As the pipeline died during the translation, I didn't test [pipeline/translate/merge-corpus.sh](https://github.com/mozilla/firefox-translations-training/pull/70/files#diff-34dc0281212cf1ba2fd14da997aede9d1623476f224bdd1ae8d7280f9d35b270) but i tested the change outside and it worked.
> 
> @todo to create a config that trains quickly enough for test purposes. The current one doesn't work well.
> 
> I think it's safe to merge.

Thanks! Yeah, I tried to create a minimalistic config, but training/translating still takes some time, we can try to reduce it even further.

PR #83, opened 2022-03-28
Preemptable marian training on slurm.

ugermann:
These changes allow marian training jobs on slurm to be interrupted
without losing training progress. The script requests an early warning
from slurm a set amount of time (currently 300 sec.) ahead of end of
time slots, runs marian in the background and shuts it down nicely
with SIGTERM so that marian can save its state of training.
It then examins marians exit status and cancels (or not) the
current slurm job array to avoid wasting subsequent slurm scheduling
slots.

Needs testing.

### DIFF
--- a/pipeline/train/train.sh
+++ b/pipeline/train/train.sh
@@ -1,10 +1,91 @@
 #!/bin/bash
+
 ##
 # Train a model.
 #
 
+##########################################################################
+# Slurm directives; see slurm documentation for what these parameters mean.
+# This script can also be run on the commandline outside of slurm.
+# In this case, slurm directives are simply ignored.
+#
+# Slurm directives start with #SBATCH. Do not remove the leading #.
+# It's part of the directive. Leading ## comments out the
+# respective slurm directive. All of the directives can also be
+# supplied on the command line when the slurm job is submitted.
+#
+# Uncomment and adjust the following to your own account settings. Of
+# course, these can also be provided with the sbatch call on the
+# command line.
+# Your account (-A) and the appropriate slurm partition (-p):
+##SBATCH -A <account>
+##SBATCH -p <partition>
+#
+# How many GPUs and nodes do you want to use?
+#SBATCH --gres=gpu:4
+#SBATCH --nodes=1
+# Set the job name so that we can find our jobs more easily in squeue.
+#SBATCH -J "marian"
+# Send me mail for everything that happens.
+#SBATCH --mail-type=ALL
+# I prefer a meaningful log file name over the one automatically assigned by
+# slurm. This will be placed in the current working directory from which
+# the slurm job was submitted.
+#SBATCH --output=train-model.log
+#
+# ---------------------------------------------------------------------------
+# Handling of time slot limitations.
+# ---------------------------------------------------------------------------
+#
+# Time slot limits imposed by slurm may not be enough to complete training.
+# Luckily, Marian will by default save its current training state if it is
+# terminated with SIGKILL (signal #15). The approach we take is as follows.
+# 1. We schedule the training as an array of slurm jobs that run consecutively.
+# 2. We run marian in the background and keep track of its process id.
+# 3. We ask slurm to send us SIGUSR1 a few minutes ahead of the time limit.
+# 4. We install a signal handler in this script to handle that signal.
+# 5. The signal handler sends SIGTERM to the marian process.
+# 6. We wait for marian to finish and interpret its exit status.
+#    - exit status 0: training finished normally. We cancel the rest of the
+#      slurm job array.
+#    - exit status 143 (= 128 + 15): marian received SIGTERM, saved its
+#      state of training and can be resumed in the next job in the job array.
+#    - any other exit status: something went wrong and we need to investigate.
+#      There's no point in continuing training, so we cancel the job array.
+#
+# Maximum time allowed on CSD3 is 36 hours. Not sure if this needs to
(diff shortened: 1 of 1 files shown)


ugermann:
One open question here is what Snakemake does with non-0 exit codes of scripts. I am completely new to Snakemake ...

ugermann:
@eu9ene For testing on slurm, you could just replace the marian job by `tail -f /dev/null&`. That will also return 143 when killed with SIGTERM.

pipeline/train/train.sh
```diff
@@ -1,11 +1,92 @@
...
+# Slurm directives start with #SBATCH. Do not remove the leading #.
+# It's part of the directive. Leading ## comments out the
+# respective slurm directive. All of the directives can also be
+# supplied on the command line when the slurm job is submitted.
+#
+# Uncomment and adjust the following to your own account settings. Of
+# course, these can also be provided with the sbatch call on the
+# command line.
+# Your account (-A) and the appropriate slurm partition (-p):
+##SBATCH -A <account>
```
eu9ene, line 21: This is not a slurm job script. [This is](https://github.com/mozilla/firefox-translations-training/blob/main/profiles/slurm-csd3/jobscript.sh) a template of slurm jobscript that calls another script generated by snakemake and this another script calls this one at some point. So, these directives won't work.

pipeline/train/train.sh
```diff
@@ -48,7 +129,37 @@ echo "### Training ${model_dir}"
...
+    # Do not cancel the job array! We want to continue in the next slot
+else
+    if [ "$marian_exit_status" == 0 ]; then
+        echo "### Model training is completed: ${model_dir}"
+        echo "###### Done: Training a model"
+    else
+        echo "### Model training failed: ${model_dir}"
+        echo "###### Not done: Training a model"
+    fi;
+    if [[ "${SLURM_JOBID:-''}" != "" ]] ; then
```
eu9ene, line 159: this script is not supposed to know that it's run on slurm. It can be run locally or on any other cluster supported by snakemake.

eu9ene (changes requested):
@ugermann thank you for this PR!

It is a good idea to add graceful shutting down for marian but all slurm related code should be in the appropriate profile like https://github.com/mozilla/firefox-translations-training/tree/main/profiles/slurm-csd3. 

Also, we should merge https://github.com/mozilla/firefox-translations-training/pull/80 first. Killing doesn't seem to be an issue when I tested this PR on my cluster, it loads the latest checkpoint anyway.

Any PR should be tasted by a contributor prior to merging. I recommend switching config to `confit.test.yml` and run `make run-slurm` or `make run-slurm-container` to run a relatively quick test.

pipeline/train/train.sh
ugermann: Checking whether or not this is a slurm script is excactly what this line of code does. If it's not a slurm script, why would SLURM_JOBID be anything other than either not set or empty? Why on earth would any non-slurm environment have SLURM_JOBID set?

pipeline/train/train.sh
ugermann: We appear to disagree what a slurm jobscript is. I'm used to using slurm on CSD3, and I invite you to tell me how my script is fundamentally different from the official documentation examples, e.g. at /usr/local/Cluster-Docs/SLURM/slurm_submit.wilkes2 on CSD3. Just because you prefer singularity containers via snakemake, it does not mean that any script that does not rely on that setup is an invalid slurm script. I'm willing  to learn and collaborate, but I do not like to be brushed off. As far as I can tell, this is a valid slurm script, although it may not work under your setup. As I said: willing to learn and collaborate, but there needs to be a spirit of collaboration, not scoring points.

eu9ene: I welcome collaboration and my comment here is purely technical. Per my understanding, this is not a "slurm script" because it is not designed to be launched separately using `sbatch`. In the current architecture, it is launched by snakemake through multiple other scripts and I would call only the high level script that is actually launched with `sbatch` a "slurm script" (or jobscript) and only this script should be allowed to include slurm directives. Currently, it is [profiles/slurm-csd3/jpbscript.sh](https://github.com/mozilla/firefox-translations-training/blob/main/profiles/slurm-csd3/jobscript.sh).

Currently, everything inside the `pipeline` directory is not really designed to be launched without snakemake, even though it is technically possible. We can discuss this if you see value in running scripts separately locally or on slurm.

pipeline/train/train.sh
eu9ene: Let me try to explain my point better: this is a valid code and it definitely checks things correctly, but this script should not cancel any jobs if we look at it from a current architecture perspective. This script should not be aware of where it is being run. Job scheduling, status checking and cancelling should happen in the profile. 

However, if we'll try to implement training using job arrays, this hack might be handy. I tried to handle this cancelling in the [status script](https://github.com/mozilla/firefox-translations-training/blob/main/profiles/slurm-csd3/status.py) when the job array step is completed, but it's too late and the next step had already started. After that snakemake deletes output of the completed step and breaks everything. This is a bit complicated. But again we'll need this only if we decide to hack job arrays or run this script separately from snakemake.

eu9ene:
> One open question here is what Snakemake does with non-0 exit codes of scripts. I am completely new to Snakemake ...

Snakemake deletes the job output that is specified as "output" in the rule. It is a safety mechanism to prevent corrupted data.

eu9ene:
I believe this is meant to solve specific issues on HPC that we're not using at the moment. Closing

PR #84, opened 2022-04-12
Fine-tune teachers to parallel corpora

lisskor:
This introduces two new options:

1. In addition to external test sets, extract held-out test sets from each parallel training corpus and evaluate models on those held-out sets as well. Controlled by `held-out-dev-test`, `held-out-dev-size`, and `held-out-test-size` in the `datasets` section of config. If using held-out sets but not teacher fine-tuning (see item 2), `held-out-dev-size` should be set to 0, as held-out dev sets will not be utilized.
[Rulegraph of the previously existing workflow](https://github.com/mozilla/firefox-translations-training/files/8474398/rulegraph_existing_workflow.pdf)
[Rulegraph with held-out test sets](https://github.com/mozilla/firefox-translations-training/files/8474400/rulegraph_existing_workflow_with_held_out_test_sets.pdf)

2. Add the step of fine-tuning teacher models to each of the parallel corpora. The parallel data is then forward-translated with the corresponding teacher. Controlled by `fine-tune-to-corpus` in the `experiment` section of config. If `fine-tune-to-corpus: true`, held-out dev and test sets must be used.
[Rulegraph with teachers fine-tuned to parallel corpora](https://github.com/mozilla/firefox-translations-training/files/8474401/rulegraph_with_teacher_fine_tuning_on_parallel_corpora.pdf)

If `held-out-dev-test` and `fine-tune-to-corpus` are set to `false` or not set in the config, the workflow should be exactly the same as it was without my changes.

I have not run a sufficiently large experiment with the new workflow yet, so can't say how significantly it influences multi-domain performance in real-life scenarios.

These changes can be kept in a separate branch if needed.


eu9ene:
I'm trying to understand the logic of fine-tuning on held-out datasets. I imagined that domain adaptation is when we train a model on generic datasets and then finetune on in-domain datasets (for example medical ones). Here we train on all provided datasets including back-translated, then finetune on parallel only, then split all parallel datasets to train/dev/test, and fine-tune on `train` parts only but still on all provided parallel datasets. So there is no option to train on generic parallel datasets and fine-tune on in-domain ones. Do I misunderstand how it's intended to use?

Snakefile
```diff
@@ -524,58 +676,170 @@ if augment_corpus:
...
+            lambda wildcards: expand(f"{translated}/corpus/file.{{part}}.nbest.out",
+                part=find_parts(wildcards, checkpoints.split_corpus))
+        output: f'{translated}/corpus.{trg}.gz'
+        params: src_corpus=clean_corpus_src
+        shell: 'bash pipeline/translate/collect.sh {translated}/corpus {output} {params.src_corpus} >> {log} 2>&1'
+
+# Workflow with fine-tuning teachers to multiple parallel corpora
+# Translate with multiple domain teachers
+if fine_tune_to_corpus:
+    checkpoint split_corpus_domains:
```
eu9ene, line 766: I'm concerned about workflow complexity and duplication of its big chunk. These 4 rules appear to be the same for both scenarios. The only difference is that the second ones use `{{dataset}}` wildcard. I think if we leave just them and use variables for inputs/outputs Snakmake will be able to infer that it has to use those rules for both scenarios (basically `corpus` will be just a special case of `{{dataset}}`).

Snakefile
```diff
@@ -804,3 +1069,36 @@ rule eval_quantized:
         decoder_config='../quantize/decoder.yml'
     shell: '''bash pipeline/eval/eval-quantized.sh "{input.model}" "{input.shortlist}" "{params.dataset_prefix}" \
             "{input.vocab}" "{params.res_prefix}" "{params.decoder_config}" >> {log} 2>&1'''
+
+if fine_tune_to_corpus:
```
eu9ene, line 1073: same here, I think we can make it work for both use cases

Snakefile
lisskor: Totally agree that this doesn't look good. This is the best I've come up with so far, but I'll look into it again and see if I can do this with better wildcard usage and less code

lisskor:
Yes, the goal here is to do decoding with specialized teachers and improve the resulting student performance on the corresponding datasets. We would like to still have a single student model in the end. You got it all correct (except that the held-out train/dev/test sets are actually created before any training so that the teacher models do not see the test data either).

As for introducing manually defined groups of corpora, that's exactly what I was planning to add as a next step!

PR #87, opened 2022-04-21
CI: add automatic test workflow on every push

AmitMY:
I have been having a problem running the `make test` command locally on my machine, so I created continuous integration for the repository, at least until that step.

This PR adds a workflow to run on the designated OS (`ubuntu-18.04`), with no GPUs.

You can see the output of such workflow here:
https://github.com/AmitMY/firefox-translations-training/runs/6109206105

And see, for example, that it fails to run the `make test` command.

### DIFF
--- a/.github/workflows/test-run.yml
+++ b/.github/workflows/test-run.yml
@@ -13,8 +13,8 @@ jobs:
 
       - name: Edit Makefile
         run: |
-          sed -i 's/NUM_GPUS=8/NUM_GPUS=0/' Makefile
-          sed -i 's/SHARED_ROOT.*/SHARED_ROOT=\/tmp/' Makefile
+          sed -i 's/^NUM_GPUS=8/NUM_GPUS=0/' Makefile
+          sed -i 's/^SHARED_ROOT.*/SHARED_ROOT=\/tmp/' Makefile
 
       - name: Install necessary dependencies
         run: |
--- a/.github/workflows/test-run.yml
+++ b/.github/workflows/test-run.yml
@@ -1,0 +1,29 @@
+name: Test Run
+
+on: push
+
+
+jobs:
+  test-run:
+    runs-on: ubuntu-18.04
+
+    steps:
+      - name: Checkout repository files
+        uses: actions/checkout@v2
+
+      - name: Edit Makefile
+        run: |
+          sed -i 's/NUM_GPUS=8/NUM_GPUS=0/' Makefile
+          sed -i 's/SHARED_ROOT.*/SHARED_ROOT=\/tmp/' Makefile
+
+      - name: Install necessary dependencies
+        run: |
+          make conda
+          make snakemake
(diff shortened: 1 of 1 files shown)


eu9ene:
Thank you for this PR! I've been thinking about incorporating the test run in CI but didn't have time to work on it. I think having `dry-run` in CI is useful right away and we can merge it. As for the `test`, the pipeline is not really designed to work without a GPU, this might be the reason why fails. Or Github workflow is missing some system components that are usually preinstalled on a regular server machine. We should probably put `/tmp/logs/ru-en/test` to a CI artifact to investigate this.

Another challenge with test run on CI is that it still takes quite a lot of time even on a server with 8 GPUs (maybe a day) because of decoding. If we'll be able to produce a super stripped config that will make it run in, say, an hour on 1 GPU but still pass all the steps, we should definitely include it in CI.

.github/workflows/test-run.yml
```diff
@@ -0,0 +1,29 @@
...
+      - name: Install necessary dependencies
+        run: |
+          make conda
+          make snakemake
+          make git-modules
+
+      - name: Dry run
+        run: make dry-run
+
+      - name: Test run
```
eu9ene, line 28: So, I suggest removing the test step for now and merging the dry-run. 

We can continue working on the test step in a separate PR. It will require including a GPU machine in the setup and outputting or saving logs of failed jobs to investigate issues. + experimenting with the test config to make it run faster

AmitMY:
Thanks @eu9ene , I wasn't aware, and didn't expect the `test` command should take a day.
I have currently removed it from the github actions pipeline.

However, the point I was (probably unsuccessfully) trying to make with this PR is that the `make test` command fails. I started by running it on my machine, with 4 GPUs, on which it failed as well, and here tried to replicate exactly all necessary steps to make it pass, but still it fails. 

In any case, I agree that a smaller config would be ideal, perhaps trying to overfit 2 sentences, with a small number of layers, just to see that all pipeline steps work as intended and are not broken (like I currently suspect `make test` is)

eu9ene:
> Thanks @eu9ene , I wasn't aware, and didn't expect the `test` command should take a day. I have currently removed it from the github actions pipeline.
> 
> However, the point I was (probably unsuccessfully) trying to make with this PR is that the `make test` command fails. I started by running it on my machine, with 4 GPUs, on which it failed as well, and here tried to replicate exactly all necessary steps to make it pass, but still it fails.
> 
> In any case, I agree that a smaller config would be ideal, perhaps trying to overfit 2 sentences, with a small number of layers, just to see that all pipeline steps work as intended and are not broken (like I currently suspect `make test` is)

Please create an issue and attach the log of the failed job and what machine are you on. If it is some environment or system libraries issue, I recommended using containerized version (changing `test: run-local` to `test: run-local-container`). It is supposed to be more stable.

PR #93, opened 2022-06-08
Strip HTML from ELRC luxembourg data

jelmervdl:
The `mtdata_ELRC-luxembourg.lu-1-eng-fra.{en,fr}.gz` files had HTML in them. This is an attempt to strip that.

This patch is still untested. I've run this script manually on the files, not through the Snakemake pipeline. Hence draft.

### DIFF
--- a/pipeline/clean/fixes/mtdata_ELRC-luxembourg.lu-1.sh
+++ b/pipeline/clean/fixes/mtdata_ELRC-luxembourg.lu-1.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+set -euo pipefail
+tools/strip-html.sh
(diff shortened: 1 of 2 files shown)


jelmervdl:
To fix: I noticed that there is a `\s?` in the first regex. That might not be great if the second column in a corpus contains one of those tags, because then that will strip the `\t` that separates the two sides of the sentence pair.

eu9ene:
Closing this for now since we're going to migrate to OpusCleaner for language and dataset-specific fixes.

marco-c:
@eu9ene won't we need to do this in the OpusCleaner world too?

eu9ene:
We want to clean everything with OpusCleaner eventually including the custom fixes of the datasets.

marco-c:
@eu9ene and we will need to strip HTML from ELRC data in the OpusCleaner config, right? Isn't it still worth keeping this open so we don't forget to do it?

eu9ene:
This change and dataset are not critical regardless and I prefer not to have very old PRs hanging. We'll add this if needed as a part of the OpusCleaner config.

PR #103, opened 2022-08-31
Pass `best_model` name to `train.sh` script

### LINKED ISSUE #102: Bug: `train.sh` does not respect `best-model` setting
In the train file, we save the `final` model always to be the `chrf` model. 
https://github.com/mozilla/firefox-translations-training/blob/main/pipeline/train/train.sh#L53-L54

This makes it so the model has the wrong name.

The pipeline wants:
https://github.com/mozilla/firefox-translations-training/blob/main/Snakefile#L433
> /training/models/spoken-signed/spoken_to_signed/backward/final.model.npz.best-ce-mean-words.npz

But the file created is
> /training/models/spoken-signed/spoken_to_signed/backward/final.model.npz.best-chrf.npz

And because the file has a bad name, the rest of the pipeline crashes.

AmitMY:
fixes #102
(tested for my training, this works)

One thing that is a bit ugly:
To make it so `marian` early stops on the correct metric, I added it to the beginning, such that now it says:
```
[config] valid-metrics:
[config]   - ce-mean-words
[config]   - chrf
[config]   - ce-mean-words
[config]   - bleu-detok
```
Duplicating `ce-mean-words`.

Someone who knows bash better than me should help here


eu9ene (approved):
Good catch, we indeed could configure the thing in a way that the best model doesn't match the validation metric. Your fix should address this. Another option would be to just hardcode `chrf` since we decided that it's the best metric to validate on. If the duplication is fine by marian, we can merge.

Issue #104, opened 2023-01-03
Tabulate 0.9.0 causes issues


hobodrifterdavid:
Was hitting an issue when running 'make dry-run' in local-container mode, Ubuntu 20.04, issue looked like this:

https://stackoverflow.com/questions/74226320/snakemake-tutorial-error-attributeerror-str-object-has-no-attribute-name

The suggested fix (downgrading tabulate to 0.8.10) worked:
https://github.com/snakemake/snakemake/issues/1899

The commands I ran where:
source ../mambaforge/etc/profile.d/conda.sh ; conda activate snakemake
mamba install 'tabulate=0.8.10'

AmitMY:
thanks for pinpointing the issue. fixed in PR, hopefully it gets merged

PR #106, opened 2023-01-11
fix(snakemake): pin tabulate version

AmitMY:
fixes https://github.com/mozilla/firefox-translations-training/issues/104
(confirmed to fix the issue)


eu9ene (approved):
thanks!

PR #117, opened 2023-05-03
Use pretrained OPUS-MT models as teachers and backward models (and other changes)

TommiNieminen:
Hi,

As requested by @andrenatal, here are the current changes made to the pipeline in connection with the GreenNLP project. The main change is the possibility to use OPUS-MT models as teachers and backward models (including multilingual models). The changes also make it possible to use multiple teachers.

In addition to download scripts for models and corpora, I've added some OPUS-MT-specific preprocessing scripts to the pipeline, as OPUS-MT models do not use Marian's inbuilt SentencePiece support. Some of the pre- and post-processing has been added directly to the Snakefile.

I've also added Snakemake configs for supercomputers from CSC (Finnish HPC provider), and made some changes to the containerization.

-Tommi


pipeline/clean/clean-mono.sh
```diff
@@ -42,19 +42,21 @@ else
         | pigz >"${output_prefix}.${lang}.monofix.gz"
 fi
 
+# disabled bue to errors in langid, need to debug sometime
```
eu9ene, line 45: should we just add a config setting to disable it? I don't think we had problems with it for our training

pipeline/clean/clean-corpus.sh
```diff
@@ -71,26 +71,31 @@ test -s "${output_prefix}.${SRC}${TRG}.rule-based.gz" ||
   pigz >"${output_prefix}.${SRC}${TRG}.rule-based.gz"
 
 ######################################################################
-echo "### Language identification"
-test -s "${output_prefix}.${SRC}${TRG}.langid.gz" ||
-  pigz -dc "${output_prefix}.${SRC}${TRG}.rule-based.gz" |
+#Disabled language identification due to crashes, maybe due to small corpus, try with bigger one
```
eu9ene, line 74: should we just add a config setting to disable it?

pipeline/translate/decoder.yml
```diff
@@ -1,10 +1,17 @@
...
+#mini-batch-words: 2000
 maxi-batch: 1000
 maxi-batch-sort: src
 max-length: 200
 max-length-crop: true
 beam-size: 8
-quiet-translation: True
\ No newline at end of file
+quiet-translation: true
+#precision: float16
```
eu9ene, line 14: we use this one for faster encoding. Should we just add another decoder config for opus models and add a config option for this?

profiles/local/config.yaml
```diff
@@ -1,20 +1,20 @@
 verbose: false
 use-conda: true
-resources: gpu=8
-cores: all
+resources: gpu=0
```
eu9ene, line 3: I believe the idea for this config was to train locally on a machine with GPUS

eu9ene:
@TommiNieminen thank you for contributing! I'm happy to see that you used the pipeline. Using pre-trained OPUS models will be valuable for us. 

I see that it's a draft but I briefly looked at it. You made great use of singularity images, configs and snakemake profiles. From the code standpoint, I would suggest preserving the capability to train from scratch using the original bergamot recipe unless we want to fix some bugs and improve it. We can add some config settings and conditional operators for this to distinguish from OPUS use case. I'm not working actively on the project though and @andrenatal should have more context on our direction.

TommiNieminen:
Thanks for the review, @eu9ene. I pushed a quick draft now in case you want to use the OPUS-MT parts of the pipeline for your Taskcluster workflow, but as you noticed, there's some loose ends in the code that I need to clean up.

I saw that elsewhere you were discussing whether to maintain Snakemake support in addition to Taskcluster. On my part, I'm going to keep working with the Snakemake workflow, since I'll be working in a HPC environment with SLURM. So I will be maintaining a Snakemake fork at least for some years, although I expect it to eventually diverge quite a lot from your work (I will be mostly working on retrieval-augmented MT).

bhearsum:
> Thanks for the review, @eu9ene. I pushed a quick draft now in case you want to use the OPUS-MT parts of the pipeline for your Taskcluster workflow, but as you noticed, there's some loose ends in the code that I need to clean up.
> 
> I saw that elsewhere you were discussing whether to maintain Snakemake support in addition to Taskcluster. On my part, I'm going to keep working with the Snakemake workflow, since I'll be working in a HPC environment with SLURM. So I will be maintaining a Snakemake fork at least for some years, although I expect it to eventually diverge quite a lot from your work (I will be mostly working on retrieval-augmented MT).

:wave:. I'm the person who's overseeing the migration into Taskcluster. @andrenatal and I spoke about this a bit earlier today. We agreed that for the time being we're going to maintain support for both Slurm/Snakemake and Taskcluster. I don't think we know what will happen the medium or long term yet, though (and Andre is probably better positioned to talk about it anyways...).

I merged the first part of this work just now, which has caused merge conflicts here due to some minor changes in the pipeline scripts (that Taskcluster is also using). If you want a hand unbitrotting those please let me know and I'd be happy to do so.

andrenatal:
Hi @TommiNieminen. Would please you join our Matrix room so we would coordinate a way to meet: https://matrix.to/#/#firefoxtranslations:mozilla.org? You can find me there under the handle `anatal`

andrenatal:
That's awesome @TommiNieminen, thanks for that!! I'll review it and we can chat about it tomorrow.

configs/config.opusmt-multilingual-test.yml
```diff
@@ -0,0 +1,88 @@
...
+# Training low resource languages might require more tuning of pipeline/training/configs
+###
+
+
+experiment:
+  name: opusmt
+  src: en
+  trg: sw
+  src_three_letter: eng
+  trg_three_letter: swa
```
bhearsum, line 13: Putting this here instead of calling `mtdata` every time we need it is a nice improvement!

configs/config.opusmt-test.yml
```diff
@@ -0,0 +1,84 @@
+####
+# Example of a production config
+# Change language pair, experiment name, datasets and other settings if needed
+# Training low resource languages might require more tuning of pipeline/training/configs
+###
+
+
+experiment:
+  name: opusmt
+  src: en
```
bhearsum, line 10: I'll note that the existing test config uses ru -> en. I think this is to ensure that the reverse locale handling in certain pipeline scripts works OK. Probably not crucial - just calling it out.

bhearsum (commented):
> There are many scripts where the amount and order of parameters has changed, so the Taskcluster setup probably won't work without changes.

Don't worry about this. Most of your changes affect things we haven't set-up in Taskcluster yet, and I'll deal with any fallout that comes from this PR.

It also looks like this should apply cleanly on top of #125 - so that shouldn't disrupt this PR if it lands first. I'll try to hold off merging anything that would conflict until you finish up here.

eu9ene (changes requested):
@TommiNieminen @bhearsum I did another pass at reviewing this. Here are my thoughts:

What I'm not concerned about:
- adding new profiles
- adding new configs
- adding new pipeline steps
- installation scripts
- bicleaner fixes
- backward compatible changes inside the `pipeline` directory

What I'm concerned about:
- **(!) backward incompatible changes inside the `pipeline` directory** This means we'll have to modify the task cluster scripts and retest the training from scratch and I don't think it's the right time to do that
- modifying existing configs including the training one (ideally, everything related to OPUS use case should be in separate configs)
- readme
- changed default singularity image
- complexity of the Snakefile (I don't have any good suggestions here except maybe adding a new file for OPUS use case. Maybe it's fine and we can live with it since we're moving to the TC.) - not critical

On one hand, we need this merged to start experimenting with the pre-trained OPUS models and @TommiNieminen did a great job of making this work (thanks again!). On the other hand, I see risks to break our main training procedure due to the complexity and cost of retesting. Especially since we still testing the integration with the Task Cluster which will serve as our main training scheduler soon.

I believe, to move forward with this, we should address the backward compatibility issues to at least keep our current pipeline that was migrated to Task Cluster working. Even if there are bugs in Snakefile, it would be not that critical. Then we can start experimentation with OPUS models using Snakemake and then gradually migrate it to Task Cluster if it proves to be useful. 

I'm also not sure whether @TommiNieminen has time to address those concerns. Let me know what you think.

eu9ene (approved):
I've tested it and it's ready to merge into a separate branch. I'll address my concerns with readme, configs and compatibility with TC afterward.

PR #118, opened 2023-05-11
Combine split_by_provider and dataset_substitution transforms into a more general from_datasets one

bhearsum:
After a conversation with @andrenatal last week I have a better understanding of how datasets, locale pairs, and different kinds of tasks interact. The tl;dr is that both dataset and locale pair may impact whether or not a task is valid - so we need to be able to generate (or not generate) specific types of tasks by either or both of these.

A concrete example that I'm dealing with here is `bicleaner-ai` not being runnable for `en-ru` due to the datapack not existing.

Unfortunately, this is not fully testable in a PR. Due to the tasks being changed here not running by default (to save resources), and some longstanding issues with Taskcluster's GitHub integration - we can't get them running here. In lieu of that I pushed this to the staging repository and [ran some of the relevant tasks](https://treeherder.mozilla.org/jobs?repo=staging-firefox-translations-training&revision=3d1b7e3f5a7282a617081f9ef78b7c35463b5e81).

(I'm hoping we can find a workaround for this to make future PRs easier to test - but it's not super high priority at the moment.)


bhearsum:
@ahal - apologies that this is removing/reorganizing a bunch of code you just reviewed...my understanding of things changed quite recently. I hope this is decent clean up at least. (I'm still trying to sort out why I can't formally request review from you...)

taskcluster/translations_taskgraph/transforms/from_datasets.py
```diff
@@ -0,0 +1,155 @@
...
+                    continue
+
+                for pair in locale_pairs:
+                    if pair in exclude_datasets.get(provider, {}).get(dataset, []):
+                        continue
+
+                    if pair in exclude_locales:
+                        continue
+
+                    subjob = copy.deepcopy(job)
```
ahal, line 132: Oh wow, this could fan out quite a lot! I glossed over that in the previous implementation. I see why it gets expensive.

ahal (approved):
Seems reasonable!

PR #125, opened 2023-05-30
Implement merge-devset, train-backwards, and evaluate-backwards pipeline steps

bhearsum:
The commit messages have explanations for everything here - but the most interesting parts are the enhancements to some of the transforms to support some scenarios (like dataset categories).

I also want to call the change to `mkdir` in `eval.sh`. I'm pretty sure there's always been a bug here that's simply never been uncovered (making a directory based on `basename` instead of `dirname`) - but it would be great if @eu9ene or @andrenatal could take a look to confirm that. (There's also the usual "support zstd" changes to some scripts - but those are quite straightforward I think.)

Here's a task group from staging showing all of the new steps: https://firefox-ci-tc.services.mozilla.com/tasks/groups/eT1xPEhwRnOzXJEiP1zjOw


gabrielBusta (approved):

gabrielBusta:
Is there a way to add tasks that are not generated for locale pairs, but rather for a single locale? Or I guess I'd just forgo the per_dataset transforms? I was hoping to leverage these

bhearsum:
> Is there a way to add tasks that are not generated for locale pairs, but rather for a single locale? Or I guess I'd just forgo the per_dataset transforms? I was hoping to leverage these

For the puproses of all the `mono` tasks, I think we may want a new `TransformsSequence` in `from_datasets` that yields one task for `src` and one for `trg` (and would probably set the locale as `locale` in each task). That way we're still processing all of the correct locales per dataset. (Happy to chat about this more elsewhere...)

gabrielBusta:
Oh interesting. I like that approach. I'll give it a try on the mono-patch

bhearsum:
This is blocking additional pipeline development so I'm going to go ahead and merge it on the assumption that the pipeline script changes are OK. I'll follow-up with any fixes later if needed.

PR #127, opened 2023-06-06
Add clean mono pipeline step

bhearsum:
As usual, lots of explanation & details in each commit. The most notable part of this is the new `from_datasets:mono` TransformsSequence, which is intended to be use with `mono` dataset types.

Sample run in staging can be found at https://firefox-ci-tc.services.mozilla.com/tasks/groups/ASXO3HEsS1eyQPiYvyt5ag


gabrielBusta (approved):
Thanks Ben! That was fast. I am gonna go ahead and stamp this so we can squash and I can re-base.

PR #128, opened 2023-06-06
Add `merge-mono` pipeline step

gabrielBusta:
Sample run in staging can be found at https://firefox-ci-tc.services.mozilla.com/tasks/groups/Bt9yeBnmQLiKdSMa5uADhQ

I added a new `mono` transform sequence to the `find_upstreams` transforms. Not sure if I should just fuse them into one transform (I think that is possible? I could check for `locale` v. `locale-pair` in the `upstream-config` and make `locale-pair` and `locale` optional in the schema.) I kind of like that, but it's a lil bit more coupling between mono vs. not mono.

### DIFF
--- a/pipeline/clean/merge-mono.sh
+++ b/pipeline/clean/merge-mono.sh
@@ -13,13 +13,24 @@ output=$1
 max_sent=$2
 datasets=( "${@:3}" )
 
-dir=$(dirname "${output}")
-mkdir -p "${dir}"
+LOCALE="${LOCALE:-}"
+COMPRESSION_CMD="${COMPRESSION_CMD:-pigz}"
+ARTIFACT_EXT="${ARTIFACT_EXT:-gz}"
 
-pigz -dc "${datasets[@]}" |
-  ${BIN}/dedupe |
-  shuf -n "${max_sent}" |
-  pigz >"${output}"
-
+if [ -z "${LOCALE}" ]; then
+  dir=$(dirname "${output}")
+  mkdir -p "${dir}"
+  ${COMPRESSION_CMD} -dc "${datasets[@]}" |
+    ${BIN}/dedupe |
+    shuf -n "${max_sent}" |
+    ${COMPRESSION_CMD} >"${output}"
+else
+  datasets=($( printf '%s\n' "${datasets[@]}"| grep ${LOCALE} ))
+  output="${output}/mono.${LOCALE}.${ARTIFACT_EXT}"
+  ${COMPRESSION_CMD} -dc "${datasets[@]}" |
+    ${BIN}/dedupe |
+    shuf -n "${max_sent}" |
+    ${COMPRESSION_CMD} >"${output}"
+fi
 
 echo "###### Done: Merging monolingual datasets"
--- a/pipeline/clean/merge-mono.sh
+++ b/pipeline/clean/merge-mono.sh
@@ -13,24 +13,15 @@ output=$1
 max_sent=$2
 datasets=( "${@:3}" )
 
-LOCALE="${LOCALE:-}"
 COMPRESSION_CMD="${COMPRESSION_CMD:-pigz}"
-ARTIFACT_EXT="${ARTIFACT_EXT:-gz}"
 
-if [ -z "${LOCALE}" ]; then
-  dir=$(dirname "${output}")
-  mkdir -p "${dir}"
-  ${COMPRESSION_CMD} -dc "${datasets[@]}" |
-    ${BIN}/dedupe |
-    shuf -n "${max_sent}" |
-    ${COMPRESSION_CMD} >"${output}"
-else
-  datasets=($( printf '%s\n' "${datasets[@]}"| grep ${LOCALE} ))
-  output="${output}/mono.${LOCALE}.${ARTIFACT_EXT}"
-  ${COMPRESSION_CMD} -dc "${datasets[@]}" |
-    ${BIN}/dedupe |
-    shuf -n "${max_sent}" |
-    ${COMPRESSION_CMD} >"${output}"
-fi
+dir=$(dirname "${output}")
+mkdir -p "${dir}"
+
+${COMPRESSION_CMD} -dc "${datasets[@]}" |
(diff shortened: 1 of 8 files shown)


taskcluster/ci/merge-mono/kind.yml
```diff
@@ -0,0 +1,118 @@
...
+    dataset-config:
+        substitution-fields:
+            - label
+            - description
+            - name
+            - treeherder.symbol
+            - worker.env
+            - upstreams-config.locale
+
+    upstreams-config:
```
bhearsum, line 39: I _think_ you can forego the usage of this transform because your upstream is always the same. Instead, you should be able to directly add dependencies & fetches (similar to what we do for `train-vocab`, which is _always_ downstream of `merge-corpus`: https://github.com/mozilla/firefox-translations-training/blob/main/taskcluster/ci/train-vocab/kind.yml#L79-L89).

pipeline/clean/merge-mono.sh
```diff
@@ -13,13 +13,24 @@ output=$1
...
 datasets=( "${@:3}" )
 
-dir=$(dirname "${output}")
-mkdir -p "${dir}"
-
-pigz -dc "${datasets[@]}" |
-  ${BIN}/dedupe |
-  shuf -n "${max_sent}" |
-  pigz >"${output}"
+LOCALE="${LOCALE:-}"
```
bhearsum, line 16: The changes to this file essentially mean we have two different scripts in one (because all of the actual work commands are duplicated). This isn't great, because it means they're likely to get out of sync at some point.

Is there a reason you can't pass args similar to the existing pipeline script, and avoid the if/else? It looks to me like this ends up getting passed things like:
```
 + output=/data/rw/andre/data/ca-en/dev/clean/mono.ca.gz
 + max_sent=100000000
 + datasets=("${@:3}")
```

As far as I can tell, the zstd files present on in `$MOZ_FETCHES_DIR` in your tasks should all contain `$LOCALE`, so I'm not sure why we need to filter that. And you should be able to set the full output file name in the kind as well?

Let me know if I'm missing something here.

bhearsum (changes requested):
Woohoo, that was quick! I think there's a couple of opportunities for simplification here, noted below. But the step seems to run well and have the correct output.

taskcluster/ci/merge-mono/kind.yml
```diff
@@ -0,0 +1,119 @@
...
+                $VCS_PATH/pipeline/clean/merge-mono.sh
+                /builds/worker/artifacts/mono.$LOCALE.zst
+                {max_sent}
+                $MOZ_FETCHES_DIR/*.zst
+
+    fetches:
+        toolchain:
+            - preprocess
+        clean-mono:
+            - artifact: "{dataset_sanitized}.{locale}.zst"
```
gabrielBusta, line 79: Could this be more than one artifact? Like if we have datasets for different years?

gabrielBusta:
This is closer to what we need. A sample run in staging can be found at: https://firefox-ci-tc.services.mozilla.com/tasks/groups/D2HZ__9jQvy67woCNZc1yQ

taskcluster/translations_taskgraph/transforms/find_upstreams.py
```diff
@@ -134,3 +155,83 @@ def upstreams_for_locales(config, jobs):
...
+            elif dataset_category == "mono-trg":
+                locale = trg
+            else:
+                raise Exception("Don't use `find_upstreams:mono` without the `mono-src` or `mono-trg` category!")
+
+            # Filter out any tasks that aren't for the correct locale.
+            if task.attributes["locale"] != locale or task.attributes["src_locale"] != src or task.attributes["trg_locale"] != trg:
+                continue
+
+            subs = {
```
gabrielBusta, line 200: I forgot to delete this

taskcluster/ci/merge-mono/kind.yml
```diff
@@ -0,0 +1,110 @@
...
+                {max_sent}
+                $MOZ_FETCHES_DIR/*.zst
+
+    fetches:
+        toolchain:
+            - preprocess
+
+tasks:
+    src:
+        label: merge-mono-{provider}-mono-src-{src_locale}-{trg_locale}
```
gabrielBusta, line 79: we might have more than one dataset provider? this should probably be `merge-mono-src-{src_locale}-{trg_locale}`

taskcluster/ci/config.yml
```diff
@@ -86,6 +86,15 @@ datasets:
               trg: fr
             - src: fr
               trg: en
+        news.2019:
```
bhearsum, line 89: Good idea to add this to make sure multiple datasets work here.

taskcluster/ci/merge-mono/kind.yml
bhearsum: Your suggested name makes sense to me. Even if there's multiple providers, they'll still all collapse into the `merge-mono` tasks for `src` and `trg` locales.

taskcluster/translations_taskgraph/util/dataset_helpers.py
```diff
@@ -0,0 +1,9 @@
+def shorten_dataset_name(dataset):
```
bhearsum, line 1: Thanks for moving this to a shared spot!

bhearsum (changes requested):
Other than the couple of things you already pointed out, this looks good to me. The tasks seem to outputting sensible looking artifacts as well. Nice work!

taskcluster/translations_taskgraph/util/dataset_helpers.py
```diff
@@ -7,3 +7,7 @@ def shorten_dataset_name(dataset):
         .replace("news.2020", "n2020")
         .replace("Neulab-tedtalks_train-1", "Ntt1")
     )
+
+
+def sanitize_dataset_name(dataset):
```
bhearsum, line 12: Another nice improvement!

bhearsum (approved):

PR #130, opened 2023-06-08
Fix issue with duplicate tasks/treeherder symbols in `mono` tasks

bhearsum:
This showed up when we added extra locales to the config in #126


gabrielBusta (approved):

gabrielBusta:
Lets try it

PR #133, opened 2023-06-09
Replace DAG.pdf with a more accurate version

bhearsum:
While working on the Taskcluster pipeline I found some information missing from the DAG. Most notably, `merge_augmented` and `merge_translated` are not downstream of any training steps. This seems like it's happening because Snakemake can't calculate the DAG correctly when they are behind a `checkpoint`.To fix this, I temporarily hacked those checkpoints out with more hardcoded versions (link to the diff below).

I also noticed that the `evaluate` steps have an implicit dependency on `train_vocab` (it is referenced by the decoder ymls spit out by the `train` steps - but never explicitly noted in the Snakefile).

The new graph is much, much uglier, of course - but it was also more reflective of reality. It will also be immediately out of date once #117 lands - so maybe we ought to hold off landing this until that is merged, and I can regenerate with that merged in.

Here's a diff of the changes I used to generate this: https://github.com/mozilla/firefox-translations-training/compare/main...bhearsum:firefox-translations-training:dag-hack?expand=1


bhearsum:
Andre and I talked about this on Zoom - he told me just to merge it.

PR #134, opened 2023-06-09
Add split-{corpus,mono} pipeline steps

bhearsum:
The new steps are pretty straightforward, I think. The main notes here are:
- Adjusting the new `merge-mono` steps to use `from_datasets` to ensure we get versions of them for all locales (my bad for not catching this in review...)
- The outputs of these new steps getting compressed into a tarball, so we only have to download one artifact in downstream steps. (This makes it easier to specify `fetches` in those downstream tasks, since the number of files these steps generate won't be known until runtime.)

Runs of these are at https://firefox-ci-tc.services.mozilla.com/tasks/groups/VHrihuG6SXmjr3QkzkyUjg and https://firefox-ci-tc.services.mozilla.com/tasks/groups/HEvP_NzJRPKe7e4K_O05vw

### DIFF
--- a/pipeline/translate/split-corpus.sh
+++ b/pipeline/translate/split-corpus.sh
@@ -11,6 +11,9 @@ corpus_trg=$2
 output_dir=$3
 length=$4
 
+COMPRESSION_CMD="${COMPRESSION_CMD:-pigz}"
+ARTIFACT_EXT="${ARTIFACT_EXT:-gz}"
+
 mkdir -p "${output_dir}"
-pigz -dc "${corpus_src}" |  split -d -l ${length} - "${output_dir}/file."
-pigz -dc "${corpus_trg}" |  split -d -l ${length} - "${output_dir}/file." --additional-suffix .ref
+${COMPRESSION_CMD} -dc "${corpus_src}" |  split -d -l ${length} - "${output_dir}/file."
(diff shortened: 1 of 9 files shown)


hneiva (approved):
Rubber stamp'd. Syntax looks good.

bhearsum:
I'm putting this back in Draft state because I realized that the variable number of files these tasks currently output will cause issues for downstream tasks. I'll need some enhancements to make this work better for Taskcluster.

bhearsum:
I worked through the aforementioned issues, and settled on spitting out N files from this task (allowing downstream tasks to run in N chunks with just a single, known-ahead-of-time, upstream artifac tto fetch), with a bit of additional code to handle possible no-op chunks (rare, but we don't want to fail because of them...).

I've got a couple of runs to point at here. [This one ends up with 2 files](https://firefox-ci-tc.services.mozilla.com/tasks/groups/CohQYwXOQmuvCaxpKjeGGA), to verify the no-op case. [This one ends up with much more than 10 files](https://firefox-ci-tc.services.mozilla.com/tasks/groups/ZQ_4rzGOTny66VmQZES4nA), to ensure that more than 1 file per chunk is handled correctly.

Two other notable things here:
* The addition of the Taskcluster-specific wrapper script for `split` - the main purpose being to handle the aforementioned possibility of empty chunks
* The new `taskcluster` section of the pipeline config. For now, this just exists to configure the `split-chunks` in a central location (which will be used by these tasks as well as at least one downstream one).

gabrielBusta (approved):

PR #139, opened 2023-07-13
Follow-up fixes for Taskcluster pipeline after training a larger model

bhearsum:
A few independent things here that I found while doing a trial run of `fr-en`:
* We need bigger disks for at least some steps. I added 300gb and 1tb options, and used those for some of the steps based on some analysis of artifacts from the Berlin cluster. It's very possible we will need larger disks in other steps as well - but I'd rather underallocate until we have a better idea of what needs to be bumped up.
* Switched GPU instances from `t-linux` to `b-linux` naming.
* Fixed a bug in bicleaner script where it didn't detect reversed locale pairs correctly
* Added a `stage` attribute for `bicleaner`, so that it can be targeted & removed the individual targeting of `bicleaner` and `bicleaner-ai` (only one will ever run).
* Fixed a race condition in the `clean` shell scripts that caused file corruption to break those jobs
* Dropped misusage of `from_datasets:mono` in jobs that are not per-dataset.
* Ensure `src_locale` and `trg_locale` attributes are set on all jobs (this was lost in some places when `from_datasets` was removed from some jobs)
* Increase task deadline to maximum (5 days). We'll need this higher at some point, but we need Taskcluster to support higher values first...

This was tested with [this fr-en run](https://firefox-ci-tc.services.mozilla.com/tasks/groups/QzQ7DoOtR9mEeZqfcmwk3Q).


taskcluster/ci/config.yml
```diff
@@ -1,6 +1,7 @@
 ---
 trust-domain: "translations"
 task-priority: low
+task-deadline-after: 5 days
```
gabrielBusta, line 4: Could a task last longer than this? Do we need to bump this limit in TC?

gabrielBusta (approved):

taskcluster/ci/config.yml
gabrielBusta: Ah never-mind I see you mentioned it in the commit message. What is "much, much higher"?

taskcluster/translations_taskgraph/transforms/cached_tasks.py
```diff
@@ -74,3 +88,48 @@ def add_cache(config, jobs):
...
+        cache = task.pop("cache", None)
+        if cache is None:
+            yield task
+            continue
+
+        dependency_digests = []
+        for p in task.get("dependencies", {}).values():
+            # Here in Translations, we explicit do _not_ invalidate cached
+            # training jobs when non-training jobs are invalidated.
+            if any([kind in p for kind in DONT_INVALIDATE_KINDS]):
```
gabrielBusta, line 115: Mm this might be nice on some other projects. But, here it's really needed.

taskcluster/ci/config.yml
bhearsum: Ultimately, I think we should set task deadlines at least a month ahead. (Individual tasks won't take that long, but deadlines aren't about run time, but rather about "this task must be completed by".)

taskcluster/translations_taskgraph/transforms/cached_tasks.py
bhearsum: Hm...this makes me wonder if this is upstreamable as well...

taskcluster/docker/train/Dockerfile
```diff
@@ -12,6 +12,8 @@ RUN apt-get update -qq \
                           libboost-filesystem1.74.0 \
                           libboost-iostreams1.74.0 \
                           pigz \
+                          curl \
+                          wget \
```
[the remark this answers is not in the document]
bhearsum: Yeah, I can see that argument. I'm mostly adding them both to futureproof against different pipeline scripts wanting both at some point.

Issue #157, opened 2023-08-03
Remove retries for all tasks except the ones that depend on external resources


eu9ene:
If the task is broken and expensive to run, the current 5 retries make the cost of failing even higher. For example, if something is broken in training and we can see it in the logs from the first attempt, we don't really want to run automatic retries. I think the only use case when it might be useful is in those places where we call external API that might be temporarily unavailable, for example, dataset downloading. Even for those, I think a couple of retries with a time interval should be enough. Otherwise, we just delay the notification of failing. Let's discuss.

gregtatum:
It seems reasonable to be able to control this and fine-tune it for our tasks. I created a new "cost & perf" label.

eu9ene:
We can disable all retries for now

eu9ene:
It turns out automatic retries run only for infrastructure issues and require manual specification of `retry-exit-status` which is what we want.

PR #158, opened 2023-08-03
Disable automatic retries in the Taskcluster pipeline

gabrielBusta:
We can retry manually as needed.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -158,7 +158,7 @@ tasks:
                           'tasks_for == "cron"': low
                           'tasks_for == "github-push"|| isPullRequest': very-low
                           $default: lowest  # tasks_for == 'action'
-                  retries: 5
+                  retries: 0
 
                   payload:
                       $let:


gabrielBusta:
#157

gabrielBusta:
Or we could pass 0 retries to the tasks that do not depend on external resources explicitly. But I am leaning towards we explicitly passing retries to the task we want to retry.

eu9ene (commented):
Let's add 2 retires for `download_corpus` and `download_mono` tasks. We periodically hit rate limits on external resources for those

bhearsum:
> Let's add 2 retires for `download_corpus` and `download_mono` tasks. We periodically hit rate limits on external resources for those

If we think this will be a common problem we can also reduce the parallelization of these jobs. (I suggest filing this separately if we want to do that.)

.taskcluster.yml
```diff
@@ -158,7 +158,7 @@ tasks:
                           'tasks_for == "cron"': low
                           'tasks_for == "github-push"|| isPullRequest': very-low
                           $default: lowest  # tasks_for == 'action'
-                  retries: 5
+                  retries: 0
```
bhearsum, line 161: I _think_ this will only disable retries on the decision task. I don't think there's any way to set a default number of retries per task unfortunately :(.

bhearsum (changes requested):

.taskcluster.yml
bhearsum: Yeah. This makes me wonder if `taskgraph.transforms.task` should support setting max retries...

ahal:
Looks like this can be closed out?

PR #162, opened 2023-08-11
Update taskgraph; get rid of some custom transforms that are no longer necessary

bhearsum:
This is mostly just dropping the `chunking`, `task_substition`, and `command_context` transforms in favour of similar code that I recently upstreamed. The most notable part is that there's no longer an `unchunk` TransformsSequence - @ahal pointed out that we can just `from_deps` for this instead. There's more details on these changes in the individual commits.

While testing this I also found a bug in the taskcluster wrapper script for `split-corpus` that caused the `src` and `ref` files to chunk inconsistently. In my run, things broke because there mismatched chunks where some had data and some did not. In larger runs this would be much more serious: we'd end up scoring based on mismatched data.

My staging run with this is over at https://firefox-ci-tc.services.mozilla.com/tasks/groups/ZJOQOrm0QyuEu-7wlsEDwA


taskcluster/ci/extract-best/kind.yml
```diff
@@ -17,8 +18,8 @@ kind-dependencies:
     - translate-corpus
 
 tasks:
-    "{src_locale}-{trg_locale}-{this_chunk}/{total_chunks}":
-        description: extract best {src_locale}-{trg_locale} {this_chunk}/{total_chunks}
+    "{src_locale}-{trg_locale}-{{this_chunk}}/{{total_chunks}}":
```
gabrielBusta, line 21: This might be kind of weird to someone reading the code but I can't think of anything besides escaping the `{ }`.

Adding a comment might be helpful.

taskcluster/scripts/pipeline/split-corpus-taskcluster.sh
```diff
@@ -21,15 +21,16 @@ cd "${output_dir}"
 ls file* | grep -v "\.ref" | sort > src-files.txt
 ls file* | grep "\.ref" | sort > ref-files.txt
 for i in $(seq 1 ${chunks} | tr '\n' ' '); do
-  src_files=$(split -n l/${i}/${chunks} src-files.txt | tr '\n' ' ')
+  # Using `r` ensures that src and ref are chunked consistently.
+  src_files=$(split -n r/${i}/${chunks} src-files.txt | tr '\n' ' ')
```
gabrielBusta, line 25: Nice catch

gabrielBusta (approved):

taskcluster/ci/extract-best/kind.yml
bhearsum: Yeah, it's not ideal. This is actually why [we allow partial substitutions](https://github.com/mozilla/firefox-translations-training/blob/bf7e1b1d080772efb5abdaffad26a2a3e8327146/taskcluster/translations_taskgraph/util/substitution.py#L1) for the remaining transforms here. I don't think that's something we actually want upstream though...

I'll add a comment.

Issue #146, opened 2023-07-20
add testing & linting for taskcluster directory


bhearsum:
At the very least: python and yaml

We might want to consider adding some linting and checks around the `pipeline` directory at the same time.

PR #168, opened 2023-08-29
Remove github workflows

gregtatum:
The workflows have been broken since the taskcluster work has begun. In order to get ready for issue #146, we should probably have a clean indicator to see if the CI is working.


eu9ene:
@gregtatum I would rather fix this one. It tests that Snakemake dry-run works, which is a quick way to check that the DAG is correct. We need snakemake pipeline to be working until we deprecate it and switch to TaskCluster.

gregtatum:
Fair enough I filed #169 instead.

PR #184, opened 2023-09-07
Migrate the github actions to taskcluster

gregtatum:
This is an easy first step for getting the rest of the linting jobs in place, as they can all run in this docker image, and tests/kind.yml file.


gabrielBusta (approved):

Makefile
```diff
@@ -4,15 +4,15 @@
 SHELL=/bin/bash
 
 ### 1. change these settings or override with env variables
-CONFIG=configs/config.prod.yml
-CONDA_PATH=../mambaforge
-SNAKEMAKE_OUTPUT_CACHE=../cache
-PROFILE=local
+CONFIG?=configs/config.prod.yml
```
eu9ene, line 7: My only concern with those is that there might be some random envs on Berlin cluster that will override the settings. I guess let's see. Let's merge this and check afterwards.

eu9ene (approved):

Makefile
eu9ene: @gregtatum or you can check before, since you already have access to the cluster

gregtatum: If that happens I'll follow up by prefixing them.

PR #185, opened 2023-09-07
Add the Black formatter to CI

gregtatum:


gregtatum:
Example intentional CI failure:

image

image

gabrielBusta (approved):

gabrielBusta (changes requested):
Looks like this is in #187?

gregtatum:
I meant for the commits to be looked at individually, but the reviewers seemed to have looked at all commits. I'll close in favor of where the review is actually happening.

PR #194, opened 2023-09-14
Set deadline and max-run-times

gabrielBusta:
90 days is the current value of `queue.max_task_deadline_days` on the cluster.

I set `maxTaskRunTime` to 30 days on the translation workers.

### DIFF
--- a/taskcluster/ci/collect-mono-trg/kind.yml
+++ b/taskcluster/ci/collect-mono-trg/kind.yml
@@ -37,7 +37,7 @@ task-defaults:
     worker-type: b-linux-large
     worker:
         docker-image: {"in-tree": "train"}
-        max-run-time: 86400
+        max-run-time: 2592000
         artifacts:
             - name: public/build
               path: /builds/worker/artifacts
--- a/taskcluster/ci/tests/kind.yml
+++ b/taskcluster/ci/tests/kind.yml
@@ -21,7 +21,7 @@ tasks:
     # taskcluster is the preferred execution environment.
     worker-type: b-linux-large
     worker:
-      max-run-time: 86400
+      max-run-time: 2592000
       docker-image: {in-tree: test}
     run-on-tasks-for: ["github-push", "github-pull-request"]
     run:
@@ -46,7 +46,7 @@ tasks:
     # Run python's black formatter, which formats python files.
     worker-type: b-linux-large
     worker:
-      max-run-time: 86400
+      max-run-time: 2592000
       docker-image: {in-tree: test}
     run:
       command:
@@ -60,7 +60,7 @@ tasks:
     # Run ruff, a python linter.
     worker-type: b-linux-large
     worker:
-      max-run-time: 86400
+      max-run-time: 2592000
       docker-image: {in-tree: test}
     run:
       command:
--- a/taskcluster/ci/alignments/kind.yml
+++ b/taskcluster/ci/alignments/kind.yml
@@ -42,7 +42,7 @@ tasks:
         worker-type: b-linux-large
         worker:
             docker-image: {"in-tree": "train"}
-            max-run-time: 86400
+            max-run-time: 2592000
             artifacts:
                 - name: public/build
                   path: /builds/worker/artifacts
--- a/taskcluster/ci/bicleaner/kind.yml
+++ b/taskcluster/ci/bicleaner/kind.yml
@@ -48,7 +48,7 @@ task-defaults:
             - run.command
 
     worker:
-        max-run-time: 86400
+        max-run-time: 2592000
         env:
             SRC: "{src_locale}"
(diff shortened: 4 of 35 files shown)


gabrielBusta:
Ah `b-linux-large` is `docker-worker` :/

bhearsum (commented):
As I think you noticed elsewhere, I don't think the increased maxRunTime will work for the docker-worker tasks -- but I also don't think they need a max run time longer than 24h. @eu9ene - maybe I'm wrong though? Are there tasks other than the `train` tasks that can have very long runtimes?

gabrielBusta:
OK, I will set it on the "train" tasks

taskcluster/ci/tests/kind.yml
```diff
@@ -72,6 +75,7 @@ tasks:
     run-on-tasks-for: ["github-push", "github-pull-request"]
 
   taskgraph-definition:
+      expires-after: "100 days"
```
bhearsum, line 78: Is this necessary given the change to the default in `config.yml`?

taskcluster/ci/tests/kind.yml
```diff
@@ -19,9 +19,10 @@ tasks:
   snakemake-dry-run:
     # Ensure that the snakemake workflow is still executing correctly, even though
     # taskcluster is the preferred execution environment.
+    expires-after: "100 days"
     worker-type: b-linux-large
     worker:
-      max-run-time: 86400
+      max-run-time: 3600
```
bhearsum, line 25: Good call to revert these back.

bhearsum (commented):
It sounds like we should bump it up for `bicleaner-ai` per the commentary on Slack. I'd like him to double check the overall tasks that are being bumped as well, eg: I don't know if the `evaluate` tasks need such a long run time.

taskcluster/ci/tests/kind.yml
gabrielBusta: This value is defaulting to 28 days and a task can't expire before it's deadline.

That's the default for tasks that run on `"github-pull-request"`

https://github.com/taskcluster/taskgraph/blob/3cd7f06a4a1e3ef72e70d4063e1170bdd8fd74b4/src/taskgraph/transforms/task.py#L1078

https://github.com/taskcluster/taskgraph/blob/3cd7f06a4a1e3ef72e70d4063e1170bdd8fd74b4/src/taskgraph/parameters.py#L233

So it should not be an issue for the training tasks. But, I don't really want to set this on all the CI tasks :/

I could also override the deadline. But I got some schema error when I tried

bhearsum: > This value is defaulting to 28 days and a task can't expire before it's deadline.
> 
> That's the default for tasks that run on `"github-pull-request"`
> 
> https://github.com/taskcluster/taskgraph/blob/3cd7f06a4a1e3ef72e70d4063e1170bdd8fd74b4/src/taskgraph/transforms/task.py#L1078
> 
> https://github.com/taskcluster/taskgraph/blob/3cd7f06a4a1e3ef72e70d4063e1170bdd8fd74b4/src/taskgraph/parameters.py#L233
> 
> So it should not be an issue for the training tasks. But, I don't really want to set this on all the CI tasks :/
> 
> I could also override the deadline. But I got some schema error when I tried

Ah, sorry! I was thinking about `deadline`, but this is about `expires`. 100 days is fine for that.

gabrielBusta: I made a patch for the schema thing https://github.com/taskcluster/taskgraph/pull/344

I feel like overriding the deadline makes more sense for these, because this tasks should probably expire sooner.

gabrielBusta: I made a patch for the schema thing https://github.com/taskcluster/taskgraph/pull/344

I feel like overriding the deadline makes more sense here. These tasks should probably expire sooner.

bhearsum (approved):
I'm approving this, but please don't merge until @eu9ene has had a chance to sanity check the set of tasks with extended runtimes.

taskcluster/ci/collect-mono-trg/kind.yml
```diff
@@ -37,7 +37,7 @@ task-defaults:
     worker-type: b-linux-large
     worker:
         docker-image: {"in-tree": "train"}
-        max-run-time: 86400
+        max-run-time: 2592000
```
gabrielBusta, line 40: If you squint - you will realize this is not going to work 😅

fwiw it would be caught by a test run

eu9ene (approved):
Let's run the whole thing on the full dataset at least once, then we can look at run time per task. However without inspecting the resource utilization it's hard to say whether the tasks utilize resources properly, so they might become faster in future. I would just set everything to max by default, then run it and then adjust accordingly.

PR #195, opened 2023-09-14
Add TC test run to CI

### LINKED ISSUE #193: Include test run in CI
The pipeline is costly to retest if we don't setup an automatic way to check at least that the graph is correct and all the steps run on minimal data.

We already test `make dry-run`

On each PR we should run:
- `make test` to test it under Snakemake
- run under TC with the test config
- when [opusmt integration](https://github.com/mozilla/firefox-translations-training/pull/192) is merged, `CONFIG=configs/config.opusmt-test.ym make run`
- when we add opusmt support to TC, run opusmt test config under TC

We can start with the current settings, but to minimize CI running time we should:
-  optimize the test config and datasets to run as fast as possible
- make sure we utilize caching (it's especially important for compiling Marian and other tools)

eu9ene:
Change the default settings to run faster and include a TC test run in CI on pull request.

Do not run the snakemake pipeline. The complexity to set it up is higher and It's a low priority since we're focusing on TC.

Run time for each task doesn't exceed several minutes:
https://firefox-ci-tc.services.mozilla.com/tasks/groups/MfCuemYfQKC6Td9j0P1BNw
https://firefox-ci-tc.services.mozilla.com/tasks/groups/NZU-xKg7RsGpHeFWh67Ywg

We should run it by default and use `[skip ci]` directive if we want to push something to a PR without rerunning the pipeline.
https://docs.taskcluster.net/docs/reference/integrations/github

It seems the directive didn't work for my commits here though.

fixes #193


taskcluster/ci/tests/kind.yml
```diff
@@ -41,7 +44,43 @@ tasks:
...
+          make dry-run                                                      &&
+          make test-dry-run
+
+  snakemake-run:
+    # Ensure that the snakemake workflow is still executing correctly, even though
+    # taskcluster is the preferred execution environment.
+    worker-type: b-linux-v100-gpu
+    worker:
+      max-run-time: 86400
+      docker-image: {in-tree: train}
```
gabrielBusta, line 56:
```suggestion

```

gabrielBusta: This `docker-image` attr only applies to `docker-worker` (i.e. a container)

https://github.com/mozilla/firefox-translations-training/blob/main/taskcluster/ci/config.yml#L31

`b-linux-v100-gpu` is a `generic-worker` (i.e. a VM)

gabrielBusta: This `docker-image` attr only applies to `docker-worker` (i.e. a container)

`b-linux-v100-gpu` is a `generic-worker` (i.e. a VM)

https://github.com/mozilla/firefox-translations-training/blob/main/taskcluster/ci/config.yml#L31

gabrielBusta (approved):
This is awesome :)

gabrielBusta:
Anyone can open a PR and run this CI - we might want to consider bumping the priority on #197

gabrielBusta:
RE the [skip ci] issue - it's probably related to taskgraph. I wonder if we could make that work @bhearsum

gabrielBusta:
Mm I see you tried to rerun a task on the pull request. I think I am going to whip up a patch so that the on push decision task doesn't get trampled by the pull request decision task. iiuc it's possible to rerun, cancel, and run "Train" on push tasks.

eu9ene:
> Mm I see you tried to rerun a task on the pull request. I think I am going to whip up a patch so that the on push decision task doesn't get trampled (on the GitHub UI) by the pull request decision task. iiuc it's possible to rerun, cancel, and run `Train` on push task groups.

Yeah, it didn't let me. It would be useful to quickly retest with a different config or cancel the run if you don't want it.

eu9ene:
> RE the [skip ci] issue - it's probably related to taskgraph. I wonder if we could make that work @bhearsum

If I got it right, it says in the doc that for pull request action the [skip ci] should be not in the commit but in the PR description.

PR #199, opened 2023-09-20
Split src and trg mono tasks

### LINKED ISSUE #198: TaskCluster pipeline downloads mono datasets for src and trg at the same time
The news-crawl step tries to download mono data for the same year for both src and trg at once, but the configs for mono-src and mono-trg are different.

It also generates incorrect cleaning tasks in the graph (newscrawl, sacrebleu and flores are not supposed to be there):
Screenshot 2023-09-19 at 5 20 55 PM

eu9ene:
The mono src and trg tasks operate on different dataset categories, so we should split them. They also used to generate incorrect `clean-corpus` tasks for a pair of monolingual data for the same year that is not in fact a parallel corpus. 

We can probably find a way to implement this in a more elegant way to reduce copy-paste but we have a problem with code duplication for many other steps too.

Ideally, the tasks that work with monolingual data should also be cached regardless of the direction of training. For example, English monolingual data downloading and cleaning would be retrieved from the cache for almost all experiments for both xx-en and xx-en language pairs. I'm not sure whether it will work here because we mention `{src_locale}` and `{trg_locale}` in the task attributes.

fixes #198


bhearsum:
I'll let @gabrielBusta do the full review, but on this point:

> I'm not sure whether it will work here because we mention {src_locale} and {trg_locale} in the task attributes.

Tasks are cached by what ends up in the `digest_data` - I don't think we include either of this in that - it's usually things like scripts we depend on, parameters that are used, etc. (We also implicitly include [the digest of any tasks a given task depends on](https://github.com/mozilla/firefox-translations-training/blob/52e5fa15d9209c4c995116beb15c041de63dca30/taskcluster/translations_taskgraph/transforms/cached_tasks.py#L101-L124. I don't think this has bearing on this idea, but just FYI). We also set-up caches by cache `type` - which is usually just the name of the pipeline step, eg: `clean-mono`.

With all of that in mind, this _might_ actually work already for mono tasks such as `clean-mono` that are early in the pipeline, assuming the dataset tasks are also cached, before any tasks that do things that are language pair specific (I don't think things later on like `translate-mono` could be re-used even if we wanted to though?)

This is probably quite easily testable by running the `ru-en` config in CI, and then running `en-ru` without tweaking anything else. It's possible that the second run may cache away all of the fetches, cleaning, etc. If not, there are ways we can go about debugging locally.

gabrielBusta:
Thanks @eu9ene the artifacts on the CI run look much better

gabrielBusta (approved):

PR #203, opened 2023-09-20
Override default 28 days expiration on github-pull-request tasks

### LINKED ISSUE #201: ERROR - `expires` cannot be before `deadline`
This error happens while creating a decision task in two unrelated PRs. Something is completely broken in TC.

https://github.com/mozilla/firefox-translations-training/pull/192/checks
https://github.com/mozilla/firefox-translations-training/pull/163/checks

gabrielBusta:
fixes #201

### DIFF
--- a/taskcluster/ci/alignments/kind.yml
+++ b/taskcluster/ci/alignments/kind.yml
@@ -40,6 +40,7 @@ tasks:
                 - worker.env
                 - attributes
         worker-type: b-linux-large
+        expires-after: "90 days"
         worker:
             docker-image: {"in-tree": "train"}
             max-run-time: 86400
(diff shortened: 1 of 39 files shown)


bhearsum (approved):

eu9ene (approved):
@gabrielBusta I'm approving, but it's not clear to me what fixes were added to the transforms. Will we be able to land them in https://github.com/taskcluster/taskgraph/tree/main/src/taskgraph/transforms later?

gabrielBusta:
Yes that's right @eu9ene I will land the fixes upstream. Then, when there's a taskgraph release we can update it here and I will delete these in-tree transforms (and add a way to set the default expiration at the project level)

Issue #205, opened 2023-09-21
Pull request and push decision tasks have the same name


gabrielBusta:
Because these tasks are both called "Decision" the pull request task tramples on the push task in the GitHub UI. Training runs can only be kicked off on push task groups because we can't add tasks to groups created for GitHub pull request.

PR #207, opened 2023-09-21
Run the translations pipeline on fxci level 1 workers and rename Decision Task

### LINKED ISSUE #206: Run the translations pipeline on fxci level 1 workers
Let's run the translations pipeline on fxci level 1 workers  while we iterate on #182 

cc @eu9ene @hwine @bhearsum @aerickson

gabrielBusta:
fixes #206 (and #205)


bhearsum (approved):
Good idea to deal with the decision task name at the same time.

PR #221, opened 2023-10-10
Fixes after training on a full dataset

eu9ene:
- increase reserved GPU workspace to speed up training
- use 4 GPUs for scoring to speed it up
- use a larger worker for student training to prevent out-of-disk and memory errors (40 CPUs, 256GB RAM, 300GB disk)
- add an example full TC config


gabrielBusta (approved):

gabrielBusta:
Looks good to me. The example production TC config is a nice addition!

Issue #154, opened 2023-08-02
Support training continuation


eu9ene:
Sometimes things break or we want to fix something in the training config (for example the stopping criteria) and continue model training from the last checkpoint because it's expensive to start from scratch. When we train on slurm+snakemake, the data stays on the disk with all the Marian checkpoints saved. This means we continue training by default when we rerun the pipeline. If we want to start from scratch, we delete the directory. This works even if we change the training corpus and rerun some cleaning scripts.

We should think about how to implement this manual control on TaskCluster with its different caching paradigm. One idea: we can add a config setting like `training: continue` for each of the models. If it says `continue` we preload artifacts for this step from a previous attempt of training for this experiment and language pair (key experiment_name+lang_pair).

petemoore:
Note, under generic worker it is reasonably trivial to mount artifacts that were created in a previous task. See [`task.payload.mounts`](https://github.com/taskcluster/taskcluster/blob/v54.4.2/workers/generic-worker/schemas/multiuser_posix.yml#L274-L283) and [`readOnlyDirectory`](https://github.com/taskcluster/taskcluster/blob/v54.4.2/workers/generic-worker/schemas/multiuser_posix.yml#L680-L743).

Note, the name "read only directory" is a misnomer - the created directory is read/writable. It really should have just been called "directory" or "unretained directory" or something like that. The point is that if you make changes to it, no worker cache will be updated, because after the task completes and artifacts have been uploaded, the local copy is discarded. This is not a problem, because you can still publish the contents of the directory as a directory artifact, or create an archive from the directory (e.g. a tar.gz file or .zip file, or .lz4 file etc) and publish that as an artifact.

gregtatum:
I wonder if the artifacts should be continued based on the hashing of the previous inputs. That way it will automatically happen until we invalidate part of the previous task graph. To me this seems preferable to having a manual `training: continue`. Perhaps we could also have a cache buster property that is incorporated into the hash, so you could manually update that to re-trigger the training.

eu9ene:
[...]

Issue #161, opened 2023-08-05
Integrate OpusTrainer


eu9ene:
[OpusTrainer](https://github.com/hplt-project/OpusTrainer) is a new training and data augmentation tool developed as a part of the [HPLT](https://hplt-project.org/) project.

This tool can potentially help solve quality issues like casing (see #129 #73). However, it seems like the tool is designed for more advanced use cases and in future will run multiple training runs end-to-end including the training of the backward model. This doesn't fit in our architecture and backward training is already implemented as a separate step. Maybe we can use it only as a data augmentation tool.

eu9ene:
I looked into OpusTrainer and I think we can use it for:
1. On-the-fly data augmentation using modifiers which is supposed to fix some issues like casing, translation of unseen words and symbols etc.
2. Simplifying the pipeline by replacing the two-step teacher training (pre-train and fine-tune) with one. We could use an OpusTrainer config like this one:
```
datasets:
  original: path/to/original # Original parallel corpus
  backtranslated: path/to/backtranslated # Back-translated data

stages:
  - pretrain
  - finetune

pretrain:
  - original 0.5
  - backtranslated 0.5
  - until original 2 # General training until 2 epochs of original

finetune:
  - original 1.0
  - backtranslated 0.0
  - until original inf # Fine-tuning only on original until the early stopping
```
3. Later, training using curriculum learning to have more fine-grained control on how to sequence datasets of different quality. We would likely need to split our `datasets` section in the pipeline config into several sections (dirty, mid, clean) for that. Similar to how it was done here ([config](https://github.com/hplt-project/bitextor-mt-models/blob/main/swa-eng/teacher/opustrainer.config.sw-en.yml), [splitting](https://github.com/hplt-project/bitextor-mt-models/blob/main/swa-eng/teacher/data/1_split_tatoeba_data.py)). I found some papers ([1](https://www.acl-bg.org/proceedings/2017/RANLP%202017/pdf/RANLP050.pdf), [2](https://arxiv.org/pdf/2109.11177.pdf), [3](https://arxiv.org/pdf/2105.04475.pdf)) but it's not clear to me how much it would help to improve quality or speed of training in our case.

[...]

PR #226, opened 2023-10-17
Training continuation

gabrielBusta:
This is work around #154 

I iterated a few times so there's some fluff, and I am not sure how to deal with the vocab.

There's an unused wrapper script called `taskcluster/scripts/pipeline/spm-vocab-taskcluster.sh`

I added that script originally because I thought we only needed to replace the `train-vocab` step on the graph. But actually the vocabulary is specific to each ensemble/model. That means it needs to be specified per pre trained model. So I made the vocab url a part of the pre trained model config for the teacher ensemble. Similarly, when we add continuation to the backward model, we need to specify the vocabulary for that particular backwards model.

On this instance, that means that the `train-teacher` kind (when it is being continued) doesn't really depend on the `train-vocab` step that executed on the pipeline.

I configure the pre trained models in the action input like this:

```
datasets:
  devtest:
    - flores_dev
    - sacrebleu_wmt19
  mono-src:
    - news-crawl_news.2008
  mono-trg:
    - news-crawl_news.2007
  test:
    - flores_devtest
  train:
    - opus_ada83/v1
    - opus_ELRC-3075-wikipedia_health/v1
experiment:
  backward-model: NOT-YET-SUPPORTED
  best-model: chrf
  bicleaner:
    dataset-thresholds:
      opus_ELRC-3075-wikipedia_health/v1: 0.6
      opus_ada83/v1: 0.0
    default-threshold: 0.5
  mono-max-sentences-src: 10000
  mono-max-sentences-trg: 10000
  name: ci
  split-length: 5000
  spm-sample-size: 10000
  spm-vocab-size: 1000
  src: ru
  teacher-ensemble: 2
  trg: en
  use-opuscleaner: 'true'
  vocab: NOT-YET-SUPPORTED
  pretrained-models:
    teacher-base:
      urls:
        - "https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/YQUp-mUgTNSsJS6RjVSUrQ/artifacts/public/build"
        - "https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/bpUaubrpQgOPegv8ZH9xTg/artifacts/public/build"
      mode: use
      type: default
      vocab:
        url: "https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/Vof1ufPxTOyG3OiSj4oHwQ/artifacts/public/build"
marian-args:
  decoding-backward:
    mini-batch-words: '2000'
  decoding-teacher:
    mini-batch-words: '1000'
    precision: float16
  training-backward:
    after: 10u
    dim-vocabs: 1000 1000
    disp-freq: '1'
    save-freq: '5'
    valid-freq: '10'
  training-student:
    after: 10u
    dim-vocabs: 1000 1000
    disp-freq: '1'
    save-freq: '5'
    valid-freq: '10'
  training-student-finetuned:
    after: 10u
    dim-vocabs: 1000 1000
    disp-freq: '1'
    save-freq: '5'
    valid-freq: '10'
  training-teacher-base:
    after: 20u
    dim-vocabs: 1000 1000
    disp-freq: '1'
    save-freq: '5'
    task: transformer-base
    valid-freq: '10'
  training-teacher-finetuned:
    after: 10u
    dim-vocabs: 1000 1000
    disp-freq: '1'
    save-freq: '5'
    task: transformer-base
    valid-freq: '10'
target-stage: train-teacher
taskcluster:
  split-chunks: 2
```


gabrielBusta:
Blah maybe I do need the wrapper around vocab? I started to see this dataset mismatches https://firefox-ci-tc.services.mozilla.com/tasks/GjNToUN5TLOsiRPREl3Ing/runs/0/logs/public/logs/live.log

gabrielBusta:
Ah it's probably because translate mono depends on train-vocab :/

gabrielBusta:
@bhearsum there's some [additional context here](https://docs.google.com/document/d/1fG2YK6ZttBFVZMWHuZjwunc-jdwmN2rBvokoXhgs1bQ/edit#heading=h.jotjwzta1ziy)

gabrielBusta:
Does the backward model have to use the same vocab as the teacher ensemble @eu9ene? I thought they could be different

gabrielBusta:
I also had a "Continue Training" action (more or less a re-trigger but it mounts the artifacts from the task it continues from) but I decided to take out because it would be tricky to deal with edge cases (what happens if you "Continue Training" a task that was in "init" mode for example? The `--pretrained-model` flag would affect the continuation, when it should actually behave like "continue".) Also, not sure how I could simulate a spot outage to actually validate it. So I stashed it.

eu9ene:
@gabrielBusta great job! We're getting close.

> Does the backward model have to use the same vocab as the teacher ensemble @eu9ene? I thought they could be different

Yes, we can use pre-trained teachers and backward models that were trained with different vocabs.

The vocab should basically become a part of the model. It should be stored in the same directory, as a model. It should be copied before training if we train it as a part of the pipeline or copied together with the other model artifacts in pre-trained mode.

For "translate" steps we should use the vocab from the train step artifacts and not the one we trained separately because it might be different for the pre-trained models. 

I like the idea of pulling the pre-trained vocab too if needed, but I would remove the vocab URLs from the config for the pre-trained models and just retrieve it together with the model artifacts. It would be less error-prone and simplify usage.

gabrielBusta:
I see. To echo the idea (now that we discussed it more.) We can make the vocabulary a model was trained on an artifact of the training task. That way we can omit `pretrained-models.teacher-base.vocab.url` (we'd get the vocabulary as an artifact stored along-side the model.) Tasks where the vocabulary must be the one that is bundled with the model could depend on the training task and use the vocabulary artifact there. The training task will use the spm vocab from `train-vocab` if it's not a continuation. Tasks where we want to use the vocabulary trained in the pipeline can depend on `train-vocab` (for example the `alignment` tasks)

bhearsum:
> The vocab should basically become a part of the model. It should be stored in the same directory, as a model. It should be copied before training if we train it as a part of the pipeline or copied together with the other model artifacts in pre-trained mode.
> 
> For "translate" steps we should use the vocab from the train step artifacts and not the one we trained separately because it might be different for the pre-trained models.

:grimacing: - sounds like this is another bug in the original Taskcluster pipeline (we're using using vocab from `train-vocab` everywhere as far as I know). It looks like that will get fixed here though?

> I like the idea of pulling the pre-trained vocab too if needed, but I would remove the vocab URLs from the config for the pre-trained models and just retrieve it together with the model artifacts. It would be less error-prone and simplify usage.

Just to make sure I understand: the idea is that we would simply assume it exists next to any pretrained model we point at? (Or would we point at a taskid instead, or...?)

eu9ene:
> 😬 - sounds like this is another bug in the original Taskcluster pipeline (we're using using vocab from train-vocab everywhere as far as I know). It looks like that will get fixed here though?

It's not a bug. We use the trained vocab everywhere now. When we implement training continuation, it should be flexible enough to use a bunch of different pre-trained models with different vocabs, so it makes sense to make the vocab a part of the model artifacts to make sure we use the right one for each model.

> Just to make sure I understand: the idea is that we would simply assume it exists next to any pretrained model we point at? (Or would we point at a taskid instead, or...?)

Yes, the vocab that we use to train the model should be copied to the model output artifacts. When we use a pre-trained model we should assume it's in the folder with the other artifacts. We can still reuse the pre-trained vocab if we implement pulling it by URL and skipping the vocab training step, but it's optional.

taskcluster/translations_taskgraph/transforms/training_continuation.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+@transforms.add
+def training_continuation(config, jobs):
+    pretrained_models = config.params["training_config"]["experiment"].get("pretrained-models", {})
+    if not any(pretrained_models):
+        for job in jobs:
+            yield job
+    else:
+        model_config = pretrained_models.get("teacher-base")
```
bhearsum, line 55: This transform's name (`training_continuation`) suggests that it could be used on any training step, but some of the details such as this make it clear it only works for `train-teacher` right now. It would be lovely if it could be made generic enough to work with any training step (eg: by parameterizing some things). If not, it ought to be renamed.

(It can probably be refactored in such a way that most of the code can be re-used even if we maintain separate transforms for each training step.)

taskcluster/translations_taskgraph/parameters.py
```diff
@@ -20,10 +20,6 @@ def get_defaults(_):
                 "src": "ru",
                 "trg": "en",
                 "teacher-ensemble": 1,
-                # Used for providing a pretrained backward model. We do not support this yet.
-                "backward-model": "NOT-YET-SUPPORTED",
```
bhearsum, line 24: I will call out that removing these means that we're diverging further from the snakemake configuration. That's probably OK and necessary for this case - and perhaps a sign that keeping up the pretense of having compatible configs is not worthwhile at this point.

taskcluster/translations_taskgraph/parameters.py
```diff
@@ -148,6 +142,23 @@ def get_defaults(_):
...
+                        Required("urls"): [str],
+                        Required("mode"): str,
+                        Required("type"): str,
+                    },
+                    Optional("teacher-finetuned"): {
+                        Required("urls"): [str],
+                        Required("mode"): str,
+                        Required("type"): str,
+                    },
+                    Optional("backwards"): {
```
bhearsum, line 156: This raises a question: are we going to support testing pretrained model support in the quick CI runs? Currently we don't, as these are absent from the defaults above. (If we are, that could totally be added in a follow-up, or filed as an issue for the future.)

taskcluster/ci/train-backwards/kind.yml
```diff
@@ -29,15 +29,19 @@ tasks:
...
                 from-parameters:
                     marian_args: training_config.marian-args.training-backward
+                    pretrained_backward: training_config.experiment.pretrained-models.backward
         task-context:
             from-parameters:
                 best_model: training_config.experiment.best-model
                 src_locale: training_config.experiment.src
                 trg_locale: training_config.experiment.trg
+                pretrained_backward_mode: training_config.experiment.pretrained-models.backward.mode
+                pretrained_backward_type: training_config.experiment.pretrained-models.backward.type
```
bhearsum, line 44: I'm assuming this is just not quite complete yet? (Some of these arguments are being added, but the `training_continuation` transform is not being used yet.)

bhearsum (commented):
This work looks very promising!

There's seems to some issues with the new vocab getting saved -- I'm guessing that there's a `cd` that happens at some point which invalidates the relative path, but I haven't verified that. You should probably ensure there's an absolute path in the `train-taskcluster.sh` wrapper regardless (unless there's a env var we can use to get the absolute path to the artifacts dir).

> [task 2023-10-19T01:24:54.154Z] trainer_interface.cc(605) LOG(INFO) Saving model: ./artifacts/vocab.spm.model
> [task 2023-10-19T01:24:54.197Z] [2023-10-19 01:24:54] Error: SentencePiece vocabulary error: Permission denied: "./artifacts/vocab.spm.model": No such file or directory Error #2
[task 2023-10-19T01:24:54.197Z] [2023-10-19 01:24:54] Error: Aborted from virtual void marian::SentencePieceVocab::create(const string&, const 

I think I could use a quick explainer on why we're using mounts to implement this as well. Is it just to let generic-worker deal with the downloading? AFAICT we don't save on any work compared to having our own script fetch. (In fact, it might even be slower if large files are involved, because the logs appear to show at least one copy being made:
> [taskcluster 2023-10-19T01:24:38.573Z] [mounts] Copying /home/ubuntu/downloads/Dkb7WdruTIyAPyekGV_dzw to /home/ubuntu/tasks/task_169767843132649/artifacts/vocab.spm

bhearsum:
> > Just to make sure I understand: the idea is that we would simply assume it exists next to any pretrained model we point at? (Or would we point at a taskid instead, or...?)
> 
> Yes, the vocab that we use to train the model should be copied to the model output artifacts. When we use a pre-trained model we should assume it's in the folder with the other artifacts. We can still reuse the pre-trained vocab if we implement pulling it by URL and skipping the vocab training step, but it's optional.

Got it (I see there's a large list of other files in the new transform as well), and that we're pointing at a directory (not a specific file) in the new training configuration -- makes sense!

taskcluster/translations_taskgraph/parameters.py
gabrielBusta: Yeah. I went even further and used the kind names in the configuration to simplify the code in the transform

taskcluster/ci/train-backwards/kind.yml
gabrielBusta: Yup. It is going through the transform now.

taskcluster/translations_taskgraph/parameters.py
gabrielBusta: I see there's a `taskcluster` key in the config for `Taskcluster-specific pipeline configuration`. Maybe I can put these configuration fields in there.

taskcluster/translations_taskgraph/parameters.py
gabrielBusta: Is it possible to do that, but run the usual training if the training task was modified? I could trigger an action on the  push decision.

taskcluster/scripts/pipeline/train-taskcluster.sh
```diff
@@ -0,0 +1,61 @@
...
+model_dir=$7
+best_model_metric=$8
+pretrained_model_mode=$9
+pretrained_model_type=${10}
+extra_params=( "${@:11}" )
+
+if [ "$pretrained_model_mode" == "None" ]; then
+    vocab="$MOZ_FETCHES_DIR/vocab.spm"
+else
+    vocab="./artifacts/vocab.spm"
```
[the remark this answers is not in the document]
gabrielBusta: a transform puts them in there

taskcluster/translations_taskgraph/parameters.py
gabrielBusta: I like the idea...

gabrielBusta:
> I think I could use a quick explainer on why we're using mounts to implement this as well. Is it just to let generic-worker deal with the downloading?

gabrielBusta:
> I think I could use a quick explainer on why we're using mounts to implement this as well. Is it just to let generic-worker deal with the downloading?

We could pass down the url and do the downloading in the script. Maybe with a transform that modifies the run command to insert the url for the model we want to use? Or maybe add an environment variable that the script can read? Mounting the files was straight forward 🤷 maybe the wrapper could (should?) be a Python script.

taskcluster/translations_taskgraph/parameters.py
bhearsum: If we're diverging in such big ways I'm not sure we need to worry much about separating the Taskcluster specific things to be honest. @eu9ene - do you have an opinion here?

taskcluster/translations_taskgraph/parameters.py
bhearsum: I imagine it would be possible to dynamically choose whether or not to use a pretrained X.

The downside to doing it with an action task is that it can only be done by those without specific permissions. Not the end of the world, I guess, as anyone who could review/merge presumably has those permissions...

bhearsum:
> > I think I could use a quick explainer on why we're using mounts to implement this as well. Is it just to let generic-worker deal with the downloading?
> 
> We could pass down the url and do the downloading in the script. Maybe with a transform that modifies the run command to insert the url for the model we want to use? Or maybe add an environment variable that the script can read? Mounting the files was straight forward 🤷 maybe the wrapper could (should?) be a Python script.

I don't really have a strong opinion here - if Taskcluster/generic-worker can do it, that's great! I'm just not sure if I'm missing something. When I see `mounts` I guess I expect some sort of disk image, but it sounds like when a `mount` is a task artifact it really just means "download this thing" (which...is not really mounting it...which is why I'm double checking).

gabrielBusta:
@bhearsum ah OK. I think about it more as "downloads this thing." I came across them modifying [the `mounts` that download `run-task` and `fetch-content`](https://searchfox.org/mozilla-central/source/taskcluster/gecko_taskgraph/transforms/job/run_task.py#189-205) on [Bug 1760675](https://bugzilla.mozilla.org/show_bug.cgi?id=1760675).

gabrielBusta:
tbh, I did consider refactoring into a (Python) script (and passing the config through the task environment or command argument.) It seems like it would be more flexible. But the `mounts` are at a point now where we can get value out of this patch soon. So, I was thinking of working out the kinks and landing something that can be used for the continuation of `train-teacher` and `train-backwards` (and `finetune-teacher`?)

gabrielBusta:
I could do some refactoring now, but me and @eu9ene were chatting earlier this week about this and #161 / #219 and it seems once the continuation and opus trainer land, it will be a good time to re-group.

eu9ene:
> I could do some refactoring now, but me and @eu9ene were chatting earlier this week about this and #161 / #219 and it seems once the continuation and opus trainer land, it will be a good time to re-group.

Yes, I'd prefer not to do a major refactoring of the training part at this point. Let's land OpusTrainer and this with the current approach and when we have everything working let's think how to reduce complexity. We'll likely need it to integrate pre-trained OPUS-MT models.

gabrielBusta:
@eu9ene could you try this out? I've been running into issues, but I think some of it might be configuring things in ways that don't reflect actual use cases. Can you do basic continuation/use with the teacher and backward models?

gabrielBusta:
@eu9ene do you know what this could be about?

```
[task 2023-11-14T00:01:33.671Z] [2023-11-14 00:01:33] Error: Requested shape shape=1000x512 size=512000 for existing parameter 'Wemb' does not match original shape shape=32000x1024 size=32768000
[task 2023-11-14T00:01:33.671Z] [2023-11-14 00:01:33] Error: Requested shape shape=1000x512 size=512000 for existing parameter 'Wemb' does not match original shape shape=32000x1024 size=32768000
[task 2023-11-14T00:01:33.671Z] [2023-11-14 00:01:33] Error: Aborted from marian::Expr marian::ExpressionGraph::param(const string&, const marian::Shape&, marian::Ptr<marian::inits::NodeInitializer>&, marian::Type, bool, bool) in /builds/worker/fetches/marian-source/src/graph/expression_graph.h:410
```

https://firefox-ci-tc.services.mozilla.com/tasks/EFSrTdfiTEe17M3qyFE3hw/runs/0/logs/public/logs/live.log#L576-578

(it works fine with the taskcluster model)

eu9ene:
> @eu9ene do you know what this could be about?
> 
> ```
> [task 2023-11-14T00:01:33.671Z] [2023-11-14 00:01:33] Error: Requested shape shape=1000x512 size=512000 for existing parameter 'Wemb' does not match original shape shape=32000x1024 size=32768000
> [task 2023-11-14T00:01:33.671Z] [2023-11-14 00:01:33] Error: Requested shape shape=1000x512 size=512000 for existing parameter 'Wemb' does not match original shape shape=32000x1024 size=32768000
> [task 2023-11-14T00:01:33.671Z] [2023-11-14 00:01:33] Error: Aborted from marian::Expr marian::ExpressionGraph::param(const string&, const marian::Shape&, marian::Ptr<marian::inits::NodeInitializer>&, marian::Type, bool, bool) in /builds/worker/fetches/marian-source/src/graph/expression_graph.h:410
> ```
> 
> https://firefox-ci-tc.services.mozilla.com/tasks/EFSrTdfiTEe17M3qyFE3hw/runs/0/logs/public/logs/live.log#L576-578
> 
> (it works fine with the taskcluster model)

It's probably because this model was trained with different marian parameters

eu9ene:
> How do you know what parameters it was trained on?

You can look at the train.log. Marian outputs the parameters with `[config]` prefix.

taskcluster/translations_taskgraph/parameters.py
eu9ene: It's fine, I have separated example configs for Snakemake and TC anyway

taskcluster/translations_taskgraph/transforms/training_continuation.py
```diff
@@ -0,0 +1,94 @@
+from taskgraph.transforms.base import TransformSequence
+from urllib.parse import urljoin
+import os
+
+CONTINUE_TRAINING_ARTIFACTS = (
+    "devset.out",
+    "model.npz",
+    "model.npz.best-bleu-detok.npz",
```
eu9ene, line 8: It's fine to have this list but there might be more metrics used, for example OPUS-MT use perplexity

taskcluster/translations_taskgraph/transforms/training_continuation.py
```diff
@@ -0,0 +1,94 @@
...
+    "model.npz.optimizer.npz",
+    "model.npz.progress.yml",
+    "model.npz.yml",
+    "train.log",
+    "valid.log",
+    "vocab.spm",
+)
+
+INITIALIZE_MODEL_ARTIFACTS = (
+    "model.npz.best-bleu-detok.npz",
```
eu9ene, line 24: There should be final.* name. It's the file we copy in the end which might not be the case for older models. Also only one model file is required for initialization.

taskcluster/scripts/pipeline/train-taskcluster.sh
```diff
@@ -0,0 +1,60 @@
...
+else
+    vocab="$TASK_WORKDIR/artifacts/vocab.spm"
+fi
+
+export MARIAN=$MOZ_FETCHES_DIR
+
+case "$pretrained_model_mode" in
+    "use")
+        echo "The training mode is 'use', using existing model without further training."
+        cp "$TASK_WORKDIR/artifacts/model.npz.best-$best_model_metric.npz" "$TASK_WORKDIR/artifacts/final.model.npz.best-$best_model_metric.npz"
```
eu9ene, line 37: again it's fine to use with the current pipeline, but  theoretically `best_model_metric` setting can be different in those runs, so we can just copy final.model.npz.best-*.npz

taskcluster/scripts/pipeline/train-taskcluster.sh
```diff
@@ -0,0 +1,60 @@
...
+        cp "$TASK_WORKDIR/artifacts/model.npz.best-$best_model_metric.npz.decoder.yml" "$TASK_WORKDIR/artifacts/final.model.npz.best-$best_model_metric.npz.decoder.yml"
+        exit 0
+        ;;
+    "continue"|"init"|"None")
+        if [ "$pretrained_model_mode" == "init" ]; then
+            extra_params+=("--pretrained-model" "$TASK_WORKDIR/artifacts/model.npz.best-$best_model_metric.npz" "--no-restore-corpus")
```
eu9ene, line 43: why don't we init from the same model we copy to use? final.*

taskcluster/ci/finetune-teacher/kind.yml
eu9ene, line 1: this step will go away after OpusTrainer is landed

eu9ene:
It would be great to add docs on how to use it, structure of the configs and behavior of each mode.

taskcluster/ci/evaluate-teacher-ensemble/kind.yml
```diff
@@ -110,13 +109,11 @@ tasks:
             unique-kinds: false
             kinds:
                 - finetune-teacher
-                - train-vocab
             fetches:
                 finetune-teacher:
                     - artifact: final.model.npz.best-{best_model}.npz
                       dest: model{this_chunk}
                     - artifact: final.model.npz.best-chrf.npz.decoder.yml
-                train-vocab:
```
bhearsum, line 119: praise: Thank you for cleaning this up here, and elsewhere. It's nice to avoid the unnecessary direct dependency on `train-vocab`.

taskcluster/scripts/pipeline/train-taskcluster.sh
```diff
@@ -0,0 +1,60 @@
...
+trg=$4
+train_set_prefix=$5
+valid_set_prefix=$6
+model_dir=$7
+best_model_metric=$8
+pretrained_model_mode=$9
+pretrained_model_type=${10}
+extra_params=( "${@:11}" )
+
+if [ "$pretrained_model_mode" == "None" ]; then
```
bhearsum, line 26: nit: It's odd to be comparing against `None` here. Usually in bash you expect an empty or undefined string as a null value.

taskcluster/scripts/pipeline/train-taskcluster.sh
```diff
@@ -0,0 +1,60 @@
...
+    vocab="$MOZ_FETCHES_DIR/vocab.spm"
+else
+    vocab="$TASK_WORKDIR/artifacts/vocab.spm"
```
bhearsum, line 29: question: I want to make sure I'm understanding this correctly. If we're using a pretrained model, we pull the vocab from a mount? And if not, we pull it from a fetch?

That all seems fine, but if this is true, I would prefer to see the fetch and dependency on `train-vocab` not happening at all if we're using a pretrained model. It's not a large file, but it makes it much less clear where the vocab comes from if we still maintain this dependency while not using anything from it. Please make sure there's a sanity check that ensure that either a pretrained model being used OR a dependency on `train-vocab` in this case.

taskcluster/scripts/pipeline/train-taskcluster.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
+
+set -x
+set -euo pipefail
+
+[[ -v MOZ_FETCHES_DIR ]] || { echo "MOZ_FETCHES_DIR is not set"; exit 1; }
+[[ -v VCS_PATH ]] || { echo "VCS_PATH is not set"; exit 1; }
```
bhearsum, line 7: todo: Replace this with https://github.com/mozilla/firefox-translations-training/blob/0e757b0070a98599d4ade0df01eb85e131718d43/taskcluster/scripts/pipeline/translate-taskcluster.sh#L11 (there's no need to require callers to pass this in - you can figure it out more reliably in the script).

taskcluster/scripts/pipeline/train-taskcluster.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
```
bhearsum, line 1: question: Should we be supporting pretrained models for student training as well? (This is probably a question for @eu9ene).

(If we should, I suggest this is a future enhancement, and not something for this PR.)

taskcluster/scripts/pipeline/train-taskcluster.sh
```diff
@@ -0,0 +1,60 @@
...
+if [ "$pretrained_model_mode" == "None" ]; then
+    vocab="$MOZ_FETCHES_DIR/vocab.spm"
+else
+    vocab="$TASK_WORKDIR/artifacts/vocab.spm"
+fi
+
+export MARIAN=$MOZ_FETCHES_DIR
+
+case "$pretrained_model_mode" in
+    "use")
```
bhearsum, line 35: question: what is the use case for the `use` mode? If we already have models that we want to use, can we not just start training at a later phase, and pick them up as a cached task?

taskcluster/scripts/pipeline/train-taskcluster.sh
```diff
@@ -0,0 +1,60 @@
...
+        echo "The training mode is 'use', using existing model without further training."
+        cp "$TASK_WORKDIR/artifacts/model.npz.best-$best_model_metric.npz" "$TASK_WORKDIR/artifacts/final.model.npz.best-$best_model_metric.npz"
+        cp "$TASK_WORKDIR/artifacts/model.npz.best-$best_model_metric.npz.decoder.yml" "$TASK_WORKDIR/artifacts/final.model.npz.best-$best_model_metric.npz.decoder.yml"
+        exit 0
+        ;;
+    "continue"|"init"|"None")
```
bhearsum, line 41: question: is there an explainer for why we need both of these modes? I see they mount different artifacts, but it's not clear to me what circumstances we would use them both in.

taskcluster/scripts/pipeline/train-taskcluster.sh
bhearsum, line 1: praise: I think it was a great idea to handle all of the training continuation bits in this script instead of trying to have `train.sh` support it in addition to its core job.

idea: Is there a use case for supporting training continuation outside of Taskcluster? If so, we may want to get an issue on file for making this script Taskcluster agnostic. (I explicitly don't want to add that work to this PR even if it may be useful.)

taskcluster/translations_taskgraph/transforms/training_continuation.py
```diff
@@ -0,0 +1,94 @@
...
+    "model.npz.best-ce-mean-words.npz",
+    "model.npz.best-chrf.npz",
+)
+
+
+def get_artifact_mount(url, directory, artifact_name):
+    normalized_url = f"{url}/" if not url.endswith("/") else url
```
bhearsum, line 31: nitpick: It may be better to run this through `urlparse` rather than try to cleanup the url yourself.

taskcluster/translations_taskgraph/transforms/training_continuation.py
```diff
@@ -0,0 +1,94 @@
...
+        for artifact_name in artifact_names:
+            artifact_mounts.append(get_artifact_mount(url, directory, artifact_name))
+        yield artifact_mounts
+
+
+transforms = TransformSequence()
+
+
+def validate_pretrained_models(config):
+    pretrained_models = config.params["training_config"]["experiment"].get("pretrained-models", {})
```
bhearsum, line 53: suggestion (non-blocking): This looks like it's purely validating parameters. If so, this is probably better to do when we first see them in `train.py`.

taskcluster/translations_taskgraph/transforms/training_continuation.py
```diff
@@ -0,0 +1,94 @@
...
+            raise Exception(
+                f"The experiment's 'pretrained-models.backward.urls' ({len(train_backwards['urls'])}) "
+                f"must be equal to one (1). "
+                f"The pipeline's backward model is _not_ an ensemble."
+            )
+    return pretrained_models
+
+
+@transforms.add
+def mount_pretrained_model_training_artifacts(config, jobs):
```
bhearsum, line 75: nitpick: `add_pretrained_model_mounts` would be a better name. The current name suggests that this function is actually mounting something, which is obviously not the case.

taskcluster/translations_taskgraph/transforms/training_continuation.py
```diff
@@ -0,0 +1,94 @@
...
+            pretrained_model: get_artifact_mounts(
+                pretrained_models[pretrained_model]["urls"],
+                "./artifacts",
+                INITIALIZE_MODEL_ARTIFACTS
+                if pretrained_models[pretrained_model]["mode"] == "init"
+                else CONTINUE_TRAINING_ARTIFACTS,
+            )
+            for pretrained_model in pretrained_models
+        }
+        kind = job["task"]["tags"]["kind"]
```
bhearsum, line 88: suggestion (blocking): Use `config.kind` here. (The metadata is not guaranteed to be present.)

bhearsum (changes requested):
I think this looks pretty reasonable overall. I agree with @eu9ene's sentiment that some docs and further explanation would be useful. I have a few specific questions in this regard further down.

I know this needs rebasing, but if I understand correctly the main change will be that all of the `finetune-teacher` references will simply use `train-teacher` instead - which I've assumed while doing this review.

Also, I'm trying to use https://conventionalcomments.org/#labels here, hopefully it helps!

taskcluster/scripts/pipeline/train-taskcluster.sh
eu9ene: Snakemake already supports this because it works with files. This work is Taskcluster specific, so this looks like a good place to handle it.

eu9ene (approved):
I'll be unavailable so approving now. It looks fine overall. I'm ok with merging after addressing the issues with the "init" mode and "final*" models. + @bhearsum's comments.

gabrielBusta:
> I think this looks pretty reasonable overall. I agree with @eu9ene's sentiment that some docs and further explanation would be useful. I have a few specific questions in this regard further down.
> 
> I know this needs rebasing, but if I understand correctly the main change will be that all of the `finetune-teacher` references will simply use `train-teacher` instead - which I've assumed while doing this review.
> 
> Also, I'm trying to use https://conventionalcomments.org/#labels here, hopefully it helps!

@bhearsum there was a little more to it because of the alignments and evaluate stages. CI is back to healthy now

bhearsum:
> > I think this looks pretty reasonable overall. I agree with @eu9ene's sentiment that some docs and further explanation would be useful. I have a few specific questions in this regard further down.
> > I know this needs rebasing, but if I understand correctly the main change will be that all of the `finetune-teacher` references will simply use `train-teacher` instead - which I've assumed while doing this review.
> > Also, I'm trying to use https://conventionalcomments.org/#labels here, hopefully it helps!
> 
> @bhearsum there was a little more to it because of the alignments and evaluate stages. CI is back to healthy now

It looks like you may have missed many of my earlier review comments - can you take a look at those before I review this again, please?

taskcluster/scripts/pipeline/train-taskcluster.sh
gabrielBusta: Mm I can probably do something regarding that in training_continuation.py

taskcluster/scripts/pipeline/train-taskcluster.sh
gabrielBusta: Are we handling the spot terminations gracefully in Taskcluster now? Last time I tried to continue from a preempted VM I couldn't because the worker didn't fail gracefully and lost the checkpoint artifacts.

taskcluster/translations_taskgraph/transforms/training_continuation.py
gabrielBusta: Even when I run it through `urlparse` if the input is missing that `/` there are issues using `urljoin` :/

taskcluster/scripts/pipeline/train-taskcluster.sh
gabrielBusta: We want to be able to load and use models from external archives

taskcluster/scripts/pipeline/train-taskcluster.sh
gabrielBusta: `continue` is "pick up where we left off using the checkpoint" and `init` means start from the beginning but use the model provided as the base model (ie pass in the `--pretrained-model` flag to `marian`)

taskcluster/scripts/pipeline/train-taskcluster.sh
gabrielBusta: I refactored the code so we don't have the fetch and dependency on `train-vocab` if the `vocab.spm` file is mounted.

gabrielBusta:
I think I will do the doc / explainer on a separate [skip CI] PR

bhearsum (approved):
This looks good to me - nice work!!

taskcluster/scripts/pipeline/train-taskcluster.sh
gabrielBusta: Yeah, it's for fine-tuning/refining a pre-existing model

PR #237, opened 2023-10-27
Chunk the 'all' task to avoid max dependency issues

bhearsum:
This gives us N instances of the `all` task to handle cases where we have more than 100 export+evaluate tasks.


gabrielBusta (approved):
😔

gabrielBusta:
Oh actually, this is more straight forward than I thought (since it's a leaf task)

gabrielBusta:
ah, look at the checks, only the first all-run-en task is required to merge

gabrielBusta:
ah, look at the checks, only the first `all-ru-en` task is required to merge

gabrielBusta:
actually, maybe we can just update `all-ru-en` to `all-ru-en-1`?

bhearsum:
> actually, maybe we can just update `all-ru-en` to `all-ru-en-1`?

That seems plausible, yeah. If we really wanted we could also adjust the transform to keep the name the same...but I think it would end up being confusing in training runs where we end up having more than one chunk.

PR #240, opened 2023-11-01
Add CI optimizations for training tasks

gregtatum:
I believe this will prevent CI from running training tasks when touching other folders.

### DIFF
--- a/taskcluster/ci/tests/kind.yml
+++ b/taskcluster/ci/tests/kind.yml
@@ -93,8 +93,6 @@ tasks:
       optimization:
         skip-unless-changed:
             - pipeline/**
-            - envs/**
-            - configs/**
             - taskcluster/**
       run:
           command: >-
--- a/taskcluster/ci/tests/kind.yml
+++ b/taskcluster/ci/tests/kind.yml
@@ -25,6 +25,11 @@ tasks:
       max-run-time: 3600
       docker-image: {in-tree: test}
     run-on-tasks-for: ["github-push", "github-pull-request"]
+    optimization:
+        skip-unless-changed:
+            - pipeline/**
+            - envs/**
+            - configs/**
     run:
       command:
         - bash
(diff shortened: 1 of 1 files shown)


taskcluster/ci/tests/kind.yml
```diff
@@ -85,6 +90,12 @@ tasks:
           max-run-time: 3600
       description: "Test the full `translations_taskgraph` to validate the latest changes"
       run-on-tasks-for: ["github-push", "github-pull-request"]
+      optimization:
+        skip-unless-changed:
+            - pipeline/**
```
gregtatum, line 95: @eu9ene Please double check that I'm covering the required folders here and above.

gregtatum:
I guess the only way to test this is to merge it since I'm touching the taskcluster file here.

taskcluster/ci/tests/kind.yml
```diff
@@ -85,6 +90,12 @@ tasks:
...
+            - envs/**
```
eu9ene, line 96: envs and configs are currently used only for snakemake

eu9ene (commented):
Can we add this optimization to the main Taskcluster CI pipeline run? So that it doesn't rerun if we change anything except the pipeline and taskcluster dirs?

bhearsum (changes requested):
> Can we add this optimization to the main Taskcluster CI pipeline run? So that it doesn't rerun if we change anything except the pipeline and taskcluster dirs?

Big +1. This is done with the `run-on-tasks-for` in https://github.com/mozilla/firefox-translations-training/blob/1d1b4922bbc5f18e312446a8897b7fae42bd5376/taskcluster/ci/all/kind.yml#L34, so adding an `optimization` there would most likely do it.

However...this would also cause real training runs to get optimized away unless they're on a branch that happens to modify one of the `pipeline` or `taskcluster` files (which is not a guarantee, especially after the pipeline has stabilized more).

My suggestion would to add a `all-pr` entry in `all/kind.yml`, and have that run on PRs with this optimization, and then `all` can be saved for real training runs. (Make sure you update `stage` otherwise both will fire in real training runs.

gregtatum:
> Can we add this optimization to the main Taskcluster CI pipeline run? So that it doesn't rerun if we change anything except the pipeline and taskcluster dirs?

I'd like this to be a follow-up unless this is actually blocking this PR. I'm low-context on this repo, especially after spending a couple of months working on Gecko stuff. It will take me a bit to get back up to speed on how everything works, and this seems mergeable with @eu9ene 's suggestion addressed.

bhearsum (approved):

eu9ene (approved):

PR #241, opened 2023-11-02
fix(snakefile): get cuda path from the machine

AmitMY:
This is just an idea, and could and probably should be extended to other `tool` variables.

On the University of Zurich slurm cluster, you must run `module load gpu` to make cuda available, and when it is, cuda is set weirdly, like in `/apps/opt/spack/linux-ubuntu20.04-x86_64/gcc-9.3.0/cuda-11.8.0-ifsva5xy3gliwzlokewxc5lxklbej3gj` for example.

I don't think this path is persistent, and so I was thinking it would be nicer to just take the path from the machine rather than the config.

(Note: I did not yet successfully run even a `test` job on this cluster, but this was one issue I had to solve)

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -1,5 +1,6 @@
 import yaml
 import os
+import subprocess
 
 from snakemake.utils import min_version
 from pipeline.bicleaner import packs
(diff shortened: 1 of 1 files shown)


eu9ene (commented):
Hi @AmitMY! I'm glad you're trying to use the training pipeline for your project!

Ideally, the cluster-specific modification should be implemented in the profiles. You could try to detect the cuda dir in the [submit.py](https://github.com/mozilla/firefox-translations-training/blob/c3b5b63cc503ba026238b0cfe4cfddb1bc0ae866/profiles/slurm-moz/submit.py#L12) for your profile and update the snakemake config dynamically. 

Often the head node where the snakemake process runs doesn't have any GPUs. Also, we might want to run the `snakemake dry-run` on any machine without a GPU. You can see that the CI is failing now.

PR #243, opened 2023-11-04
Update intel keys from 2019 to 2023

AmitMY:
I don't know why, but I was having trouble using the 2019 file, and it works for me with the 2023 file


eu9ene (approved):
We had this issue in CI for the model registry too, and I updated it there in the same way. Thanks!

PR #267, opened 2023-11-18
fix(custom-mono): create parent directory if not exists

### LINKED ISSUE #100: `download_mono` fails to load custom datasets with `/` in name
To set a `custom-corpus`, I do something like this, which works at loading and saving the data:
>  custom-corpus_/custom_corpus/fingerspelling/devtest

However, doing something similar for `mono-corpus`:
>  custom-mono_/custom_corpus/common_words/mono

It can not create the file, because it is creating this directory:
```
/training/data/spoken-signed/spoken_to_signed/original/mono/custom-mono_/custom_corpus/common_words/original/
```
but trying to save the file in this path
```
/training/data/spoken-signed/spoken_to_signed/original/mono/custom-mono_/custom_corpus/common_words/original/custom-mono_/custom_corpus/common_words/mono.original.spoken.gz
```

https://github.com/mozilla/firefox-translations-training/blob/main/pipeline/data/download-mono.sh#L28

---- 
Current workaround: manually create the directory it is trying to save to

AmitMY:
fixes #100

### DIFF
--- a/pipeline/data/importers/mono/custom-mono.sh
+++ b/pipeline/data/importers/mono/custom-mono.sh
@@ -13,6 +13,7 @@ lang=$1
 output_prefix=$2
 dataset=$3
 
+mkdir -p "$(dirname "$output_prefix")"
 cp "${dataset}.${lang}.gz" "${output_prefix}.gz"
 
 


gregtatum (approved):
Thanks! Looks good to me.

PR #276, opened 2023-11-27
Add taskgraph-diff ci job

bhearsum:
These task diffs the full JSON graph of a couple of different training runs, comparing a PR against a base revision, or a push against the previous push.
    
It is an informational task designed not to fail even if there are differences. Any parameters that cause changes in the resulting taskgraph will cause an artifact with a diff of the before/after to be generated.

(Marking this as a draft for now - it most likely depends on #275 landing first to function correctly.)

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -37,11 +37,22 @@ tasks:
                   'tasks_for == "github-push"': ${event.ref}
                   'tasks_for == "github-release"': '${event.release.target_commitish}'
                   'tasks_for in ["action", "cron"]': '${push.branch}'
+          base_ref:
+              $switch:
+                  'tasks_for[:19] == "github-pull-request"': ${event.pull_request.base.ref}
+                  'tasks_for == "github-push" && event.base_ref': ${event.base_ref}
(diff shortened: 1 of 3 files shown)


bhearsum:
I believe this is ready to go. Quantize is likely going to fail due to #281 - but that is certainly unrelated to this work.

@gregtatum - I'm mainly flagging you for the Makefile changes, as the required arguments to the new target are breaking new ground there, but feel free to comment on anything else of course!

gregtatum (approved):
The Makefile Seems reasonable to me.

gabrielBusta (approved):

gabrielBusta:
Nice. Kinda makes the "tests-taskgraph-definition" CI step redundant (since "tests-taskgraph-diff " generates the full taskgraph for the current rev too)

bhearsum:
> Nice. Kinda makes the "tests-taskgraph-definition" CI step redundant (since "tests-taskgraph-diff " generates the full taskgraph for the current rev too)

This is a good point. I'm a bit split on whether or not we should combine them...I think I'll leave things be for now.

PR #278, opened 2023-11-29
Don't run the training pipeline in PRs unless the pipeline changes (fixes #260)

### LINKED ISSUE #260: Add optimization to the main taskcluster run
In PR #240 I added some initial optimization calls to only run tests when pertinent files changed. This should be done for the main runs.

@evgeny stated:

> Can we add this optimization to the main Taskcluster CI pipeline run? So that it doesn't rerun if we change anything except the pipeline and taskcluster dirs?

@bhearsum replied:

> Big +1. This is done with the `run-on-tasks-for` in https://github.com/mozilla/firefox-translations-training/blob/1d1b4922bbc5f18e312446a8897b7fae42bd5376/taskcluster/ci/all/kind.yml#L34, so adding an `optimization` there would most likely do it.
>
> However...this would also cause real training runs to get optimized away unless they're on a branch that happens to modify one of the `pipeline` or `taskcluster` files (which is not a guarantee, especially after the pipeline has stabilized more).
>
> My suggestion would to add a `all-pr` entry in `all/kind.yml`, and have that run on PRs with this optimization, and then `all` can be saved for real training runs. (Make sure you update `stage` otherwise both will fire in real training runs.

bhearsum:
This needs a bit more testing and tohught still, but in my quick local test it seems to have worked.

It required a bit more work than I thought, as there's no way to tell `skip-unless-changed` to look at all subdirectories _except_ one (which is important if we want to exclude the pipeline when the `test` kind changes).


bhearsum:
This PR also raises the fact that we want to remove `all-ru-en-1` as a required status check, since we will no longer expect it to always run. (We could add some/all of the `test` kind tasks as required, though.)

bhearsum:
In addition to what's run here, I've sanity checked what tasks we get with `target-graph` with a few different parameter sets:
- The [checked in action task parameters](https://github.com/mozilla/firefox-translations-training/blob/main/taskcluster/test/params/large-lt-en.yml) still select `all-ru-en-1`, as expected
- [On-push parameters](https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/cNUieFmvQCy_4nawQxm2tg/runs/0/artifacts/public%2Fparameters.yml) select the `docker-image`, `fetch`, `tests`, and `toolchain` tasks, as expected
- Running against the same parameters used in the decision task from this PR, but tweaking `base_rev` to be the same as `head_rev` results in the `docker-image`, `fetch`, `tests`, and `toolchain` tasks. This confirms that this PR does what it intends (not fire any pipeline tasks if the pipeline doesn't change between base and head rev).

gabrielBusta (approved):

PR #287, opened 2023-12-05
Use shorter version of `base_ref` for branches in the canonical repository

bhearsum:
This fixes a regression caused by #276 that has caused decision tasks to fail for non `main` branches in this repository.

The "fix" is to refer to these branches by their short name (rather than `refs/heads/foo`). (This is not a great fix...but it's what we have to do to work around https://github.com/taskcluster/taskgraph/issues/380 for now.)

https://firefox-ci-tc.services.mozilla.com/tasks/Y6m23CknT3ig_4k8kvhCGQ shows a push task to this repository that is now working.

(The diff here looks bad, but it's mostly whitespace changes...the actual changes are the inclusion of the definition of `short_base_ref`, and the changes to pass it along to taskgraph instead of `base_ref`.)

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -69,222 +69,233 @@ tasks:
           isPullRequest:
               $eval: 'tasks_for[:19] == "github-pull-request"'
       in:
-          $if: >
-              tasks_for in ["action", "cron", "github-push"]
-              || (isPullRequest && pullRequestAction in ["opened", "reopened", "synchronize"])
-          then:
-              $let:
-                  level: 1
-              in:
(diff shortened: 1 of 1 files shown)


gabrielBusta (approved):

gabrielBusta:
Interesting, I have been running into this... I don't totally understand this issue

bhearsum:
> Interesting, I have been running into this... I don't totally understand this issue

I don't fully understand it either, if I'm being honest. But `refs/heads/foo` is not a valid ref to the `foo` branch when you clone this repo. Frankly, I think it's a bug in taskgraph...

PR #284, opened 2023-12-02
Train en-ca

gregtatum:
Creating a branch was not enough, let's try a PR. [skip ci]

Issue #292, opened 2023-12-11
opuscleaner fails to pip install


bhearsum:
From [a recent PR](https://firefox-ci-tc.services.mozilla.com/tasks/HGA6hgcuTGiLFKAjUuj3Tg/runs/0/logs/live/public/logs/live.log):
```
[task 2023-12-11T14:11:35.151Z] Building wheels for collected packages: sacremoses, opuscleaner
[task 2023-12-11T14:11:35.151Z]   Building wheel for sacremoses (setup.py): started
[task 2023-12-11T14:11:35.484Z]   Building wheel for sacremoses (setup.py): finished with status 'done'
[task 2023-12-11T14:11:35.487Z]   Created wheel for sacremoses: filename=sacremoses-0.0.53-py3-none-any.whl size=895260 sha256=79c9665f6124a686bb06caf5351973d2c33879e234bd146d715e1bf599e00d07
[task 2023-12-11T14:11:35.487Z]   Stored in directory: /tmp/pip-ephem-wheel-cache-7ltzqd1n/wheels/00/24/97/a2ea5324f36bc626e1ea0267f33db6aa80d157ee977e9e42fb
[task 2023-12-11T14:11:35.490Z]   Building wheel for opuscleaner (pyproject.toml): started
[task 2023-12-11T14:11:35.639Z]   Building wheel for opuscleaner (pyproject.toml): finished with status 'error'
[task 2023-12-11T14:11:35.645Z]   error: subprocess-exited-with-error
[task 2023-12-11T14:11:35.645Z]   
[task 2023-12-11T14:11:35.645Z]   × Building wheel for opuscleaner (pyproject.toml) did not run successfully.
[task 2023-12-11T14:11:35.645Z]   │ exit code: 1
[task 2023-12-11T14:11:35.645Z]   ╰─> [18 lines of output]
[task 2023-12-11T14:11:35.645Z]       Traceback (most recent call last):
[task 2023-12-11T14:11:35.645Z]         File "/usr/lib/python3/dist-packages/pip/_vendor/pep517/in_process/_in_process.py", line 363, in <module>
[task 2023-12-11T14:11:35.645Z]           main()
[task 2023-12-11T14:11:35.645Z]         File "/usr/lib/python3/dist-packages/pip/_vendor/pep517/in_process/_in_process.py", line 345, in main
[task 2023-12-11T14:11:35.645Z]           json_out['return_val'] = hook(**hook_input['kwargs'])
[...]

PR #298, opened 2023-12-12
Add documentation to various parts of the scripts and pipeline

gregtatum:
This is work that I split off from my #284 work. My goal here is to:

* Document argument parameters in the scripts and Taskcluster tasks.
* Document scripts with a header that explains how they are used with examples.
* Document every parameter of the example training config.
* Add links to literature and prior art for when certain hyper-parameters are chosen or design choices made
* Be consistent with argument and variable naming to make it easier to search across the codebase.

The commits are logically ordered if you wish to review them that way.

### DIFF
--- a/configs/tc.prod.yml
+++ b/configs/tc.prod.yml
@@ -1,9 +1,154 @@
 # Example of a production config for Taskcluster
+#
+# See the training guide for more information:
+#   https://mozilla.github.io/firefox-translations-training/training-guide.html
+#
+# The defaults for these parameters are available at:
+#   taskcluster/translations_taskgraph/parameters.py
+
+# An "experiment" is an individual training run.
+experiment:
+  # Provide an identifiable name for your experiment.
+  name: baseline_en_ru
+
+  # The source and target languages. This is the language tag part of the
+  # BCP 47 locale identifier.
+  src: en
+  trg: ru
+
+  # The metric to use for computing the best model validation.
+  #   cross-entropy, ce-mean-words, perplexity, valid-script, translation, bleu,
+  #   bleu-segmented, chrf
+  # See: https://github.com/marian-nmt/marian/blob/65bf82ffce52f4854295d8b98482534f176d494e/src/common/config_parser.cpp#L588-L592
+  best-model: chrf
+
+  # Use the Opus Cleaner tool on the data cleaning step.
+  # https://github.com/hplt-project/OpusCleaner
+  use-opuscleaner: "true"
+
+  # Bicleaner is a tool that aims at detecting noisy sentence pairs in a parallel corpus.
+  # It indicates the likelihood of a pair of sentences being mutual translations (with a
+  # value near to 1) or not (with a value near to 0). Sentence pairs considered very
+  # noisy are scored with 0. Bicleaner AI will be used first if the language is
+  # available, otherwise it will fallback to the original non-AI Bicleaner.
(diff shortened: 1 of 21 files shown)


configs/tc.prod.yml
```diff
@@ -1,9 +1,154 @@
...
+  # See section 4 for their conclusions.
+  #   https://aclanthology.org/2020.findings-emnlp.352.pdf
+  #
+  # The larger the vocab, the slower the training and the inference will be as each
+  # token in the vocab contributes to the amount of probabilities that will need
+  # to be generated when making a token prediction. This is a trade off on the quality
+  # of the translation, and the performance.
+  spm-vocab-size: 32_000
+
+  # Determine how many teachers to train. The teachers will be identical except they
```
eu9ene, line 90: I think the better place for the extended docs would be in the training guide in the /docs. I would recommend limiting comments here to one liners with a simple description of the parameter, because we often use this config as a template for the new runs and it will be harder to navigate with all the embedded docs.

configs/tc.prod.yml
```diff
@@ -1,9 +1,154 @@
...
+
+  # The number of sentences split out into separate files for the monolingual data.
+  # Also see taskcluster.split-chunks below.
+  split-length: 2_000_000
+
+  # How many bytes of the corpus are sampled to be used in SentencePiece
+  # tokenization training.
+  spm-sample-size: 10_000_000
+
+  # How big the vocab will be. The vocab is shared between the source and the
```
eu9ene, line 67: It's unlikely we'll need to train something with less than 5M sentences, so the vocab size will likely never change. I recommend moving this extra info to the training guide. It will allow adding proper links and reduce size of this config.

eu9ene (commented):
I suggest consolidating docs in the /docs directory and keeping only simple descriptions or code clarifications in the code and in the configs. It will allow finding things easier in one place and use a nice formatting for links etc.

pipeline/translate/collect.sh
```diff
@@ -1,25 +1,53 @@
 #!/bin/bash
 #
-# Merges translation outputs into a dataset
+# Collects chunked translation data of the form "out-file.N.out" where N is a number.
+# The datasets are initially chunked so that tasks can work on smaller sets of data
```
bhearsum, line 4: suggestion (non-blocking): Perhaps something like "The datasets are chunked in earlier in the pipeline by `split-{corpus,mono}.sh`'? (The current usage of "initially" here made it sound at first like the chunking was happening _here_.

eu9ene:
> Thank you for adding such great documentation on these pipeline scripts! I struggled with these when implementing the Taskcluster pipeline, and I think these new docs will make it much easier for others to understanding how the pieces fit together without relying on squinting at logs.

+1 Thank you for making the pipeline easier to understand!

configs/tc.prod.yml
gregtatum: Ok, seems reasonable. I'll document the reasoning too. This was the source of one of my failures in training en-ca, since the pipeline script failed for me. 32,000 seems like a good default to me.

configs/tc.prod.yml
gregtatum: Sounds good. I think a paragraph is still fine, but something this big can live in the docs folder.

gregtatum:
CI is hitting #292.

PR #302, opened 2023-12-13
Add a serve-docs make command

gregtatum:
This at least works on macOS if you install `rbenv`, which I don't do in the script. macOS ships with an out of date ruby install by default, which gets in the way, so `rbenv` is a good way to work around it.


docs/_config.yml
```diff
@@ -1,4 +1,5 @@
 remote_theme: just-the-docs/just-the-docs
```
eu9ene, line 1: as far as I understand you replaced the remote github theme to a local theme. Do we still need the "remote_theme" here then?

Makefile
```diff
@@ -193,3 +193,16 @@ download-logs:
 tensorboard:
 	poetry install --only tensorboard
 	marian-tensorboard --offline -f logs/*.log
+
+# Run the GitHub pages Jekyll theme locally.
+serve-docs:
+	echo "This command requires"
+	echo "  rbenv: https://github.com/rbenv/rbenv"
+	echo "  rbenv install 3.2.2"
```
eu9ene, line 201: I think I had some issues with ruby installation on mac os, not sure. This is better than nothing but would it be easier to take some off the shelve docker image and just do docker run here? Maybe something like this https://hub.docker.com/r/bretfisher/jekyll

docs/_config.yml
gregtatum: I think we need both? I have the theme in the `gem` install for local, and for GitHub we can specify the remote one.

Makefile
eu9ene: ok

docs/_config.yml
eu9ene: I'm not sure about this, I didn't find in Github docs how to use a local theme. It says it supports either a "supported theme" or the one hosted on HitHub with `remote-theme`. https://docs.github.com/en/pages/setting-up-a-github-pages-site-with-jekyll/adding-a-theme-to-your-github-pages-site-using-jekyll#adding-a-theme
So adding both might break Github deployment. We can try redeploying docs from your branch.

eu9ene: But this won't work because it's a fork. I guess we can try merging and see what happens.

eu9ene (approved):
Approving, but we might need to remove "theme" setting if it break the deployment.

docs/_config.yml
gregtatum: I'll fast follow if it breaks.

Issue #301, opened 2023-12-13
PRs are not always picking up cached tasks


bhearsum:
This is due to https://github.com/taskcluster/taskgraph/issues/222.

The reason it's been working up until now is because this repository was previously level 3, and we had some older cached tasks sitting in indexes because of that. As those are expiring or become invalid, PRs are losing their ability to find any caches.

We can't immediately bring this repository back up to level 3 (it implies various things that we don't meet at the moment), but we have some other ideas on how to fix this.

PR #303, opened 2023-12-14
do not land: test possible fix for #301

bhearsum:
This pulls in an unreleased version of taskgraph that may fix the issues we're seeing around PRs not pulling cached tasks correctly. (There's also a bunch of unrelated spam in this because that change is based on a new major version of Taskgraph that we haven't upgraded to yet.)

This PR is being opened as an early test of that work - and should not be merged at this time.


bhearsum:
We're pulling in a released version of this work in #316

PR #305, opened 2023-12-14
Improve documentation in transforms

bhearsum:
(And also remove some dead code.)

I've skipped `training_continuation.py` here, because @gabrielBusta is best positioned to write that.


taskcluster/translations_taskgraph/transforms/dependency_dummies.py
```diff
@@ -1,8 +1,21 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# This transform sequence is used to workaround the fact that Taskcluster only
```
gregtatum, line 5: Praise: This is the kind of thing that docs are really nice to have for. Thanks for taking the time for writing this and the others.

gregtatum (approved):
Thanks for these docs, this makes this area of the codebase much easier to understand!

PR #309, opened 2023-12-15
Add publication package

vrigal:
From prototype developed at https://github.com/mozilla/translations-experiment-tracking.

An example of publication to W&B is visible [on Weight & Biases](https://wandb.ai/teklia/projects).

This adds real time logs parsing from Marian (in a TC context) and publication to Weight & Biases.


vrigal:
Should I add some tests for the parser ? Maybe a few "end to end" like tests could be fine (e.g. expecting output from parsing the sample file) ?

gregtatum:
> Should I add some tests for the parser ? Maybe a few "end to end" like tests could be fine (e.g. expecting output from parsing the sample file) ?

Yeah, this sounds great!

eu9ene:
Early piece of feedback. I see that it's not ready for review and I'll review more thoroughly when it is. 

Please rename `publication` to `tracking` which is a term that we use more often (experiment tracking, resource utilization tracking) and move this module from the `pipeline` directory to the root since this is a side functionality and not a part of the core training pipeline. We'll put all the related functionality there.

docs/publication.md
```diff
@@ -0,0 +1,5 @@
+# Metrics publication
```
eu9ene, line 1: this section would be the "Experiment tracking" that can include publication of metrics and other things like logs, configs etc. and details on integration with Taskcluster

pipeline/publication/LICENSE
eu9ene, line 1: we already have a license in the root of the repo

eu9ene:
We should discuss shipping strategy for this. Since it's tied to this code base and not going to be used outside of it I doubt we need a dedicated Python package and versioning. I see that we decided to use it from the same command as training so we'll need a way to make it available. @bhearsum would toolchain be appropriate for this? Or maybe we can leave it as a local package and do pip install right before we run the command? @vrigal Was it a plan? One thing that might be inconvenient with this approach is adding a bunch of fetches to Taskcluster task to install it.

bhearsum:
> We should discuss shipping strategy for this. Since it's tied to this code base and not going to be used outside of it I doubt we need a dedicated Python package and versioning. I see that we decided to use it from the same command as training so we'll need a way to make it available. @bhearsum would toolchain be appropriate for this? Or maybe we can leave it as a local package and do pip install right before we run the command? @vrigal Was it a plan? One thing that might be inconvenient with this approach is adding a bunch of fetches to Taskcluster task to install it.

There's little benefit to using a toolchain considering there isn't any remotely expensive build to do.

If we do `pip install` it, we should make sure we first `pip install` all of the dependencies from a `requirements` file that is version and hash pinned. (The pinning is important to avoid unwanted upgrades breaking things.)

It's also not unreasonable to make this runnable in place without installing - although perhaps we're too far down this road already.

pipeline/publication/translations_parser/cli/task_cluster.py
```diff
@@ -0,0 +1,118 @@
...
+        help="Print debug messages.",
+        action="store_const",
+        dest="loglevel",
+        const=logging.DEBUG,
+    )
+    return parser.parse_args()
+
+
+def task_cluster_log_filter(headers):
+    """
```
bhearsum, line 69: Nit: `taskcluster` is one word.

bhearsum:
Here's a comparable thing we do in firefox-android:
* [A directory with scripts & requirements](https://github.com/mozilla-mobile/firefox-android/tree/main/taskcluster/scripts/lint)
* [Installation and running as part of a task](https://github.com/mozilla-mobile/firefox-android/blob/37ec91c2c65a65dc7a5208c1534e3cbccd00d0dd/taskcluster/ci/lint/buildconfig.yml#L25)

We also have a fair amount of this in Gecko. In these cases, the `mach` wrapper handles all of the package installation and running though, so it's difficult to compare this to what we're doing here.

vrigal:
Thank you @bhearsum for the suggestions about packaging.
I created the pinned requirements file. If I understand correctly I must only use relative imports within parser's submodules to make it work ?

La0:
As discussed today, you can add support for the `speed` metric directly in this PR

vrigal:
@eu9ene I pushed a new version of this patch, supporting metrics (please correct me if it is not `chrf` on 1st line and `bleu_detok` on the 2nd line). It also pushes metrics from `models/<project>/<group>/evaluation/speed` on the fake group `group_logs`. Those values are only displayed in the bar chart actually. I also had to patch several little things while testing the publication.
The current structure have been uploaded to https://wandb.ai/teklia/projects.

Also I don't understand why the linting job fails in the CI (sorting import block), it seems to pass on my side when running `make lint`.

I'm writting simple tests, then the PR should be fine.
I'll work on parsing `experiments/<project>/<group>/config.yml` files on a separated PR (issue #313).

tracking/translations_parser/parser.py
```diff
@@ -0,0 +1,288 @@
...
+        """
+        try:
+            self._parse()
+        except StopIteration:
+            # A StopIteration can be raised if some required lines are never found.
+            raise ValueError("Logs file ended up unexpectedly")
+
+        count = sum(len(vals) for vals in self.indexed_logs.values())
+        logger.info(f"Successfully parsed {count} lines")
+        logger.info(f"Found {len(self.training)} training entries")
```
eu9ene, line 279: Thanks for reporting these stats! It might be valuable to also report how many lines failed to publish (I think the only case when this happen is an error in W&B).

tests/test_tracking_cli.py
```diff
@@ -0,0 +1,169 @@
...
+
+import pytest
+from translations_parser.cli import experiments as experiments_publish
+from translations_parser.cli import taskcluster as tc_publish
+
+"""
+Tests tracking parser and publication via CLI entrypoints
+"""
+
+OUTPUT_DIR = Path(gettempdir()) / ".test_translations_parser_tmpdir"
```
eu9ene, line 18: I believe we settled on using`data/tests` directory for temporary test data (see other tests).

tests/test_tracking_cli.py
eu9ene, line 1: Thank you for adding tests! Can we add a test case when W&B is temporarily unavailable but the training / parsing continues?

eu9ene (changes requested):
Great job overall! I have mostly comments regarding structure, code quality and refactorings and a few functional things that I'd like to be fixed before we merge it. I understand that this code comes from a prototyping repo but since we'll need to maintain it here in the future it's better to invest in typing, readability, comments etc. right away. (here is some inspiration in terms of docs written by Greg: https://github.com/mozilla/firefox-translations-training/blob/main/utils/find_corpus.py)

This code will run in conjunction with the main training procedure which is expensive to stop/break, so we need to make sure it's reliable and handle all the edge cases which you mostly did well already.

eu9ene:
Please sync with the main branch. The main pipeline CI shouldn't run if the "pipeline" and some other related directories are not changed. Only tests and linter should run.

tracking/translations_parser/parser.py
```diff
@@ -0,0 +1,288 @@
...
+    r"Ep\.[ :]+(?P<epoch>\d+)[ :]+"
+    r"Up\.[ :]+(?P<up>\d+)[ :]+"
+    r"Sen\.[ :]+(?P<sen>[\d,]+)[ :]+"
+    r"Cost[ :]+(?P<cost>[\d.]+)[ :]+"
+    r"Time[ :]+(?P<time>[\d\.]+)s[ :]+"
+    r"(?P<rate>[\d\.]+) words\/s[ :]+"
+    r"gNorm[ :]+(?P<gnorm>[\d\.]+)"
+)
+# Expected version of Marian for a clean parsing
+MARIAN_MAJOR, MARIAN_MINOR = 1, 10
```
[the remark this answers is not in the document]
vrigal: I think the parsing logic could remain the same, and different regular expressions would be used depending on the version (default to latest in case Marian context is missing). Thank you for the linking, so this issue is visible.

eu9ene:
> @eu9ene I pushed a new version of this patch, supporting metrics (please correct me if it is not `chrf` on 1st line and `bleu_detok` on the 2nd line). It also pushes metrics from `models/<project>/<group>/evaluation/speed` on the fake group `group_logs`. Those values are only displayed in the bar chart actually. I also had to patch several little things while testing the publication. The current structure have been uploaded to https://wandb.ai/teklia/projects.
> 
> Also I don't understand why the linting job fails in the CI (sorting import block), it seems to pass on my side when running `make lint` (also in docker using task's image).
> 
> I'm writting simple tests, then the PR should be fine. I'll work on parsing `experiments/<project>/<group>/config.yml` files on a separated PR (issue #313).

@vrigal I noticed here https://wandb.ai/teklia/nl-en/runs/8811988u that chrF is smaller than BLEU, it should be the opposite. Usually BLEU is 20-40 and chrF is 40-60.

eu9ene:
> As discussed today, you can add support for the `speed` metric directly in this PR

I suggest adding all the evaluation steps separately in a follow-up PR since we have quite a lot here already. We can track it in this issue https://github.com/mozilla/firefox-translations-training/issues/334

vrigal:
@eu9ene thank you for the review, I applied most of the suggestions and tried to make everything cleaner (also added typing + run mypy, that was an interesting thing to do).

I let some threads opened, because it seemed much easier to work on those once this work is merged. Refactoring things on this huge PR would take a while, the structure has been specified a bit differently for the prototyping part. I can work a bit more on this if you want but I think it would be more efficient iterating on the different topics.

I rebased, but still have trouble making lint and test jobs pass, I'll have a look at this later on.

tracking/translations_parser/publishers.py
```diff
@@ -0,0 +1,137 @@
...
+            self.write_data(training_output, training_log.training, TrainingEpoch)
+
+        validation_output = self.output_dir / "validation.csv"
+        if validation_output.exists():
+            logger.warning(f"Validation output file {validation_output} exists, skipping.")
+        else:
+            self.write_data(validation_output, training_log.validation, ValidationEpoch)
+
+
+class WandB(Publisher):
```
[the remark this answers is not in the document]
vrigal: I agree on this, I did it in a generic way when implementing the CSV export initially. If we keep the API it should use real time publication only (an the inner logic can use close e.g. to write a file), but it could certainly be better. It would also be easier to work on this once this huge PR is merged.

tracking/translations_parser/parser.py
```diff
@@ -0,0 +1,288 @@
...
+
+
+class TrainingParser:
+    def __init__(
+        self,
+        logs_iter: Iterable[str],
+        publishers: Sequence[Publisher],
+        log_filter: Callable = None,
+        skip_marian_context: bool = False,
+        metrics: Sequence[Metric] = None,
```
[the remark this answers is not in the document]
vrigal: I understand, I implemented it this way so W&B does not complain about invalid step (their API is built to handle incremential steps). If I keep the same kind of publication it should be possible to use the last step value to publish all those metrics (and it will eventually make more sense), or publish them in another way.

eu9ene (approved):
Thanks for adding typing and other improvements! I'm approving to unblock other work and let's continue improving things while adding new features. 

You'll have to fix tests and linting to merge it. There is an import error in tests and I see that linting error is quite self-explanatory:
```
[task 2024-01-08T15:47:35.242Z] tracking/translations_parser/cli/experiments.py:9:1: I001 [*] Import block is un-sorted or un-formatted
[task 2024-01-08T15:47:35.242Z] tracking/translations_parser/publishers.py:1:1: I001 [*] Import block is un-sorted or un-formatted
```
You'll likely need to format imports properly. For example, PyCharm does it out of the box. It would be ideal for our formatter to fix this automatically of course. I'm not sure why it doesn't.

PR #316, opened 2023-12-18
Upgrade to Taskgraph 7

### LINKED ISSUE #301: PRs are not always picking up cached tasks

bhearsum:
Two things of note:
* ci directory becomes kinds
* taskcluster/ci/config.yml moves to taskcluster/config.yml

This version also includes some changes to cache routes for PRs which should fix #301.


bhearsum:
`taskgraph --diff` is broken here because when it's trying to generate a task graph with the new taskgraph code, but without the necessary changes to kinds, config.yml, etc. This is, unfortunately, unavoidable (`taskgraph --diff` is unable to downgrade itself when generating the older task graph...)

bhearsum:
This PR is as green as it will get - fails for the reason I mentioned, and the clean steps are failing due to #292

ahal (approved):
Fyi I've also been updating to the latest [Decision image](https://hub.docker.com/layers/mozillareleases/taskgraph/decision-c4ac262880970ca484105929e02dd12f00214d5f2603ab6ce4c0d17de5cd0280/images/sha256-2dd667994aa13fccdcdcede85c570a5eb4a5247e42875a9e95a135ef774ee469?context=explore) as part of these upgrades. But it's totally reasonable to do that separately.

Issue #270, opened 2023-11-20
Support training continuation for the failed or preempted tasks


eu9ene:
It's especially important for pre-emption of spot instances

bhearsum:
The work that @gabrielBusta is doing in #226 will be a good basis for this. The difference with spot terminations is that the tasks will automatically rerun, and we won't be able to adjust the parameters. We'll need some sort of enhancement to detect that case and automatically find the previous attempt's artifacts. Eg: a check at the start of the job if this is run#1 or higher, and if it is, attempt to pull artifacts from the previous run. (We might want more sanity checking in there, too.)

gregtatum:
(copying over my thoughts from #315).

For reference, this is the definition of a [preemtible instance](https://cloud.google.com/compute/docs/instances/preemptible).

During the Catalan run the teacher training would often take 2 or 3 times as long to run to completion since the task would get preempted. Here is an example profile of several preemption tasks happening: https://share.firefox.dev/3Rw5u5g

Also, I believe @bhearsum said that this was a blocker for this work: https://mozilla-hub.atlassian.net/browse/RELOPS-782

marco-c:
https://mozilla-hub.atlassian.net/browse/RELOPS-782 is in progress (https://github.com/mozilla-platform-ops/monopacker/pull/121).

bhearsum:
As of today, all of the instances we use in Taskcluster should handle spot terminations gracefully, and publish whatever artifacts exist at the time of the shutdown.

gregtatum:
The next step here is to load in the previous artifacts and restart the training.

gabrielBusta:
This should be do-able (in theory) since the training artifacts should be publish when an instance is prevented. AIUI, as long as the task has the training artifacts [using pretrained models](https://mozilla.github.io/firefox-translations-training/using-pretrained-models.html) should work.

bhearsum:
> This _should_ be possible now because the training artifacts are being published when an instance is preempted. AIUI, as long as the task produces the training artifacts [using pretrained models](https://mozilla.github.io/firefox-translations-training/using-pretrained-models.html) should work. I wish there was a way we could try it out, but how does one simulate a spot termination?

I learned recently that you can just press the "stop" button in the GCP console to do this :). (I believe you have the necessary access to do so.)

Issue #200, opened 2023-09-20
NetworkError while fetching log


eu9ene:
https://firefox-ci-tc.services.mozilla.com/tasks/F3k0tOoOQMaf9q_vJYqFAg/runs/0/logs/live/public/logs/live.log

<img width="1498" alt="Screenshot 2023-09-20 at 1 18 00 PM" src="https://github.com/mozilla/firefox-translations-training/assets/2486505/a5346c7c-bea5-4103-a5b4-71ba440963e1">

I see this error for the second time for a teacher training task. Last time it didn't go away until the training had finished. This blocks us from monitoring the training progress.

gabrielBusta:
I see a `CORS` error in the devtools console:

```
GET
https://firefoxci-websocktunnel.services.mozilla.com/us-central1.5838805491402473905.60099/log/cBTzdzw7Q6e29PZJADQ-bA
CORS Missing Allow Origin

Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://firefoxci-websocktunnel.services.mozilla.com/us-central1.5838805491402473905.60099/log/cBTzdzw7Q6e29PZJADQ-bA. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing). Status code: 504.
```

When check the URL to see if it is reachable

https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/F3k0tOoOQMaf9q_vJYqFAg/runs/0/artifacts/public%2Flogs%2Flive.log

It redirects me to the same URL I see in the console error

https://firefoxci-websocktunnel.services.mozilla.com/us-central1.5838805491402473905.60099/log/cBTzdzw7Q6e29PZJADQ-bA

I see the message:
```
No client is connected with that id
```

marco-c:
The CORS issue might be due to the fact that the response seems to be empty (and so CORS headers are missing).
There is a 504 from https://firefoxci-websocktunnel.services.mozilla.com/.

eu9ene:
This might be related. I tried to look at logs for this task https://firefox-ci-tc.services.mozilla.com/tasks/ZHsI6EkhS1efoPX7Hd3niA, got a similar network error and the task restarted. Maybe it restarted even before that but didn't update its status in UI, not sure.

eu9ene:
[...]

PR #320, opened 2023-12-20
do not land: try out new gpu worker changes

### LINKED ISSUE #271: Logs disappear after a task fails

bhearsum:
This may fix #271, and possibly help with #200 and #270. This PR is just to sanity check that nothing obvious is broken with the worker image.


bhearsum:
(The new image looks good - I'll be swapping it in for all of the GPU workers in the new year.)

PR #326, opened 2023-12-21
Fix the vocab training script for Taskcluster

gregtatum:
I got my Catalan training into a broken state since the word "None" was being passed in as the vocab size. The first commit fixes that. I spent some time writing a test to have some confidence over my changes. While doing that I also created a shared fixture utility called `DataDir` that automatically handles some of the test data generation and deleting tasks. This way we can be consistent across the tests. Each commit can be reviewed by itself as they are organized to make sense individually.


tests/fixtures/spm_train
eu9ene, line 1: It's an interesting idea to replace Marian and other binaries with such stubs. We don't get a real run but this would likely require building a docker file and take longer which is not great for unit tests. Let's assume Marian etc is tested in CI and here we test only the scripts for now.

eu9ene (approved):
Looks good but you need to fix linting

tests/fixtures/spm_train
gregtatum: At some point I'd like to understand the local docker situation a bit more. This at least runs all of our script and sees that we're calling out to third part libraries the way we expect. I could see doing the same thing with a real sentence piece binary and asserting on the results.

PR #335, opened 2024-01-04
Add downloading evals from Taskcluster

eu9ene:
I needed to analyze my recent experiment and there are too many tasks to look at manually, so I extended the logs downloading script to download results from the evaluation tasks and export them to a CSV.

Results are in [this doc](https://docs.google.com/spreadsheets/d/1wKdbPN8XzAmAaVJMwYENWb1JvOQe3s3ScRy8Kgz1LhQ/edit?usp=sharing)


utils/tc_marian_logs.py
```diff
@@ -1,16 +1,49 @@
...
+from taskcluster.download import downloadArtifactToBuf, downloadArtifactToFile
+
+TC_MOZILLA = "https://firefox-ci-tc.services.mozilla.com"
+
+# to parse evaluation task tag
+# examples:
+# evaluate-teacher-flores-flores_aug-title_devtest-lt-en-1_2
+# evaluate-teacher-mtdata-mtdata_aug-mix_Neulab-tedtalks_test-1-eng-lit-lt-en-1_2
+# evaluate-finetuned-student-sacrebleu-wmt19-lt-en
+eval_regex = re.compile(
```
eu9ene, line 21: ideally we should add metadata to Taskcluster tasks, so that we don't need to parse the tag which can be quite different for different tasks

utils/tc_marian_logs.py
```diff
@@ -41,6 +74,66 @@ def donwload_logs(group_id, output):
...
+            artifact["name"] for artifact in artifacts if artifact["name"].endswith(".metrics")
+        ][0]
+
+        print(f"Downloading {artifact_name} for {task_obj_label}")
+        content, _ = downloadArtifactToBuf(
+            taskId=task["status"]["taskId"],
+            name=artifact_name,
+            queueService=queue,
+        )
+        bleu, chrf, _, _ = content.tobytes().decode().split("\n")
```
gregtatum, line 109: Thought: Ah the mystery of these numbers explained.

utils/tc_marian_logs.py
```diff
@@ -41,6 +74,66 @@ def donwload_logs(group_id, output):
...
+        suffix = groups.get("suffix", "") or ""
+
+        result = (model + suffix, f"{importer}_{dataset}", augmentation, bleu, chrf)
+        print(f"Result: {result}")
+        results.append(result)
+
+    os.makedirs(output, exist_ok=True)
+    output_path = os.path.join(output, f"{group_id}-metrics.csv")
+    with open(output_path, "w") as csvfile:
+        spamwriter = csv.writer(csvfile)
```
gregtatum, line 131: question: What's a spamwriter?

utils/tc_marian_logs.py
```diff
@@ -1,16 +1,49 @@
 """
```
gregtatum, line 1: Suggestion (optional): `tc_marian_logs.py` is not really a good name for this file now, as it's more of a generic downloader.

utils/tc_marian_logs.py
gregtatum: Yeah, this is not very maintainable, and will be really brittle.

gregtatum (approved):

utils/tc_marian_logs.py
eu9ene: it got here by accident :)

Issue #281, opened 2023-11-29
quantize often dumps core since browsermt-marian was updated


bhearsum:
In a few PRs that don't touch the pipeline at all, I've had quantize dump core on me with:
```
[task 2023-11-29T15:40:38.837Z] /builds/worker/checkouts/vcs/pipeline/quantize/quantize.sh: line 41:   140 Aborted                 (core dumped) "${BMT_MARIAN}"/marian-decoder -m "${model}" -v "${vocab}" "${vocab}" -c "decoder.yml" -i "${devtest_src}" -o "${output_dir}/output.${TRG}" --shortlist "${shortlist}" false --quiet --quiet-translation --log "${output_dir}/cpu.output.log" --dump-quantmult 2> "${output_dir}/quantmults"
```

I thought it might be similar to #249, but the vocab size is still a multiple of 8. When I removed `--quiet` to try to get more information, it got past the core dump, but `extract_stats.py` failed. (I presume this is because it caused `quantmults` to get information that the script can't parse.)

It certainly seems like `--quiet` is impacting the behaviour, although that seems...surprising.

bhearsum:
For the record, [here's a diff of the changes in browsermt-marian](https://github.com/browsermt/marian-dev/compare/08b1544636fe13eaf1fbacb17c6fb050abfb8d42..master) that we took.

gregtatum:
I'm wondering if we're taking this new codepath with the ruy quantize:

https://github.com/browsermt/marian-dev/blob/11c6ae7c46be21ef96ed10c60f28022fa968939f/src/tensors/cpu/ruy_interface.h#L34-L79

eu9ene:
I actually tried to update it after it started crashing. It seems it didn't help and it randomly fails now. I think training procedure and as a result the models changed a little, which might cause different weights in the model that's why it fails on CI (maybe some 0s). I tried to extend training a bit and I thought it helped but maybe it didn't and there is still a change of crashing. It seems this issue happens only on CI.

gregtatum:
I don't think this happens anymore.

gregtatum:
I think this is fixed by #1169.

PR #341, opened 2024-01-08
Upgrade Taskgraph and cleanup firefox-translations-training

gabrielBusta:
I updated `taskgraph` to get rid of the in-tree copies of the core `docker_image` and `fetch` transforms. I also set a working default for `task-expires-after` so we only need to specify it when it is not `90 days`.


gabrielBusta:
Blah it just occurred to me that this is gonna invalidate a lot of cached tasks.

eu9ene:
> Blah it just occurred to me that this is gonna invalidate a lot of cached tasks.

We change the code so often that the caches don't survive between the experiments. I usually stick with one branch without merging from main to complete the experiment and reuse caches.

eu9ene (approved):

bhearsum (approved):
Thanks! It looks like this produces identical task graphs.

gabrielBusta:
mm `(core dumped) "${BMT_MARIAN}"/marian-decoder`

bhearsum:
> mm `(core dumped) "${BMT_MARIAN}"/marian-decoder`

That looks like https://github.com/mozilla/firefox-translations-training/issues/281

PR #342, opened 2024-01-09
fix Dataset importer problems

### LINKED ISSUE #338: Dataset importer problems
The dataset importer introduced a couple months ago has two issues:

1. https://github.com/mozilla/firefox-translations-training/blob/main/Snakefile#L281 calls it using `bash`, not `python` (then the errors are related to bash not recognizing "import" etc)
2. `ValueError: Invalid dataset name: custom-corpus_/corpora/parallel/test/all. Use the following format: <importer>_<name> or <importer>_<augmentation>_<name>.` This is because the regex checked doesn't allow for dashes. The correct regex is `^([\w-]*)_(aug[a-z\-]*)?_?(.+)$`

AmitMY:
fix #338


gregtatum (approved):
Thanks, looks good to me.

gregtatum:
Oh wait, the tests aren't passing. We should investigate that before merging.

gregtatum (changes requested):
It looks like this is causing real test failures. I'm requesting changes since I believe this is breaking things.

`sacreBLEU: Unknown test set 'aug-mix_wmt19'`

AmitMY:
Seems like `_` is a word character (`\w`), let's try with it optional.  We'll see if there's another issue in the tests

gregtatum (approved):
CI looks good! Let's wait until #362 merges, then I can rebase and merge. Thanks for the fix.

gregtatum (changes requested):
It looks like CI is still failing here.

AmitMY:
It did pass before you merged master again, iirc

gregtatum:
There is a red X next to your previous pushes, see: https://github.com/mozilla/firefox-translations-training/runs/20456494808

That looks like an intermittent failure, but it never got to the eval steps.

marco-c:
@AmitMY are you still using the pipeline?

AmitMY:
> @AmitMY are you still using the pipeline?

no

Issue #306, opened 2023-12-14
add documentation to training continuation transform


bhearsum:
We've added documentation to most transforms in #305. @gabrielBusta - could you please do the same for training continuation?

eu9ene:
Like a dupe of #299 but I'm keeping them both in case something else was meant here @gabrielBusta. We need at least user facing docs on how to use it and maybe more technical details on how it works.

Issue #299, opened 2023-12-12
Document training continuation


gregtatum:
#226 is now merged, but we still need docs for how to do it.

gregtatum:
Resolved with #349.

PR #349, opened 2024-01-09
[skip ci] add docs on `pretrained-models` configuration parameter

gabrielBusta:


gabrielBusta:
#299 #306

docs/using-pretrained-models.md
```diff
@@ -0,0 +1,55 @@
...
+
+### The Mode Key
+
+#### Use Mode
+
+In `use` mode, the pipeline only downloads the model without further training. The tasks that depend on the training task will use the downloaded model artifacts as they are.
+
+#### Continue Mode
+
+In `continue` mode the pipeline uses the downloaded model artifacts from the previous training run as a "checkpoint" and continues training.
```
eu9ene, line 47: I think it's worth mentioning how continue is different from init. The difference in usage is mostly training the same model on the same corpus versus fine-tuning an existing model on a different corpus.

eu9ene (approved):

PR #352, opened 2024-01-10
Fix unstable training

### LINKED ISSUE #314: Unstable training with OpusTrainer
It seems one of the training stages in OpusTrainer (the one that includes back-translations) reduces performance of the model. Probably it's because now we start training with the original corpus and then switch to the mixed one. Maybe it's not the best idea because the training will likely stop with early-stopping and we can't control training parameters separately for now. Related to #293 

For now we can either disable back-translations completely until proving in an experiment that they help or try changing the stages to something like a little of mixed back-translations first and then fine-tuning on the original corpus + increasing early-stopping to 30 or 40 from the default 20.

Current config:

```
datasets:
  original: <dataset0> # Original parallel corpus
  backtranslated: <dataset1> # Back-translated data

stages:
  - start
  - mid
  - end
  - finetune

 # One epoch of only original high-quality data to warm up the model
start:
  - original 1.0
  - until original 1

# Gradually add back-translations to the mix
# Back-translated corpus can vary a lot in size, so we can try using original to count epochs
mid:
  - original 0.7
  - backtranslated 0.3
  - until original 1

# Expand back-translations
end:
  - original 0.6
  - backtranslated 0.4
  - until original 1

# Fine-tuning only on original clean corpus until the early stopping
finetune:
  - original 1.0
  - until original inf
```
Graphs for en-hu:
Screenshot 2023-12-18 at 12 24 05 PM

eu9ene:
- Fix hyperparameters
- Adjust OpusTrainer settings

Training looks stable with fixed optimizer-delay.
Reducing learning rate didn't affect much.
Adding back-translations gives significant boost (top pink) but I used lower early stopping for this one to finish faster (20 vs 30 for other experiments).

Screenshot 2024-01-12 at 11 21 25 AM

Fixes #314 

[skip ci]


pipeline/train/configs/opustrainer/teacher.yml
```diff
@@ -3,28 +3,14 @@ datasets:
...
-# Back-translated corpus can vary a lot in size, so we can try using original to count epochs
-mid:
-  - original 0.7
-  - backtranslated 0.3
-  - until original 1
-
-# Expand back-translations
-end:
+# Back-translated corpus can vary a lot in size, so we can try using original one to count epochs
+pretrain:
```
eu9ene, line 10: instability was not caused by stages so it's fine to use a simpler schedule

taskcluster/kinds/tests/kind.yml
```diff
@@ -22,38 +22,39 @@ task-defaults:
...
-          make conda                                                        &&
-          make snakemake                                                    &&
-          make git-modules                                                  &&
-
-          echo "Start the dry run"                                          &&
-          make dry-run                                                      &&
-          make test-dry-run
+  # See issue: https://github.com/mozilla/firefox-translations-training/issues/363
+  #  snakemake-dry-run:
+#    # Ensure that the snakemake workflow is still executing correctly, even though
```
gregtatum, line 27: Question: Is this related to your changes?

taskcluster/kinds/tests/kind.yml
```diff
@@ -22,38 +22,39 @@ task-defaults:
...
+#    # taskcluster is the preferred execution environment.
```
gregtatum, line 28: Thought: This kind of change would be better in a separate PR. I'd prefer to merge with a dirty CI, and fast-follow with another PR. This makes it clearer then things go wrong. For now this change is fine.

pipeline/train/configs/training/teacher.train.yml
```diff
@@ -1,9 +1,10 @@
...
+optimizer-delay: 2 # Roughly GPU devices * optimizer-delay = 8, but keep as an integer
+lr-report: True
 save-freq: 5000
-valid-freq: 3000
+valid-freq: 5000
 valid-max-length: 300
 valid-mini-batch: 8
-early-stopping: 20
\ No newline at end of file
+early-stopping: 20
```
gregtatum, line 10: I spent a bit of time researching these various options to see what you changed to fully understand what was going on. Rather than doing that I have a suggestion that should be quick:

Suggestion (docs): It would be nice to include a short message explaining why you chose certain values here. It would be nice to document decisions when we change hyperparameters. This will make it easier to share our knowledge with each other and remember things for our future selves.

pipeline/train/configs/opustrainer/teacher.yml
```diff
@@ -33,8 +19,8 @@ finetune:
 
 
 modifiers:
-- UpperCase: 0.1 # Apply randomly to 10% of sentences
-- TitleCase: 0.1
+- UpperCase: 0.07 # Apply randomly to 7% of sentences
+- TitleCase: 0.05
```
gregtatum, line 23: Question: Why are you changing these here?

pipeline/train/configs/opustrainer/teacher.yml
gregtatum: Ah, I like this as being simpler.

pipeline/train/configs/training/teacher.train.yml
```diff
@@ -1,9 +1,10 @@
-# https://discourse.translatelocally.com/t/marian-configuration-to-use/24
 disp-freq: 1000
+# default learning rate for transformer-big is 0.0002 https://github.com/marian-nmt/marian-dev/blob/master/src/common/aliases.cpp
 learn-rate: 0.0003 # Turn this down if you get a diverged model, maybe 0.0001
-optimizer-delay: 1 # Roughly GPU devices * optimizer-delay = 8, but keep as an integer
+optimizer-delay: 2 # Roughly GPU devices * optimizer-delay = 8, but keep as an integer
```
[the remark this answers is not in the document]
gregtatum: I guess this is matching the `2` value from the `pretrain` step? If so please document this both here and in the pretrain step.

gregtatum (approved):
Looks good as long as there are a few more docs before merging. Thanks for looking into this.

pipeline/train/configs/training/teacher.train.yml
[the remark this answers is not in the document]
eu9ene: I did not find any resources where this recommendation comes from, but it seems it increases the update batch size that makes training more stable.


Me:
Why is it recommended for marian to set optimizer-delay so that GPU devices * optimizer-delay = 8?

ChatGPT:
The recommendation to set optimizer-delay so that GPU devices * optimizer-delay = 8 in Marian NMT is a heuristic that aims to balance the computational efficiency and the stability of the optimization process. This heuristic is derived from empirical observations that suggest this setting helps to achieve a good trade-off between the speed of computation and the stability of convergence, particularly when using multiple GPUs.

This setting effectively increases the batch size without requiring more memory, which can lead to more stable and reliable gradient estimates. It is a way to utilize the parallelism offered by multiple GPUs while also ensuring that each update is significant enough to provide stable learning, without being so large that it might cause instability due to the accumulation of too much gradient information before an update is applied.

taskcluster/kinds/tests/kind.yml
eu9ene: I just want CI to pass here and disabled this since it's related only to snakemake.

pipeline/train/configs/opustrainer/teacher.yml
eu9ene: Based on the paper https://arxiv.org/pdf/2311.14838.pdf. They set them to 0.05 but I noticed that title case performs better than upper case so I boosted it a bit. Also I ran an experiment and got satisfactory results. I added a link to the docs.

PR #353, opened 2024-01-11
Add a preflight check utility

gregtatum:
When I was working on Catalan, I had trouble understanding what was going to happen during training. This utility adds the ability to see what tasks will be generated, and with what arguments. It generates the full artifacts folder locally for inspection.

The artifacts are generated in the root directory of the project. I would prefer `data/artifacts` but I can't see a way to configure it that way.

image of the taskcluster tasks logged

the training config

The artifacts

the graph visualization


gregtatum:
Oof, I need a pre-push hook for linting and tests.

Issue #340, opened 2024-01-08
`split-mono` and `split-corpus` no longer exist


AmitMY:
https://github.com/mozilla/firefox-translations-training/blob/main/Snakefile#L443
 and
https://github.com/mozilla/firefox-translations-training/blob/main/Snakefile#L576
fail: `No such file or directory`

(I am aware that the snakemake implementation is no longer actively maintained, but I believe fixing these small things can allow more people to use your pipeline, and ideally contribute. I am not sure what you changed in this regard)

eu9ene:
Yes, we're de-prioritizing Snakemake for now until we stabilize the pipeline under Taskcluster. We'll see how many people still use it and might fix compatibility issues there in the future. Meanwhile you can use GreenNLP's [fork](https://github.com/GreenNLP/firefox-translations-training) which brings ability to use pre-trained OPUS-MT models as teachers or contribute a PR to fix the issues you've found. Ideally I would like to keep Snakemake working to make the pipeline usable for folks outside of Mozilla.

AmitMY:
Unlike #339 and #338, I don't know how to solve this specific issue, and get the splitter to run correctly.

AmitMY:
What I tried:
Replace:
```bash
bash pipeline/translate/split-mono.sh {input.corpora} {output} {split_chunks} >> {log} 2>&1
```
with 
```bash
python pipeline/translate/splitter.py \
                --output_dir={output} \
                --num_parts={split_chunks} \
                --compression_cmd=gzip \
                {input.corpora} >> {log} 2>&1
```

Which uses `compression_cmd` as `gzip` as implied as supported: https://github.com/mozilla/firefox-translations-training/blob/d35f28e542c85166368caa194d722954d9728197/pipeline/translate/splitter.py#L81

But it is not actually supported, because `gzip` has no `--rm` command

AmitMY:
Now splitting works (with #361)

And here is my current log for a new job.
It seems to be trying to `Collecting translated mono src dataset` before training a model, or running inference on it, which leads to 0 translations

```log
Running with config configs/config.spoken-to-signed.yml and profile local
Building DAG of jobs...
Using shell: /bin/bash
Provided cores: 40
Rules claiming more threads will be scaled down.
Provided resources: gpu=4
Singularity containers: ignored
Job stats:
job                    count    min threads    max threads
-------------------  -------  -------------  -------------
alignments                 1             40             40
all                        1              1              1
ce_filter                  1             40             40
[...]

PR #361, opened 2024-01-12
fix(splitter): properly support gzip

AmitMY:
splitter does not support gzip, related to #340 

and I would appreciate help for the rest of the issue.


eu9ene:
> splitter does not support gzip, related to #340
> 
> and I would appreciate help for the rest of the issue.

I think we never supported gzip in the pipeline. The alternative compression tool to `zstd` is `pigz` which is parallel gzip and a lot faster on big datasets. Please check if it works for you.

AmitMY:
The code here says `gzip` as a possible compression command:
https://github.com/mozilla/firefox-translations-training/pull/361/files#diff-37d52397e2d0d58bc72c90f12431a82e8a0cb06e5fc7fe34a3c139a5d82148c8R89
In any case, `pigz` also does not support `--rm`, so I updated the code to support it

pipeline/translate/splitter.py
eu9ene, line 1: We should just do `os.remove` to unify the behavior across the differnet tools or even better 
```
if os.path.isfile(current_name):
  os.remove(current_name)
```

The idea is not to keep compressed and uncompressed files on disk since the can be huge

eu9ene (changes requested):

pipeline/translate/splitter.py
AmitMY: done

pipeline/translate/splitter.py
```diff
@@ -18,6 +18,15 @@
 from typing import Optional
 
 
+def compress(compression_cmd: str, file_path: str):
+    print(f"Compressing {file_path} with {compression_cmd}")
+    subprocess.run([compression_cmd, file_path], check=True)
+
+    # gzip and pigz remove the file by default, but zstd does not
```
eu9ene, line 25: thanks for the comment, now it's clear what's going on here

eu9ene (approved):

eu9ene:
@AmitMY could you commit a small change to retrigger CI? I disabled it temporarily no to run the whole pipeline on each commit and now enabled again.

eu9ene:
Also snakemake dry-run got broken and now disabled so it might be worth looking into https://github.com/mozilla/firefox-translations-training/issues/363

PR #362, opened 2024-01-12
Make CI happy again

### LINKED ISSUE #365: Tracking tests randomly fail
For example in this PR where the changed code is not related to the tests https://github.com/mozilla/firefox-translations-training/pull/361/checks?check_run_id=20439578862

gregtatum:
Fixes #365

This PR is fixing a whole slew of issues with the CI.

I was sure if ruff was working because of this `--no-root` issue.

```
[task 2024-01-11T21:16:22.015Z]   • Installing wandb (0.16.1)
[task 2024-01-11T21:16:22.356Z]   • Installing ruff (0.0.287)
[task 2024-01-11T21:16:22.695Z]   • Installing translations-parser (0.1.0 /builds/worker/checkouts/vcs/tracking)
[task 2024-01-11T21:16:24.195Z] 
[task 2024-01-11T21:16:24.196Z] Installing the current project: firefox-translations-training (0.1.0)
[task 2024-01-11T21:16:24.197Z] 
[task 2024-01-11T21:16:24.197Z] The current project could not be installed: No file/folder found for package firefox-translations-training
[task 2024-01-11T21:16:24.197Z] If you do not want to install the current project use --no-root
[taskcluster 2024-01-11 21:16:25.264Z] === Task Finished ===
[taskcluster 2024-01-11 21:16:25.264Z] Successful task run with exit code: 0 completed in 18.569 seconds
```

And then pytest-clarity wasn't adding clarity due to colors not being displayed.

I also updated ruff since I was getting local errors.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -121,17 +121,17 @@ dag:
 # OpusCleaner is a data cleaner for training corpus
 # More details are in docs/cleaning.md
 opuscleaner-ui:
-	poetry install --only opuscleaner
+	poetry install --only opuscleaner --no-root
 	opuscleaner-server serve --host=0.0.0.0 --port=8000
 
 # Utils to find corpus etc
 install-utils:
-	poetry install --only utils
+	poetry install --only utils --no-root
 
 # Black is a code formatter for Python files. Running this command will check that
 # files are correctly formatted, but not fix them.
 black:
-	poetry install --only black
+	poetry install --only black --no-root
 	@if poetry run black . --check --diff; then \
 		echo "The python code formatting is correct."; \
 	else \
@@ -144,17 +144,17 @@ black:
 
 # Runs black, but also fixes the errors.
 black-fix:
-	poetry install --only black
+	poetry install --only black --no-root
 	poetry run black .
 
 # Runs ruff, a linter for python.
 lint:
-	poetry install --only lint
+	poetry install --only lint --no-root
 	poetry run ruff check .
 
 # Runs ruff, but also fixes the errors.
 lint-fix:
-	poetry install --only lint
+	poetry install --only lint --no-root
 	poetry run ruff check . --fix
 
 # Fix all automatically fixable errors. This is useful to run before pushing.
@@ -164,7 +164,7 @@ fix-all:
 
 # Run unit tests
 run-tests:
-	poetry install --only tests --only utils
+	poetry install --only tests --only utils --no-root
 	PYTHONPATH=$$(pwd) poetry run pytest tests -vv
 
 # Validates Taskcluster task graph locally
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -39,7 +39,6 @@ requests="2.26.0"
 pytest="7.4.3"
 # use the latest main, switch to PyPi when released
 opustrainer = {git = "https://github.com/hplt-project/OpusTrainer.git", rev="9133e1525c7ee37f53ea14ee6a180152bf7ea192"}
-pytest-clarity = "^1.0.1"
```
gregtatum, line 42: I thought this would give us nicer diffs, but in CI you don't get the +-. Locally you get red/green colors, but in CI there is no color output.

Makefile
```diff
@@ -150,6 +150,7 @@ black-fix:
 # Runs ruff, a linter for python.
 lint:
 	poetry install --only lint --no-root
+	poetry run ruff --version
```
gregtatum, line 153: ruff doesn't output to stdout on success, so this at least gives an indication of what is going on, and what version is running.

tests/test_tracking_cli.py
```diff
@@ -147,14 +156,19 @@ def test_experiments(wandb_mock, getargs_mock, caplog, samples_dir, tmp_dir):
     with (samples_dir / "experiments_wandb_calls.json").open("r") as f:
         assert log_calls == [call(**entry) for entry in json.load(f)]
     # Custom calls for .metrics files publication
-    assert [list(v.keys())[0] for c in metrics_calls for v in c.args] == [
+    expectations = [
```
eu9ene, line 159: assert is a way to go in pytest, let's investigate what's wrong with it and fix it so that it works https://docs.pytest.org/en/7.1.x/how-to/assert.html

eu9ene: alternatively you can disable the test and leave #365 for @vrigal  to investigate

gregtatum: `assert` works just fine. The issue is that the log ordering is different in CI, it seems like there is some kind of race condition in the logging. It doesn't look like an error in the code, just an error in the test assertion. I raised an Exception in what I changed since it seemed to give more salient failures.

I'm fine changing it to a more idiomatic `assert` if that's your preference.

eu9ene: yeah, I suggest switching to asserts since I think pytest might do some tricks there. If the order is not important you can use python sets instead of lists. If it is important, there's a bug.

gregtatum:
Also I would prefer if you could block the PR with "request changes" if you leave blocking feedback like in https://github.com/mozilla/firefox-translations-training/pull/362#discussion_r1450901493

Otherwise in my review dashboards it looks like you haven't given any feedback yet. I have to manually go back and read the PR. Giving a comment without approving is functionally equivalent to requesting changes. However, with actually changing the status it's clear on my end what needs to happen.

tests/test_tracking_cli.py
gregtatum: Ah ok, asserting on `set`s is much cleaner, thanks for the suggestion. I've updated the tests.

eu9ene (approved):

Issue #337, opened 2024-01-04
Evaluation results should be saved as JSON or YAML


gregtatum:
Right now they are formatted as `f"{bleu}\n{chrf}"` which is not very helpful. They should be output to include all of the information deduced by:

[utils/tc_marian_logs.py](https://github.com/mozilla/firefox-translations-training/pull/335/files#diff-b0e7241806fe6393f3d20c5f0a4f6b31a51c79f155f003f458eb69f563f25262) (See pull #335).

PR #364, opened 2024-01-12
Add tests for evaluation

gregtatum:
This is in preparation for #337.

I modified the bash scripts slightly by expanding 1 letter flags to their full names to help in clarity, and I added documentation.


gregtatum:
The test failure is #365.

eu9ene (commented):
I'm trying to understand what we are testing here. Is it just that command args are propagated correctly? A bug here is possible but it's a very small error surface compared to the whole evals process with all the integrations. Can we at least use the real sacrebleu instead of a stub? It will be fast on generated data and then we'll be able to test that its outputs are in the right format, so it will improve coverage. Otherwise those tests are less useful but they still add to code maintainace.

gregtatum:
> I'm trying to understand what we are testing here. Is it just that command args are propagated correctly? A bug here is possible but it's a very small error surface compared to the whole evals process with all the integrations.

For me I have little confidence touching these scripts as it's hard to verify quickly that if I change anything to know that I have any errors in the changes. These are more at the unit level compared to where I normally like to test. I do wonder if we can write tests directly from the taskgraph that is generated. I think this would be more useful at more of the integration level. I'll have to think on this more. In the meantime, I would like to have some kind of way to run this code locally even if it does add somewhat to the maintenance burden.

> Can we at least use the real sacrebleu instead of a stub? It will be fast on generated data and then we'll be able to test that its outputs are in the right format, so it will improve coverage. Otherwise those tests are less useful but they still add to code maintenace.

I didn't realize that `sacrebleu` could be pip installed. It would be much better to run that rather than invent a test fixture here! I'll update the code to use that.

gregtatum:
Ok, I found a way to test more end-to-end using the taskgraph, although I would prefer to land this, and then follow-up, as it requires a for more changes to the test infrastructure.

The task-level API would look something like this:

```py
    run_task(
        task_name="evaluate-quantized-sacrebleu-wmt09-en-ru",
        script="pipeline/eval/eval-quantized.sh",
        work_dir=test_data_dir.path,
        fetches_dir=test_data_dir.path,
        env={
            # This is where the marian_decoder_args will get stored.
            "TEST_ARTIFACTS": test_data_dir.path,
            # Replace marian with the one in the fixtures path.
            "BMT_MARIAN": fixtures_path,
            # This is included via the poetry install
            "COMPRESSION_CMD": "zstd",
        },
    )
```

gregtatum:
I have the follow-up here as draft, and it contains the run_task code: ecd27933ca12a6ec973bce31e1497e5e52097a3b 
https://github.com/mozilla/firefox-translations-training/pull/393

PR #374, opened 2024-01-19
Parse missing evaluation results

### LINKED ISSUE #334: Add missing steps for experiment tracking
I noticed that we currently publish only the steps that have training procedure for. We should also publish the results from evaluation steps to separate W&B runs for easy comparison between all the models:
- teacher-ensemble (`evaluate-teacher-ensemble-<dataset>` for Taskcluster: [example](https://firefox-ci-tc.services.mozilla.com/tasks/groups/EW7qV3U5SBSjegnTxGZHkw#evaluate-teacher-ensemble))
- quantized (`speed` directory for old experiments, `evaluate-quantized-<dataset>` results for Taskcluster: [example](https://firefox-ci-tc.services.mozilla.com/tasks/groups/EW7qV3U5SBSjegnTxGZHkw#evaluate-quantized))

Also keep in mind that sometimes we'll use pre-trained models (see #226) and will not run training at all for some steps (likely backward and teacher) but only run evaluation.

### LINKED ISSUE #336: Publish all evaluation results on experiment dashboard
There should be a way to look at all the evaluation and compare them across models and experiments. Let's think what would be the most convenient way to do that. We already publish some of them inside each step and have a dashboard to compare (it's not clear what values are those though). Let's see what's missing.

Having [such table](https://docs.google.com/spreadsheets/d/1wKdbPN8XzAmAaVJMwYENWb1JvOQe3s3ScRy8Kgz1LhQ/edit#gid=66711953) for the whole experiment would be very convenient for analysis, so we should either add an extra table to the `group_logs` step or to the overview for the group.

Screenshot 2024-01-03 at 4 36 48 PM

Related to #334

vrigal:
Closes #334 \
Closes #336

Displays a table will all evaluation metrics on `group_logs`.
Metrics by run are still published individually.

Results have been updated on https://wandb.ai/teklia/projects. \
I still need to update tests on this PR.

### DIFF
--- a/tracking/translations_parser/cli/experiments.py
+++ b/tracking/translations_parser/cli/experiments.py
@@ -55,7 +55,7 @@ def parse_experiment(
     metrics = []
     if metrics_dir:
         for metrics_file in metrics_dir.glob("*.metrics"):
-            metrics.append(Metric.from_file(model_name=name, metrics_file=metrics_file))
+            metrics.append(Metric.from_file(model_name=f"{group}.{name}", metrics_file=metrics_file))
 
     with logs_file.open("r") as f:
         lines = (line.strip() for line in f.readlines())
@@ -103,7 +103,7 @@ def publish_group_logs(
     # Add "quantized" metrics
     metrics = []
     for file in quantized_metrics:
-        metrics.append(Metric.from_file("quantized", file))
+        metrics.append(Metric.from_file(f"{group}.quantized", file))
     # Add experiment (runs) metrics
     for file in evaluation_metrics:
         model_name = file.stem.lstrip("eval_")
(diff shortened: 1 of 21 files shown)


vrigal:
@eu9ene I updated the script according to what was discussed yesterday. \
I added runs for metrics that have no training steps, and specified a row with group name on the `group_logs` table containing all metrics.
It was not possible to prefix runs with the group as we suggested, so comparing runs with similar names on different groups will only be possible from `group_logs` tables (W&B forces using the name of the run for drawing charts and prefixing everything looked confusing).

tracking/translations_parser/cli/experiments.py
```diff
@@ -71,16 +75,73 @@ def parse_experiment(
 
 
 def publish_group_logs(
```
eu9ene, line 77: I don't want to block the merging since it adds functionality we need but next we touch this function it should be quickly refactored by extracting multiple functions.

eu9ene (approved):

PR #377, opened 2024-01-19
Add the ability to run starting from a specific task (fixes #227)

### LINKED ISSUE #227: Add an ability to run starting from a specific task
We already have the `target-stage` setting which is convenient to specify where to finish. We should also add an option to specify where to start. It would speed up debugging on Taskculster a lot. Even if the caches work properly we might want to debug the later stages of the pipeline without rerunning the earlier ones, even if the upstream scripts have changed.

bhearsum:
A couple of example runs with this:
* https://firefox-ci-tc.services.mozilla.com/tasks/groups/YHAr0HzwSSe4pe5Yh9dIlg uses https://firefox-ci-tc.services.mozilla.com/tasks/groups/JjNp3KcyTUObUtOA9BgK5g as its `previous-group-id` with `start-stage: train-backwards` and `target-stage: train-teacher` - and ends up running `train-backwards, `translate-mono-trg`, `collect-mono-trg`, and `train-teacher`.
* https://firefox-ci-tc.services.mozilla.com/tasks/groups/Sm0YV_8LQP-EOE8Nz6G5Lw uses the above group as its `previous-group-id` with `start-stage: train-teacher` and `target-stage: all`. Note that it ended up depending on tasks from both the above group and the one that it was based on, and ended up scheduling `train-teacher` and everything after it (I didn't bother letting them all run - I think the scheduling is enough to verify this).

    
Big thanks to @gabrielBusta for suggesting this implementation!


docs/task-cluster.md
```diff
@@ -97,6 +97,20 @@ tasks:
...
+```
+start-stage: train-student
+target-stage: all
+previous_group_ids: ["SsGpi3TGShaDT-h93fHL-g"]
+```
+
+...will run `train-student` and all tasks _after_ it. All tasks upstream of `train-student` will be replaced with the tasks of the same name from the `SsGpi3TGShaDT-h93fHL-g` task group. (If there are required tasks that do not exist in any of the provided groups, they will be created - so you must ensure that any tasks you want to skip exist in one of the previous groups.)
+
+Note: This feature should _never_ be used for production training, as it completely bypasses all caching mechanisms, and you will most likely end up with invalid or useless models.
+
```
hneiva, line 113: great doc! 👍

hneiva (approved):

bhearsum:
@eu9ene - Can you let me know if this sounds like it addresses #227?

docs/task-cluster.md
```diff
@@ -97,6 +97,20 @@ tasks:
...
+When hacking on later parts of the pipeline it can often be useful to re-use earlier runs of the pipeline, even if those runs were done with different training parameters. To do this, we must bypass the usual caching mechanisms of Taskgraph, and force it to replace earlier tasks with ones we provide. To do this, you can run a training action as usual, but also provide `start-stage` and `previous_group_ids` parameters. For example:
+
+```
+start-stage: train-student
+target-stage: all
+previous_group_ids: ["SsGpi3TGShaDT-h93fHL-g"]
+```
+
+...will run `train-student` and all tasks _after_ it. All tasks upstream of `train-student` will be replaced with the tasks of the same name from the `SsGpi3TGShaDT-h93fHL-g` task group. (If there are required tasks that do not exist in any of the provided groups, they will be created - so you must ensure that any tasks you want to skip exist in one of the previous groups.)
```
eu9ene, line 110: > so you must ensure that any tasks you want to skip exist in one of the previous groups

I think this might become problematic because sometimes we restart an experiment many times and some cached tasks might be from older experiments. I think the code in the train action should just traverse recursively all the cached tasks and add them too. Then we can specify just one group id. Also specifying multiple group IDs will make it ambiguous if a task is duplicated in both groups.

bhearsum: That's fair - I don't see any reason that we can't do this.

gabrielBusta: So we'd specify some tasks that are existing, and then fallback to the cache?

bhearsum: I think how this would work is we would take any tasks from the group id specified that we need, and are successful, and then walk through their dependencies for any other tasks that we need. (Essentially, we're going to do some extra work in the action task to discover additional existing tasks rather than require them to be specified.)

bhearsum: I just pushed an update here that should address this. It requires that a `previous_group_id` that contains the task or tasks from the `start-stage` to exist in it - but after that it will walk back up the dependency chain to find the rest.

This prevents one from starting, eg: `train-teacher` off of a group that only ran up to `train-backwards`, but based on the description of the use case I think this is generally going to be used when you've already run to task X and then want to run it again.

@eu9ene - does this address your concerns?

(It might be possible to address this limitation in the future, but due to the limited data that action tasks have in them it's not trivial...)

This new version also depends on https://github.com/taskcluster/taskgraph/pull/418 to be merged and released.

bhearsum: Oh, and you can give it a try right now by running training actions off of https://firefox-ci-tc.services.mozilla.com/tasks/groups/Ri8ENf-AReefjxNRxBESEQ.

eu9ene: Great, this should work for us. Let's merge and I'll test it later.

eu9ene (approved):

bhearsum:
@gabrielBusta - I know Heitor reviewed this but could you take a look as well, since you've got more perspective on the training action and config changes?

taskcluster/translations_taskgraph/actions/train.py
```diff
@@ -54,49 +56,34 @@ def validate_pretrained_models(params):
...
+task is upstream of `start-stage`. (That is to say: even if a task from one of these groups
+has a cache digest that doesn't match what the downstream task wants, it will still be used. This
+can be used for quick iteration of functionality where the quality of the outputs is not important.)""",
+                "items": {
+                    "type": "string",
+                },
+            },
+            "start-stage": {
+                "type": "string",
+                "description": """The stage of the pipeline to begin at, provided replacements
```
gabrielBusta, line 73: This is also "Optional"

gabrielBusta, line 73: `start-stage` is also "Optional."

taskcluster/translations_taskgraph/actions/train.py
```diff
@@ -343,6 +330,32 @@ def train_action(parameters, graph_config, input, task_group_id, task_id):
 
     parameters = dict(parameters)
 
+    start_stage = input.pop("start-stage", None)
```
gabrielBusta, line 333: Why `pop`? Why not `get` and default to `None` like `previous_group_ids`?

taskcluster/translations_taskgraph/actions/train.py
```diff
@@ -343,6 +330,32 @@ def train_action(parameters, graph_config, input, task_group_id, task_id):
...
+    if start_stage:
+        previous_group_ids = input.get("previous_group_ids")
+
```
gabrielBusta, line 336: What happens if `previous_group_ids` is `None`? Should we handle that case and throw a `ValueError` (or let it throw a `KeyError`)?

taskcluster/translations_taskgraph/actions/train.py
```diff
@@ -343,6 +330,32 @@ def train_action(parameters, graph_config, input, task_group_id, task_id):
...
+            full_task_graph = get_artifact(graph_id, "public/full-task-graph.json")
+            combined_full_task_graph.update(full_task_graph)
+        _, combined_full_task_graph = TaskGraph.from_json(combined_full_task_graph)
+
+        # Next, we find the task id(s) corresponding of the tasks that match the stage
+        # we want to start at.
+        start_task_ids = []
+        for label, task in combined_full_task_graph.tasks.items():
+            if task.attributes.get("stage") == start_stage:
+                start_task_ids.append(label_to_task_id[label])
```
gabrielBusta, line 351: `label_to_task_id` will be the last thing `get_artifact` returns at the end of `for graph_id in previous_group_ids`. Is that intended? Would it be better to look up `combined_full_task_graph[label]`? We might throw a `KeyError`.

taskcluster/translations_taskgraph/actions/train.py
bhearsum: Good callout. The `get` for `previous_group_ids` is actually a copy/paste error from https://github.com/mozilla-releng/mozilla-taskgraph/blob/cb7a5d6440acc1cfad8f89ec226cc53d6c61f7ba/src/mozilla_taskgraph/actions/release_promotion.py now that I look at it. (Over in that code we don't put `input` into parameters, so it doesn't matter...but over here we do, so we ought to remove it.)

taskcluster/translations_taskgraph/actions/train.py
bhearsum: +1. I'll throw an error for this case. _Technically_ we could set it to an empty list and it would work...but it wouldn't do what we expected.

taskcluster/translations_taskgraph/actions/train.py
bhearsum: Ah, good catch! Yes, this is a subtle bug waiting to happen. I'll fix it.

gabrielBusta (approved):

PR #379, opened 2024-01-22
Add some docs to what dedupe is

gregtatum:
I was confused as to what this was, so I added some docs.


eu9ene (approved):
Thanks for clarifying this! Another place where we might want to add clarification is the training guide [Find datasets section](https://firefox-ci-tc.services.mozilla.com/tasks/KRXRJ_lTSWWrv0F20lORCQ).

PR #386, opened 2024-01-23
Publish experiment config

### LINKED ISSUE #332: Publish experiment config to W&B
We should parse and publish the full YAML experiment config ([example1](https://github.com/mozilla/firefox-translations-training/blob/main/configs/tc.prod.yml), [example2](https://github.com/mozilla/firefox-translations-training/blob/main/taskcluster/translations_taskgraph/parameters.py)) to the W&B experiment (group). I assume the only place we can use now is the fake `group_logs` step which is fine. It will allow us comparing those configs between experiments

vrigal:
Based on #374

Closes #332


eu9ene (approved):

eu9ene:
Looks good! Let's reupload all the GCP experiments then

PR #396, opened 2024-01-25
Update import sorting to fix local lint checks

gregtatum:
...I have no idea why this is different for me locally...

However, these files are failing for me locally but not in CI. We should figure out why, but in the short term (assuming CI passes) I would at least like to do a quick fix on these. I wonder if it's black and ruff fighting each other. It might be worth trying to only use the ruff formatter.


gregtatum:
Great, so local and CI disagree. I'm leaving this as draft to investigate.

gregtatum:
I'm closing for now as I don't want to spend more time looking into this right now.

Issue #143, opened 2023-07-20
Add support for public custom corpus' in Taskcluster


bhearsum:
The Snakemake pipeline has support for using "custom" corpus' - that is, datasets that are not available through an URL. As I understand it, these work by someone creating the dataset with a specific name somewhere in the Berlin cluster, which is then referenced in the pipeline config.

I think our best bet for doing this would be to set-up a GCP bucket to hold the custom corpus' and add a new importer that knows how to fetch from it. The new importer would receive the bucket name (or possibly URL) in addition to the usual src/trg/output_prefix/dataset. We'd specify these datasets with strings like "custom-gcp_dataset-name". The bucket would be configured in `ci/config.yml`.

This would also allow us to train against private datasets (eg: things that we purchase that cannot be redistributed publicly).

gabrielBusta:
Seems related to https://github.com/mozilla/firefox-translations-training/issues/191

I get the feeling we should get a bucket for translations?

eu9ene:
Yes, we should create a public bucket for translations where we can store old models for fine-tuning and public custom datasets (that are not available on OPUS or mtdata). I don't know about the archive of the jobs, the requirements for visibility of this one and the retention policy might be different, so I would put it in a separate bucket.

We will need the customer importer to work soon to train English to Catalan cc @gregtatum

PR #397, opened 2024-01-25
Add huggingface to the find_corpus

gregtatum:
I added this while looking into #143. I did not add a test since the rest of the file is tested, and it would be too onerous to mock out the huggingface API, and the risk of breakage is low, and the impact of breakage would be low.


utils/find_corpus.py
```diff
@@ -128,6 +128,186 @@ def get_sacrebleu(source: str, target: str):
...
+            ],
+        ]
+    )
+
+
+def is_useful_dataset(dataset: any) -> bool:
+    """Determines if a dataset is useful or not."""
+    return (
+        "task_categories:automatic-speech-recognition" not in dataset.tags
+        and dataset.downloads > 5
```
eu9ene, line 265: this one is a bit subjective. Can we just add the number of downloads to the table and let the user decide on this? I can imagine someone uploaded a fresh dataset for low resource language that has < 5 downloads

eu9ene (approved):
nice, it's definitely useful to see what's on HF! We should add a corresponding importer though. I don't think there's much risk here since it's just data.

utils/find_corpus.py
gregtatum: Sure, most of them were junk, but I can leave them unfiltered.

PR #402, opened 2024-01-26
fix(snakefile): correctly locate files to translate

AmitMY:
This regards #368 

Seems like the new splitting file saves the files in an unexpected format for the translate function

I don't know how to give it a generic `file.{{part}}.{{ext}}` to allow both gzip and other supported compressions

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -453,7 +453,7 @@ rule translate_mono_trg:
     threads: gpus_num * 2
     resources: gpu=gpus_num
     input:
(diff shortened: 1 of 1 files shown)


gregtatum (approved):

AmitMY:
Thanks @gregtatum - you sure you don't want to make it work on generic compressed extensions?

gregtatum:
@AmitMY We're not actively maintaining the Snakefile right now as we're working on the Taskcluster pipeline, so I'm happy to merge in fixes, even if they are hotfixes for the moment. I'd be happy to merge in other work to make it more correct and generic.

PR #406, opened 2024-01-29
Taskcluster task group publication

### LINKED ISSUE #312: Publish past experiment directly from Taskcluster
Depends #309 

We need a new publication mode that relies only on a Taskcluster task group ID to find all relevant tasks and their artifacts (configurations, logs and metrics) and publish them on Weight & Biases.

This will not run inside Taskcluster, but help to publish existing experiments into Weight & Biases without relying on model/experiment/logs directory.

Past experiments to publish:
- [en-hu](https://firefox-ci-tc.services.mozilla.com/tasks/groups/aY25-4fXTcuJNuMcWXUYtQ)
- [lt-en 1](https://firefox-ci-tc.services.mozilla.com/tasks/groups/EW7qV3U5SBSjegnTxGZHkw)
- [lt-en 2](https://firefox-ci-tc.services.mozilla.com/tasks/groups/V-OmRM1yS_GwESTVDPl0VQ)
- [lt-en 3](https://firefox-ci-tc.services.mozilla.com/tasks/groups/K1iHndFUSxSEDRLg_H9l1A)
- [lt-en 4](https://firefox-ci-tc.services.mozilla.com/tasks/groups/DXbS0zreSGSVYloAF8gwJg)

vrigal:
Ref. #312

Results are visible on https://wandb.ai/teklia/lt-en.
The publication is almost exactly the same as for experiments.

I wasn't able to add a test. Mocking Taskcluster queue was complex and I prefer work on other important topics.


eu9ene (commented):
Nice work! Before we merge it let's double check that it's consistent with the uploaded GCP experiments. Some things I noticed:
- Naming of the evaluation datasets (ideally they should be named exactly as specified in the experiment config). We already parse the dataset parts, so it should be easy to change the formatting.
- Marian config is missing for the training tasks (I do see [experiment config](https://wandb.ai/teklia/lt-en/runs/rwfvgbec/overview) instead which is strange)
-  No experiment config and no `group_logs`. Since we already have the code to parse and upload the configs, can we just sync all this in this PR?

I prefer fixing those before we start uploading the new experiments.

vrigal:
@eu9ene I made the `publish_group_logs` function generic, so the behavior is the same than for experiments.
This is not ideal for the Taskcluster approach, as it requires to build a structure with logs of the evaluation tasks (we could direcly read `.metrics` files here).
It is functional like this, we can open an issue once we merge this IMO to improve this specific part of the code.

> Naming of the evaluation datasets

I don't understand how I can get the information from the experiment config. Actually I use the label from taskcluster, which seems to be inconsistent sometimes (thus displaying data in different graph):
* `evaluate-student-flores-aug-typos_devtest-lt-en` becomes "Flores-aug-typos_devtest-lt-en"
* `evaluate-teacher-ensemble-flores-flores_aug-typos_devtest-lt-en` becomes "Flores-flores_aug-typos_devtest-lt-en"

I'll work on the other points this morning (run config + group logs data & config), but the above points will certainly be followups so someone else can also work on this.

vrigal:
This work should be fine now. I published data from TC group [K1iHndFUSxSEDRLg_H9l1A](https://firefox-ci-tc.services.mozilla.com/tasks/groups/K1iHndFUSxSEDRLg_H9l1A), results are visible on https://wandb.ai/teklia/lt-en-PR-406.
Two follow-ups:
* Inconsistent dataset naming (point mentioned [above](https://github.com/mozilla/firefox-translations-training/pull/406#issuecomment-1945604474))
  * Could either be other source, or automatically remove duplicate prefix
* Traversal tasks (Eva will work on this)

eu9ene:
> I don't understand how I can get the information from the experiment config. Actually I use the label from taskcluster, which seems to be inconsistent sometimes (thus displaying data in different graph):

    evaluate-student-flores-aug-typos_devtest-lt-en becomes "Flores-aug-typos_devtest-lt-en"
    evaluate-teacher-ensemble-flores-flores_aug-typos_devtest-lt-en becomes "Flores-flores_aug-typos_devtest-lt-en"


Unfortunately, Taskcluster labels are inconsistent that's why I wrote and recommended using this [regex](https://github.com/mozilla/firefox-translations-training/blob/190358a9231c5b4e057bf465fcbcef783ddbcc34/utils/taskcluster_downloader.py#L23) (which might be not compete too

eu9ene:
I also noticed `teacher-ensemble` step is missing.

eu9ene (approved):
Ok, it looks better now with the configs in place. Let's address the rest of the issue as a follow-up with priority.

- recursive publishing
- dataset naming
- teacher ensemble step is missing

eu9ene:
@vrigal please relock the poetry file and I'll merge

PR #416, opened 2024-02-01
Download taskcluster live logs for training

eu9ene:
While we're waiting for W&B integration, it's convenient to download live logs periodically to display them on tensorboard.
This works also for completed tasks.

For example:

```
LOGS_TASK_GROUP=FgNRsJFdQ1CkjT8I20utGg make download-logs
make tensorboard
```

[skip ci]

### DIFF
--- a/utils/taskcluster_downloader.py
+++ b/utils/taskcluster_downloader.py
@@ -50,7 +50,7 @@ def donwload_logs(group_id, output):
     group = queue.listTaskGroup(group_id)
 
     for task in group["tasks"]:
-        if task["status"]["state"] != "completed" and task["status"]["state"] != "running":
+        if task["status"]["state"] not in ("completed", "running"):
             continue
 
         label = task["task"]["tags"]["kind"]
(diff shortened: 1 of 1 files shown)


utils/taskcluster_downloader.py
```diff
@@ -48,7 +50,7 @@ def donwload_logs(group_id, output):
     group = queue.listTaskGroup(group_id)
 
     for task in group["tasks"]:
-        if task["status"]["state"] != "completed":
+        if task["status"]["state"] != "completed" and task["status"]["state"] != "running":
```
bhearsum, line 53: nit (non-blocking): `if task["status"]["state"] not in ("completed", "running"):` would be easier to read.

utils/taskcluster_downloader.py
```diff
@@ -64,14 +66,29 @@ def donwload_logs(group_id, output):
...
-        print(f"Downloading {output_path}")
-        with open(output_path, "wb") as f:
-            downloadArtifactToFile(
-                f,
-                taskId=task["status"]["taskId"],
-                name="public/build/train.log",
-                queueService=queue,
-            )
+        url = queue.buildUrl("getLatestArtifact", task_id, "public/logs/live.log")
+        resp = requests.get(url, stream=True, timeout=5)
```
bhearsum, line 70: suggestion (non-blocking): Consider moving this and the other parts of log processing out to an asynch function.

Now that this loop is handling running tasks as well, downloading logs synchronously may be less than ideal. For example, if there's two tasks that are both running you will end up sitting in the body of the loop for hours or even days depending on which task is looked at first. Downloading async will ensure you fetch all live logs as the tasks run.

utils/taskcluster_downloader.py
```diff
@@ -64,14 +66,29 @@ def donwload_logs(group_id, output):
...
+        print(f"Downloading {url}")
+        lines = []
```
bhearsum, line 72: suggestion (non-blocking): consider not saving all of the lines in memory. AFAICT neither `lines` nor `log_lines` are necessary - I believe you could open the output file immediately, and put the `start_writing` logic in that loop instead.

bhearsum (approved):

utils/taskcluster_downloader.py
eu9ene: I use streaming and kill the connection on timeout, so it takes 5 seconds to download any live log. I would like not to spend time on this script because it's a very temporary workaround until we download W&B which I hope will happen soon.

Issue #395, opened 2024-01-24
When using a pre-trained model there shouldn't be dependencies on dataset steps


eu9ene:
For example I want to experiment with teacher training and use a pre-trained backward model (very common use case). 

If I change the training data or cleaning procedure for the training corpus Taskcluster will rerun decoding with the pre-trained backward model because `merge-corpus`, `merge-devset` and `train-vocab` are upstream dependencies. This slows down experimentation a lot and wastes resources.

We should remove those dependencies from the graph if a pre-trained model is being used.

@gabrielBusta

bhearsum:
This could also happen when re-using other pretrained teachers. The general problem seems to be that we still have some dependencies on upstream tasks that we don't use artifacts from. (Perhaps we can do some sanity checking here that we don't have dependencies on something we don't have a `fetch` for?)

bhearsum:
@eu9ene - do you have an example of this handy? (I want to make sure I can repro locally to verify a fix...)

bhearsum:
And could you also clarify exactly which tasks run that shouldn't? I think you're talking about `collect-mono-trg`, `translate-mono-trg`, and `split-mono-trg` when you say "rerun decoding" but I want to be certain...

eu9ene:
I'm running this experiment with a pre-trained backward model now https://firefox-ci-tc.services.mozilla.com/tasks/Wz8KJcqPTAOFhV-YaRGJ0g. You could just add a pre-trained model to the CI config and check that it doesn't rerun translate steps by the backward model if you change something in the cleaning scripts.

I'm mostly talking about tranlsate-mono-trg because it's quite heavy and waste of resources. Currently f we modify anything upstream from merge-corpus it will rerun the fake train-backward and translate-trg and related steps.

bhearsum:
I think I understand what's going on here. The unnecessary dependencies when we use the `use` training continuation mode is one part of it (it causes us to pull in `merge-corpus` and `merge-devset` for `train-backwards`, for example) but the bigger problem is that because we still run a `train-backwards` step, we need to satisfy all dependencies downstream of it. Most notably in this, `translate-mono-trg` and `collect-mono-trg` exist between `train-backwards` and `train-teacher`. Unless we find these in the cache, they will run. And if _their_upstream change (such as cleaning tasks), they will be forced to run again.

[...]

Issue #417, opened 2024-02-02
Easier graph modifications


eu9ene:
We have a bunch of use cases when the DAG should be modified and we will have more in the future:
1. The number of chunks for translations
2. Skipping bilcleaner for a dataset (see #415)
3. Skipping bicleaner entirely
4. Downloading and using a pre-trained backward model (changes upstream dependencies and required fetches, see #395)
5. Downloading and using the pre-trained teacher models (skip the whole data augmentation part of the graph, backward model is still needed for cross entropy filtering step)
6. Skipping data augmentation when we want to check how much boost it brings or when we don't have monolingual data (similar to previous one, but the `merge-corpus` step looses its dependencies on translated mono corpus)
7. Adding extra pre-processing steps when using pre-trained models of different architectures (see [opusmt pr](https://github.com/mozilla/firefox-translations-training/pull/192/files#diff-47959dfd378b3cd1d39b5515418ee8e4444ab7a6036d5197a5bea82814f928a3R665))

The conclusion is that the graph is not static. It should be adjusted based on the experiment config. Currently the only way to do this is using transforms that we already do to find upstream tasks related to the dataset cleaning. Each transform corresponds to only one kind. This approach makes it hard to say how exactly the graphs will look like in the end and it's hard to understand and maintain the code that changes the graph for each task independently. It also limits us, for example we can't easily introduce a new kind for pre-trained model downloading and have to modify the `train-` kinds for this which leads to issues with redundant dependencies and incorrect caching.

The ideal solution would be similar to Snakemake graph definition: one python file where we can fully form and connect the graph based on conditions. Instead of removing dependencies for some steps in the transforms we could skip their definition in `kind.yaml` and add the dependencies that are needed in this new "connect graph" script. We can think of it as graph-level transform.

If we define all the possible tasks/kinds in yaml and need only to connect the graph, the ideal place for this operation would be somewhere before [full_task_graph step](https://taskcluster-taskgraph.readthedocs.io/en/latest/concepts/task-graphs.html) in graph generation. It would be somewhere [here](https://github.com/taskcluster/taskgraph/blob/85c360e3e2adf94d64a14988f560f6d591509483/src/taskgraph/generator.py#L294) in the TaskGraphGenerator code, before we start edding edges to the graph based on kind dependencies. If the task is disconnected from the graph, we should just skip its analysis entirely. 

[...]

PR #418, opened 2024-02-02
fix: adjust dependencies, fetches, and cache digests when using the `use` training contiunation mode

bhearsum:
Dependencies and fetches are unused, so they should be removed.

Cache digests should _only_ be influenced by the pretrained teacher parameters (file resources and other parameters are unused - so they do not influence the outcome of the task).

This largely addresses #395 - I've done a number of tests:
- Create new training that re-uses a backwards model from elsewhere; run it up to train-teacher - https://firefox-ci-tc.services.mozilla.com/tasks/groups/H6Z4Icc3Sz2kZLJZBx1qOw
- Modify something that affects mono and corpus cleaning & run training again up to train-teacher - https://firefox-ci-tc.services.mozilla.com/tasks/groups/QEvzJrasTUuj9_WK4dycpA
  - This reruns ~everything _except_ train-backwards
- Modify something that along affects corpus cleaning & un training again up to train-teacher - https://firefox-ci-tc.services.mozilla.com/tasks/groups/Ddfe_YYaTHqRnyjaHVfRFA
  - This reruns corpus chain + train-teacher, but _not_ the mono chain nor train-backwards
- Modify something that only affects mono & run training again up to train-teacher - https://firefox-ci-tc.services.mozilla.com/tasks/groups/ED3iVqD3S_apaesgpiNVQA
  - This reruns the mono chain + train-teacher but still not train-backwards

I've also started testing happens if certain parameters change. For example [this group](https://firefox-ci-tc.services.mozilla.com/tasks/groups/B4FAJPEeQEuqYHDtU5n-Qw) runs with the same parameters, except the `valid-freq` for both `train-backwards` and `train-teacher` have been tweaked to verify that parameter changes don't cause `train-backwards`. Unfortunately, `train-backwards` still ran...so there's still a theoretical issue somewhere.

In any case - I think this patch covers all of the cases we will realistically hit. I plan to try to get to the bottom of the other case in a follow-up.


bhearsum:
> I've also started testing happens if certain parameters change. For example [this group](https://firefox-ci-tc.services.mozilla.com/tasks/groups/B4FAJPEeQEuqYHDtU5n-Qw) runs with the same parameters, except the valid-freq for both train-backwards and train-teacher have been tweaked to verify that parameter changes don't cause train-backwards. Unfortunately, train-backwards still ran...so there's still a theoretical issue somewhere.

I got the bottom of this. It happens because [the command ends up as part of the digest data](https://github.com/mozilla/firefox-translations-training/blob/d14881ea710a114cbbd85c98db08a12102e328e6/taskcluster/translations_taskgraph/transforms/cached_tasks.py#L75), and my test was changing one of the parameters that influences the command. This wasn't the most realistic test anyways, so I think this actually ignorable...

I ended up doing a different test to verify that the cache digest is unchanged. [This task group](https://firefox-ci-tc.services.mozilla.com/tasks/groups/MQy9GOCeR1eKGJSVPGP_kA) was created with exactly the same parameters as the first test from my earlier comment, but I modified `train.sh` to bust the cache for anything that depends on it. As expected, `train-backwards` did not have to rerun - it only ran `train-teacher`.

eu9ene (approved):
This should work but it adds complexity to graph definition. An alternative would be to introduce a new kind to download models and adjusting dependencies in downstream tasks. Then we would have the cleaner kinds that almost don't change, only the dependencies would be adjusted. The final graph would show a different task. The downside is that the backward model is used in multiple places so we would need to modify multiple other kinds + the ones downstream from teachers which is maybe even more complex in the end.

eu9ene:
If we come up with an elegant solution for https://github.com/mozilla/firefox-translations-training/issues/417 it should help addressing the complexity.

eu9ene:
@bhearsum let's merge this. I need this fix to iterate faster. If something comes up let's iterate in a follow-up PR

PR #431, opened 2024-02-08
Move snakemake to a separate folder

### LINKED ISSUE #266: Move Snakemake code to a separate directory
It would be great to have all the code that's needed for a specific workflow manager in a separate directory, like we now have for Taskcluster. We might want to defer this until we merge #117, otherwise it will be hard to see the diff since a lot of the changes are Snakemake related.

eu9ene:
- move all snakemake related code to a separate folder, similar to taskcluster
- clarify notice in the docs about the unmaintained state
- move taskcluster configs to `taskcluster` folder
- split makefile

fixes #266 

[skip ci]


eu9ene:
I wonder why the linter works differently locally:

```
[task 2024-02-08T23:06:05.498Z] tracking/translations_parser/cli/experiments.py:9:1: I001 [*] Import block is un-sorted or un-formatted
[task 2024-02-08T23:06:05.498Z] tracking/translations_parser/publishers.py:1:1: I001 [*] Import block is un-sorted or un-formatted
```

@gregtatum any ideas?

eu9ene:
> I wonder why the linter works differently locally:
> 
> ```
> [task 2024-02-08T23:06:05.498Z] tracking/translations_parser/cli/experiments.py:9:1: I001 [*] Import block is un-sorted or un-formatted
> [task 2024-02-08T23:06:05.498Z] tracking/translations_parser/publishers.py:1:1: I001 [*] Import block is un-sorted or un-formatted
> ```
> 
> @gregtatum any ideas?

It seems it doesn't understand that `translations_parser` is an internal module locally. It expects full import path starting from `tracking.translations_parser`
```
import wandb

from translations_parser.data import Metric
```

gregtatum:
Ah the sorting based on local makes sense there. I couldn't figure it out in the time I took with #396. It might be worth seeing if we can get it to understand the difference, or maybe just disable the sorting rule. This has been really annoying me.

gregtatum (approved):
Thanks for the move! This will simplify and clarify things a bit for us.

eu9ene:
> Ah the sorting based on local makes sense there. I couldn't figure it out in the time I took with #396. It might be worth seeing if we can get it to understand the difference, or maybe just disable the sorting rule. This has been really annoying me.

I think the issue is in how the tracking package is organized. We should either add an extra PYTHONPATH everywhere or use absolute imports from the root of the training repo. cc @vrigal

PR #435, opened 2024-02-09
Work aroud fast text model downloading failures

### LINKED ISSUE #432: Issue with opuscleaner fasttext filter
https://firefox-ci-tc.services.mozilla.com/tasks/GLe3HjcMSNOQrjDz8p2U6w/runs/0/logs/public/logs/live.log

```
[task 2024-02-08T23:39:43.538Z] [1/10:fasttext_filter] Traceback (most recent call last):
[task 2024-02-08T23:39:43.538Z] [1/10:fasttext_filter]   File "/builds/worker/.local/lib/python3.10/site-packages/opuscleaner/filters/./fasttext_filter.py", line 75, in <module>
[task 2024-02-08T23:39:43.538Z] [1/10:fasttext_filter]     main()
[task 2024-02-08T23:39:43.538Z] [1/10:fasttext_filter]   File "/builds/worker/.local/lib/python3.10/site-packages/opuscleaner/filters/./fasttext_filter.py", line 62, in main
[task 2024-02-08T23:39:43.538Z] [1/10:fasttext_filter]     model = fasttext.load_model(f"{args.model_type}.bin")
[task 2024-02-08T23:39:43.538Z] [1/10:fasttext_filter]   File "/usr/lib/python3/dist-packages/fasttext/FastText.py", line 441, in load_model
[task 2024-02-08T23:39:43.539Z] [1/10:fasttext_filter]     return _FastText(model_path=path)
[task 2024-02-08T23:39:43.539Z] [1/10:fasttext_filter]   File "/usr/lib/python3/dist-packages/fasttext/FastText.py", line 98, in __init__
[task 2024-02-08T23:39:43.539Z] [1/10:fasttext_filter]     self.f.loadModel(model_path)
[task 2024-02-08T23:39:43.539Z] [1/10:fasttext_filter] ValueError: Invalid model file.
[task 2024-02-08T23:39:43.539Z] [1/10:fasttext_filter] Please download the updated model from www.fasttext.cc.
[task 2024-02-08T23:39:43.539Z] [1/10:fasttext_filter] See issue #332 on Github for more information.
[task 2024-02-08T23:39:43.539Z] [1/10:fasttext_filter] 
```

eu9ene:
It's hard to test but I think it should help. Let's see if there will be more issues in the future.

fixes #432 

[skip ci]


taskcluster/kinds/clean-corpus/kind.yml
```diff
@@ -68,7 +68,11 @@ tasks:
                 USE_OPUSCLEANER: "{use_opuscleaner}"
                 COMPRESSION_CMD: zstdmt
                 ARTIFACT_EXT: zst
-    
+
+# TODO: doesn't work. How to enable retries when the task fails?
+#        # there might be intermittent issues with OpusCleaner, specifically FastText model downloading can fail
+#        retries: 5
```
eu9ene, line 74: @bhearsum this didn't work for me. I see in the docs that task retries work only when infrastructure fails. How to add retries when the task fails?

eu9ene:
We should be able to support retries. Alternatively we can download fast text model through toolchain, but there are many tools in opus cleaner that might need to download other models so I'd rather encapsulate this logic inside OpusCleaner and implement a generic solution for intermittent downloading issues. @bhearsum I need your advice here.

taskcluster/kinds/clean-corpus/kind.yml
marco-c: I think you can use `retry-exit-status` for this: https://github.com/taskcluster/taskgraph/blob/a7345114e4af294dc31c22c438a5ddb3467676da/src/taskgraph/transforms/task.py#L349.

Another option would be to retry within the command executed by the task. It might be better: the retry would work independently whether the command is run on Taskcluster or locally, and it would be possible to have a smarter retry algorithm (e.g. exponential backoff).

marco-c: Also, by retrying the download within the command, you don't "waste" the time that has elapsed so far (e.g. installing packages, etc.).

eu9ene: There are multiple ways to add retries but it should be supported by the workflow manager. It is a standard functionality. See [snakemake](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#defining-retries-for-fallible-rules) for example. Rather than adding code in each script that has external dependencies or other intermittent issues I'd set one line in Taskcluster definitions. And we have external dependencies in many tasks (datasets, cleaning, bicleaner, pip install, github etc.).

eu9ene: If we want to be economical about spinning up the machines, we should move to docker images where everything is already installed and cached. Some installations can take quite a while.

marco-c: It is supported and you can use it. I was suggesting not to use it because I think it might better to do it explicitly within the task when it makes sense than at the task level.
Note even if you use Docker where everything is installed and cached, you're still going to pay the price of spinning up a machine and downloading the Docker image from a registry.

That said, it is just an opinion about a small thing, and it's not a strong opinion, so feel free to do what you want ;)

eu9ene: I see, I thought you're saying it's not supported. I'd rather reduce complexity of the code by handling it on the task level. This should not happen often. If it does we'll want to address this particular problem differently.

bhearsum:
> We should be able to support retries. Alternatively we can download fast text model through toolchain, but there are many tools in opus cleaner that might need to download other models so I'd rather encapsulate this logic inside OpusCleaner and implement a generic solution for intermittent downloading issues. @bhearsum I need your advice here.

Let's talk about this synchronously at our meeting today.

taskcluster/kinds/clean-corpus/kind.yml
```diff
@@ -68,7 +68,10 @@ tasks:
                 USE_OPUSCLEANER: "{use_opuscleaner}"
                 COMPRESSION_CMD: zstdmt
                 ARTIFACT_EXT: zst
-    
+            # there might be intermittent issues with OpusCleaner, specifically FastText model downloading can fail
+            retry-exit-status:
+                - 1
```
bhearsum, line 73: Are there reasons other than model downloading that this may exit with 1? Preferably, model downloading should fail with a distinct code to avoid unnecessary retries.

eu9ene: I see that OpusCleaner exits with 1 on all exceptions. I lost the failed tasks, it's hard to find them now, but I think it was 1 for fasttext

bhearsum: I'm not going to say it blocks landing this, but a nice follow-up would be to catch this specific failure (eg: by analyzing output) and returning a distinct code for it.

bhearsum (approved):

taskcluster/kinds/clean-corpus/kind.yml
eu9ene: Agreed it would be nice to reproduce. However, I still think we should handle code 1 because it's how OpusCleaner exists on error and we generally want to retry on those errors. It includes many rules and some of them have external dependencies.

PR #437, opened 2024-02-12
Add a structured logging script

gregtatum:
This is a breakout from other PRs.


gregtatum:
Here is an example run:

https://github.com/mozilla/firefox-translations-training/pull/437/checks?check_run_id=21488258146

eu9ene (approved):
Please double check the format and that it doesn't duplicate the timestamp that we already have under Taskcluster.

bhearsum (approved):

PR #438, opened 2024-02-12
Add kind that demonstrates how to modify the upstream graph in a transform

bhearsum:
This is a _very_ rough prototype for what's been discussed in https://github.com/mozilla/firefox-translations-training/issues/417 and https://github.com/taskcluster/taskgraph/issues/424. In this example, the decision about which tasks `merge-corpus` should depend on are being deferred and handled in the new `modify_graph` transform, which runs as part of the `modify-graph` task that happens right before `all`.

Because we're handling the `dependencies` later, we must also handle the `fetches` later - which means that knowledge about those artifact names moves _out_ of the `merge-corpus`  kind and into this transform.

I haven't tested this besides some `taskgraph full -Y` inspection, but it looks plausible - and the code certainly runs at the appropriate time, and has access to the necessary data to make the changes.

Any tasks that you need to modify or access _must_ be in a kind listed in the `modify-graph` kind's `kind-dependencies`, otherwise they will not show up in `config.kind_dependencies.tasks`.

And at the risk of being annoying and repeating myself, I will offer a couple of cautions:
- This sort of thing goes against the design of taskgraph. It could break without warning, and may not fit in well with future taskgraph features
- We used to manage Firefox build & release tasks/graphs with [similar imperative code](https://github.com/mozilla-releng/build-buildbotcustom/blob/master/process/factory.py#L396). Taskgraph was, in part, a reaction to how unmaintainable and spaghetti-like that became.

### DIFF
--- a/taskcluster/kinds/all/kind.yml
+++ b/taskcluster/kinds/all/kind.yml
@@ -16,6 +16,7 @@ kind-dependencies:
     - evaluate
     - evaluate-quantized
     - evaluate-teacher-ensemble
+    - modify-graph
 
 tasks:
     all:
--- a/taskcluster/kinds/merge-corpus/kind.yml
+++ b/taskcluster/kinds/merge-corpus/kind.yml
@@ -6,14 +6,12 @@
 loader: taskgraph.loader.transform:loader
 
 transforms:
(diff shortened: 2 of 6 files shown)


bhearsum:
This will never be merged; it was just an example.

PR #441, opened 2024-02-13
Fix linting for tracking

eu9ene:
Make sure `wandb` directory is not written from tests so that Ruff doesn't think that it's a module locally. This led to constant issues with imports being formatted differently locally. On CI linting is isolated from the tests.

+ relock the poetry file which is not related to this fix but won't hurt

[skip ci]


eu9ene:
@vrigal FYI I found the reason why linting didn't work locally. You should delete `wandb` directory in the root and merge this PR.

gregtatum (approved):
Praise: Thanks for finding this! This has been a real pain for my pre-push hooks.

PR #445, opened 2024-02-14
Fix random seed

### LINKED ISSUE #444: Use different random seeds for OpusTrainer teachers
I noticed the two teachers are training almost exactly under OpusTrainer, we should use the seed based on the teacher number from Taskcluster. Then OpusTrainer will shuffle the data differently. Those teachers are supposed to be different to impact the ensemble's quality.

eu9ene:
fixes #444


taskcluster/kinds/finetune-student/kind.yml
```diff
@@ -89,6 +89,7 @@ tasks:
                     $MOZ_FETCHES_DIR/corpus.aln.zst
                     None
                     None
+                    None
```
bhearsum, line 92: suggestion (non-blocking): set this to `0` where we don't care what it is. Based on what I see in https://firefox-ci-tc.services.mozilla.com/tasks/a-vmSSwNRd61tKZOI3Bnxw/runs/0/logs/live/public/logs/live.log, it looks like it gets converted to that anyways - but this will guard against any future changes that, eg: cause opustrainer not to accept the string `None` as a valid seed.

bhearsum (approved):

bhearsum:
I imagine @gregtatum may want to review this as well.

taskcluster/kinds/finetune-student/kind.yml
eu9ene: What you see in the config `seed: 0` is not OpusTrainer's random seed but Marian's one. And it's a good catch! We should change this one as well. In this case, I agree it would be easier to just make it required and pass any other number like 1111 or 0 if we don't care about it.

eu9ene:
looking at CI this couldn't affect `evaluate-quantized`. It was probably broken after refactoring.

taskcluster/kinds/finetune-student/kind.yml
gregtatum: Suggestion (optional): If we're confused about seeds it would be nice to be specific with something like `marian-seed`, although I see that the confusion is from the log from opustrainer itself, which we do not control. I'm unsure if you can pipe the output of another command into another logger so that it reports as `[opustrainer]` but that is beyond the scope of this PR.

gregtatum (approved):

taskcluster/kinds/finetune-student/kind.yml
eu9ene: I think having it as just `seed` is fine, we are just supposed to use it in all the random operations involved in training. We log opus trainer config separately as artifact.

PR #461, opened 2024-02-22
Allow group traversal from training tasks dependencies during Taskcluster task group publication

### LINKED ISSUE #312: Publish past experiment directly from Taskcluster

### LINKED ISSUE #448: Traversal tasks from Taskcluster task group publication
Discussed on #312, the idea is to traverse task groups related to the main task group and publish everything.

Single Taskcluster task group publication has been done in #406, you can work from branch `vrigal:taskcluster-publication` directly.

The default behavior of `translations_parser.cli.taskcluster_group` must follow https://github.com/mozilla/firefox-translations-training/issues/312#issuecomment-1919799719. This should be the default behavior, an argument `disable_traversal_groups` should allow to disable it.

EvaBardou:
Closes #448 

I've added a `--no-recursive-lookup` command option to @vrigal's implementation from #406. 

I've produced results with and without group traversal using the same top level task group ID `aY25-4fXTcuJNuMcWXUYtQ`:
-  `parse_tc_group aY25-4fXTcuJNuMcWXUYtQ` => **WandB** results are available [here](https://wandb.ai/teklia/en-hu-PR-461-recursive-lookup-enabled).
- `parse_tc_group aY25-4fXTcuJNuMcWXUYtQ --no-recursive-lookup` => **WandB** results are available [here](https://wandb.ai/teklia/en-hu-PR-461-recursive-lookup-disabled).

### DIFF
--- a/tracking/translations_parser/cli/taskcluster_group.py
+++ b/tracking/translations_parser/cli/taskcluster_group.py
@@ -291,6 +291,7 @@ def list_dependent_group_ids(task_id: str, known: set[str]):
         yield group_id
         known.add(group_id)
 
+        # Shared instance of `known` to propagate discovered groups in real time across all recursion branches
         yield from list_dependent_group_ids(dependent_task_id, known)
 
 
--- a/tracking/translations_parser/cli/taskcluster_group.py
+++ b/tracking/translations_parser/cli/taskcluster_group.py
@@ -205,10 +205,15 @@ def publish_task_group(group_id: str) -> None:
     group_name = f'{experiment["name"]}_{group_id}'
 
     grouped_tasks = list_completed_tasks(group_id)
+    training_tasks = list_training_tasks(group_id, grouped_tasks)
     metrics_tasks = list_metrics_tasks(group_id, grouped_tasks)
 
+    if not training_tasks:
+        logger.warning(f"Skipping task group {group_id} as it is empty")
+        return
+
     # Publish training tasks as runs
-    for training_task in list_training_tasks(group_id, grouped_tasks):
+    for training_task in training_tasks:
         # Associate metrics to each runs (evaluate tasks that depends on the training task)
         dependent_tasks = []
         for eval_id, eval_task in metrics_tasks.items():
--- a/tracking/translations_parser/cli/taskcluster_group.py
+++ b/tracking/translations_parser/cli/taskcluster_group.py
@@ -41,6 +41,11 @@ def get_args() -> argparse.Namespace:
         help="ID of the Taskcluster training task group.",
     )
     parser.add_argument(
+        "--no-recursive-lookup",
+        help="Disable group traversal from provided group_id tasks dependencies.",
+        action="store_true",
+    )
+    parser.add_argument(
         "--verbose",
         "-v",
         help="Print debug messages.",
@@ -54,6 +59,7 @@ def get_args() -> argparse.Namespace:
 def get_logs(task: dict) -> list[str]:
     """Retrieve training logs from Taskcluster"""
     task_id = task["status"]["taskId"]
+
     logger.info(f"Downloading logs for task {task_id}")
     log, _ = downloadArtifactToBuf(
         taskId=task_id,
@@ -81,11 +87,14 @@ def publish_task(project: str, group: str, name: str, task: dict, metrics: list[
 
 def get_metrics_from_task(task: dict) -> list[Metric]:
     task_id = task["status"]["taskId"]
+
     logger.info(f"Retrieving artifacts from evaluation task {task_id}")
+
     metrics = []
     for artifact in queue.listLatestArtifacts(task_id)["artifacts"]:
         if not artifact["name"].endswith(".metrics"):
             continue
+
         log, _ = downloadArtifactToBuf(
             taskId=task_id,
             name=artifact["name"],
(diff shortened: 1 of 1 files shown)


tracking/translations_parser/cli/taskcluster_group.py
```diff
@@ -95,109 +104,186 @@ def get_metrics_from_task(task: dict) -> list[Metric]:
...
-    config = task_group["extra"]["action"]["context"]["input"]
-    experiment = config["experiment"]
-    project = f"{experiment['src']}-{experiment['trg']}"
-    group_name = f"{experiment['name']}_{args.group_id}"
 
-    logger.info(f"Listing completed tasks from group {args.group_id}")
-    resp = queue.listTaskGroup(args.group_id)
-    tasks = resp["tasks"]
-    continuation_token = resp.get("continuationToken")
+def get_training_tasks(group_id: str, grouped_tasks: dict[str, list[dict]]) -> list[list[dict]]:
```
vrigal, line 137: I would rename this to `list_training_tasks`. Same for above function

tracking/translations_parser/cli/taskcluster_group.py
```diff
@@ -95,109 +104,186 @@ def get_metrics_from_task(task: dict) -> list[Metric]:
...
+
+    # Browse task dependencies
+    for dependent_task_id in task['dependencies']:
+        dependent_status = queue.status(dependent_task_id)
+
+        group_id = dependent_status["status"]["taskGroupId"]
+        if group_id in known:
+            continue
+
+        known.add(group_id)
```
vrigal, line 265: I think it is more explicit to yield each group, and iterate at above level adding elements to the set. \
Passing objects by references seems hard to follow reading code IMO.

vrigal (commented):
Great. I saw many variable have been renamed, which makes changes a bit harder to track. Those namings are much better otherwise, thanks.

tracking/translations_parser/cli/taskcluster_group.py
EvaBardou: Fixed in 0df72095f51f41de3e8d9434797a66dea560289f

tracking/translations_parser/cli/taskcluster_group.py
EvaBardou: Fixed in 0df72095f51f41de3e8d9434797a66dea560289f

tracking/translations_parser/cli/taskcluster_group.py
```diff
@@ -95,109 +104,200 @@ def get_metrics_from_task(task: dict) -> list[Metric]:
...
+    # Browse task dependencies
+    for dependent_task_id in task["dependencies"]:
+        dependent_status = queue.status(dependent_task_id)
+
+        group_id = dependent_status["status"]["taskGroupId"]
+        if group_id in known:
+            continue
+
+        yield group_id
+        known.add(group_id)
```
vrigal, line 271:
```suggestion
```

vrigal (approved):

tracking/translations_parser/cli/taskcluster_group.py
```diff
@@ -95,109 +104,200 @@ def get_metrics_from_task(task: dict) -> list[Metric]:
...
+
+        yield from list_dependent_group_ids(dependent_task_id, known)
```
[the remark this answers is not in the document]
EvaBardou: I've tried out your suggestion, but this considerably increases the execution time of the command.

**Before adding the suggestion:**
```
parse_tc_group aY25-4fXTcuJNuMcWXUYtQ  4,67s user 1,29s system 26% cpu 22,851 total
```

**After:**
```
parse_tc_group aY25-4fXTcuJNuMcWXUYtQ  25,43s user 1,56s system 19% cpu 2:15,33 total
```

This is because we don't share the knowledge (discovered groups), in real time, across recursion branches. As discussed, I've only added a comment (see e627864d635b5e9399a937bd0a75dbcc4b79ab60) to clarify this behavior and kept the original implementation as-is.

vrigal (approved):

eu9ene:
I think we should skip uploading if the group is empty for now (meaning it doesn't include any W&B runs except the `group_logs`). It's not very useful to have them even though they include the config in the group logs but makes it harder to navigate the experiments.

Screenshot 2024-02-27 at 11 59 53 AM

eu9ene (changes requested):
It looks good overall, but there are some things we should fix before merging.

In addition to the comment about removing the empty groups, please update the unit tests to reflect the changes in this PR.

EvaBardou:
Hi @eu9ene, thanks a lot for the review!

> It looks good overall, but there are some things we should fix before merging.
> 
> In addition to the comment about removing the empty groups, please update the unit tests to reflect the changes in this PR.

If I'm not mistaken, it seems that this command is not being tested at all. @vrigal mentioned that it would be complicated to mock the Taskcluster queue in #406 description:
>  I wasn't able to add a test. Mocking Taskcluster queue was complex and I prefer work on other important topics.

What should I do then?

EvaBardou:
> I think we should skip uploading if the group is empty for now (meaning it doesn't include any W&B runs except the `group_logs`). It's not very useful to have them even though they include the config in the group logs but makes it harder to navigate the experiments.

Fixed in 015587036da7906636e9be2d0590d4bbf359b1f6

New results for the `parse_tc_group aY25-4fXTcuJNuMcWXUYtQ` command are available [here](https://wandb.ai/teklia/en-hu-PR-461-skip-empty-groups).

eu9ene:
> Hi @eu9ene, thanks a lot for the review!
> 
> > It looks good overall, but there are some things we should fix before merging.
> > In addition to the comment about removing the empty groups, please update the unit tests to reflect the changes in this PR.
> 
> If I'm not mistaken, it seems that this command is not being tested at all. @vrigal mentioned that it would be complicated to mock the Taskcluster queue in #406 description:
> 
> > I wasn't able to add a test. Mocking Taskcluster queue was complex and I prefer work on other important topics.
> 
> What should I do then?

Ok, let's skip testing for now. The main scenario will be online tracking anyway.

eu9ene (approved):
Let's update with the main branch and it's good to go.

EvaBardou:
> Let's update with the main branch and it's good to go.

Done

PR #464, opened 2024-02-22
Set TERM for docker worker images

bhearsum:
This is to improve issues on docker worker interactive tasks where curses applications don't work correctly.

This is not quite perfect - but it's an improvement. There's seems to be a UI issue in Taskcluster still, where it mixes up x and y size for docker-worker sessions, which leads to a wrongly sized vim window and weird wrapping. I intend to fix that, but that's outside of the scope of this repo.

### DIFF
--- a/taskcluster/docker/base/Dockerfile
+++ b/taskcluster/docker/base/Dockerfile
@@ -31,6 +31,7 @@ RUN apt-get update -qq \
                           htop \
                           vim \
                           nano \
+                          ncurses-term \
                           zstd \
                           screen \
                           curl \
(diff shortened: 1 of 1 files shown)


bhearsum:
I ended up cancel the taskgraph here to save resources - no need to run the entire pipeline to verify this small change to the base docker image.

bhearsum:
> This is to improve issues on docker worker interactive tasks where curses applications don't work correctly.
> 
> This is not quite perfect - but it's an improvement. There's seems to be a UI issue in Taskcluster still, where it mixes up x and y size for docker-worker sessions, which leads to a wrongly sized vim window and weird wrapping. I intend to fix that, but that's outside of the scope of this repo.

I fixed the Taskcluster issue in https://github.com/taskcluster/taskcluster/pull/6856; that will take a couple of weeks to make it to the Firefox CI instance.

gabrielBusta:
Nice.

gabrielBusta (approved):

PR #467, opened 2024-02-26
Add shufflers as utilities

gregtatum:
I wrote these utilities in order to deal with shuffling large datasets and not having to load everything into memory, via python. They aren't hooked up to anything yet.

### DIFF
--- a/pipeline/common/datasets.py
+++ b/pipeline/common/datasets.py
@@ -1,0 +1,180 @@
+import os
+import tempfile
+from collections import deque
+from io import TextIOWrapper
+from random import Random
+from typing import Iterator, Optional
+
+
+def shuffle_with_max_lines(
+    line_stream: Iterator[str],
+    seed: str,
+    max_lines: int,
+    max_words_in_sentence,
+    total_byte_size: int,
+) -> Iterator[str]:
+    """
+    Shuffle a line stream, but only retain up to a maximum number of lines in memory.
+    Note that the final ordering is determined by the seed and the contents of the file. So
+    running this multiple times on the same dataset will return the same result, but running
+    it with the same seed and different content will create a different ordering.
+
+    Only run for monolingual data or where the parallel sentences are separated by a delimiter.
+
+    The distribution should be even unless the initial content is not representative of the
+    general size of the sentences, in this case the distribution will be slightly biased. See
+    the test cases for more in-depth examples.
+    """
+    lines = deque()
+
+    random = Random(seed)  # Make this deterministic based on dataset key.
+
+    total_bytes = 0
+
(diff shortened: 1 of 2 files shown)


eu9ene (commented):
I guess this will be handy if we decide to work with monolingual data from the common crawl.

I'm pretty sure you can do a similar thing with several lines using [Dask](https://docs.dask.org/en/latest/generated/dask.dataframe.DataFrame.shuffle.html) as it's designed to work with data larger than memory.

An example code from ChatGPT (not tested):

```
import dask.dataframe as dd

# Step 1: Read the file into a Dask DataFrame
# Replace 'your_file.csv' with the path to your file
df = dd.read_csv('your_file.csv')

# Step 2: Shuffle the DataFrame
# This step may take some time and resources depending on the size of your dataset
df = df.sample(frac=1, replace=False).reset_index(drop=True)

# Step 3: Take a sample of N lines
# Replace 'N' with the number of lines you want to sample
sampled_df = df.head(N)

# Optionally, you can compute the sample to get a pandas DataFrame
sampled_df = sampled_df.compute()

# If you want to save the sampled data to a file
sampled_df.to_csv('sampled_file.csv', index=False)
```

It's up to you, but I suggest minimizing the amount of custom code we need to maintain and using out-of-the-shelf libs instead where possible. What do you think?

gregtatum:
> It's up to you, but I suggest minimizing the amount of custom code we need to maintain and using out-of-the-shelf libs instead where possible. What do you think?

I spent a bit of time reading the Dask docs and it's a bit unclear how the shuffling works in practice without more testing and diving into it. The docs aren't very comprehensive there. I think at least in the short term I'd like to use a few shuffling primitives like I have here as it will be easy to compose them with streaming iterators. It's not a huge amount of code in terms of lines of code, and I wrote tests. Hopefully the maintenance burden will be small, and we can revisit the Dask decision if they aren't working well.

I've already included them in my follow-ups to the URL datasets.

eu9ene:
> > It's up to you, but I suggest minimizing the amount of custom code we need to maintain and using out-of-the-shelf libs instead where possible. What do you think?
> 
> I spent a bit of time reading the Dask docs and it's a bit unclear how the shuffling works in practice without more testing and diving into it. The docs aren't very comprehensive there. I think at least in the short term I'd like to use a few shuffling primitives like I have here as it will be easy to compose them with streaming iterators. It's not a huge amount of code in terms of lines of code, and I wrote tests. Hopefully the maintenance burden will be small, and we can revisit the Dask decision if they aren't working well.
> 
> I've already included them in my follow-ups to the URL datasets.

Ok, sounds good, let's stick with this code if it gives you more flexibility in usage.

eu9ene (approved):

PR #475, opened 2024-03-11
Add a taskcluster downloader for models

gregtatum:
This adds a little utility for downloading models. I wrote this for the en-ca model. I included a commit for some small changes to the existing script including a default output path, and listing all of the download modes.

You may have feedback on part of the patch, but I'm using the download utilities from the pipeline scripts, as I wanted a nice download function here.


eu9ene (approved):
It's a good idea to extend the current downloading script

PR #477, opened 2024-03-14
Analysis task that provides the word distribution

gregtatum:
This adds an analysis task for monolingual datasets for the distribution of the word count and codepoint count.

* [Task](https://firefox-ci-tc.services.mozilla.com/tasks/U5BHcGqnTp2BQTl9IN-W6w)
* [Codepoints graph](https://firefoxci.taskcluster-artifacts.net/U5BHcGqnTp2BQTl9IN-W6w/0/public/build/distribution-codepoints.png)
* [Words graph](https://firefoxci.taskcluster-artifacts.net/U5BHcGqnTp2BQTl9IN-W6w/0/public/build/distribution-words.png)

I just added the monolingual datasets, but can follow up with the parallel datasets.

It's not very ergonomic to view the tasks and find these, but I plan on following up and adding this to the W&B report.


bhearsum (approved):
I'm not sure how much my review is worth - but the Taskcluster bits look fine!

eu9ene (approved):

PR #478, opened 2024-03-15
Run tests with toolchain binaries under Docker

### LINKED ISSUE #456: Run tests under Docker
We should create a Dockerfile that can be built locally and be close to the production environment. We should be able to either compile or pull tools like `fast_align` and `marian` to run integration tests. 

We should run tests locally and in CI under this Dockerfile.

eu9ene:
Add a test for alignments and run it using the toolchain on CI. 

Compile the toolchain locally to run tests under Docker. Downloading some toolchain binaries from Taskcluster works but Marian has to be complied locally without GPU support and using arm64 on Apple Silicon, so it's easier to just compile the whole toolchain.

fixes #456


tests/test_alignments.py
```diff
@@ -0,0 +1,37 @@
...
+        "BIN": bin_dir,
+        "MARIAN": marian_dir,
+        "COMPRESSION_CMD": "zstd",
+        "ARTIFACT_EXT": "zst",
+        "SRC": "en",
+        "TRG": "ru",
+    }
+    shutil.copyfile("tests/data/vocab.spm", os.path.join(data_dir.path, "vocab.spm"))
+
+    data_dir.run_task("alignments-en-ru", env=env)
```
eu9ene, line 32: Running this task requires compiled `fast_align`, `extract_lex` and `marian/*` binaries.

taskcluster/kinds/alignments/kind.yml
```diff
@@ -70,7 +70,7 @@ tasks:
                     $VCS_PATH/pipeline/alignment/generate-alignment-and-shortlist.sh
                     $MOZ_FETCHES_DIR/corpus
                     $MOZ_FETCHES_DIR/vocab.spm
-                    /builds/worker/artifacts
+                    $TASK_WORKDIR/artifacts
```
bhearsum, line 73: This looks like an unrelated change; can you remove it?

bhearsum (approved):
I don't have an opinion about this, but it all looks largely fine.

taskcluster/kinds/alignments/kind.yml
bhearsum: Ah; I see. I was unfamiliar with the fixtures we're using until now.

bhearsum (approved):

docker/Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+RUN git clone https://github.com/clab/fast_align.git fast_align
+RUN ./compile-fast-align.sh fast_align/build $(nproc)
+
+RUN git clone https://github.com/kpu/preprocess.git preprocess
+RUN ./compile-preprocess.sh preprocess/build $(nproc)
+
+RUN git clone https://github.com/marian-nmt/marian-dev.git marian-dev
+RUN ./compile-marian.sh marian-dev/build $(nproc) false
+
+ENV MARIAN=/builds/worker/tools/marian-dev/build
```
gregtatum, line 52: Issue: I don't know why mine wouldn't work but your does, but I'm getting a python doesn't exist error. I found a solution though:

```diff
diff --git a/docker/Dockerfile b/docker/Dockerfile
index e9deacd..dc0aafa 100644
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -50,3 +50,7 @@ RUN git clone https://github.com/marian-nmt/marian-dev.git marian-dev
 RUN ./compile-marian.sh marian-dev/build $(nproc) false
 
 ENV MARIAN=/builds/worker/tools/marian-dev/build
+
+# Work around poetry not finding `python`.
+# https://github.com/python-poetry/poetry/issues/6371
+RUN ln -s /bin/python3 /bin/python
```

Here is the error:

```
 => CACHED [10/13] RUN git clone https://github.com/kpu/preprocess.git preprocess                                                                                                                    0.0s
 => CACHED [11/13] RUN ./compile-preprocess.sh preprocess/build $(nproc)                                                                                                                             0.0s
 => CACHED [12/13] RUN git clone https://github.com/marian-nmt/marian-dev.git marian-dev                                                                                                             0.0s
 => CACHED [13/13] RUN ./compile-marian.sh marian-dev/build $(nproc) false                                                                                                                           0.0s
 => exporting to image                                                                                                                                                                               0.0s
 => => exporting layers                                                                                                                                                                              0.0s
 => => writing image sha256:72aff1a69e0b4667b695d2cdb07c3010e4006814908dd8be68a5bd9b7ff96f17                                                                                                         0.0s
 => => naming to docker.io/library/ftt-local                                                                                                                                                         0.0s

What's Next?
  View summary of image vulnerabilities and recommendations → docker scout quickview
if [ -n "$VIRTUAL_ENV" ]; then \
		echo "Error: Virtual environment detected. Exit the poetry shell."; \
		exit 1; \
	fi && \
	if [ $(uname -m) == 'arm64' ]; then \
		echo "setting arm64 platform"; \
	  	export DOCKER_DEFAULT_PLATFORM=linux/amd64; \
	fi && \
	docker run \
		--interactive \
		--tty \
		--rm \
		--volume $(pwd):/builds/worker/checkouts \
		--workdir /builds/worker/checkouts \
		 ftt-local make run-tests
setting arm64 platform
poetry install --only tests --only utils --no-root
PYTHONPATH=$(pwd) poetry run pytest tests -vv

[Errno 2] No such file or directory: 'python'

[Errno 2] No such file or directory: 'python'
make: *** [Makefile:57: run-tests] Error 1
make: *** [run-tests-docker] Error 2
```

gregtatum, line 52: Suggestion (follow-up): It would be nice to cache the poetry install in the docker image, as currently you have to install it every time you run it locally. I think we can `ADD` the `poetry.lock` and `pyproject.toml` then run the install.

tests/test_alignments.py
```diff
@@ -0,0 +1,37 @@
+import os
+import shutil
+
+from fixtures import DataDir, en_sample, ru_sample
+
+current_folder = os.path.dirname(os.path.abspath(__file__))
```
gregtatum, line 6: Question: We have been a bit inconsistent with UPPER_CASE and lower_case variables at the root level. Do you have a preference?

tests/test_alignments.py
```diff
@@ -0,0 +1,37 @@
...
+        "TRG": "ru",
+    }
+    shutil.copyfile("tests/data/vocab.spm", os.path.join(data_dir.path, "vocab.spm"))
+
+    data_dir.run_task("alignments-en-ru", env=env)
+
+    shortlist_path = os.path.join(data_dir.path, "artifacts", "lex.s2t.pruned.zst")
+    assert os.path.exists(shortlist_path)
+    aln_path = os.path.join(data_dir.path, "artifacts", "corpus.aln.zst")
+    assert os.path.exists(aln_path)
```
gregtatum, line 37: Praise: As always, thanks for the test!

gregtatum (approved):
Thanks for splitting this out! I found one issue for my local use of this, but otherwise it worked great. I left one follow-up idea on adding the poetry install to a Docker layer.

docker/Dockerfile
eu9ene: I tried that but we install the tracking package from a source code and I don't want to include the code in the image. When mounting as a volume it's way faster to rerun the tests without rebuilding the image. We can think how to deal with it later, maybe testing the tracking differently without installing the package can work.

tests/test_alignments.py
eu9ene: usually upper case is for constants and lower case is for variables

eu9ene:
All suggestions were addressed

Issue #391, opened 2024-01-24
Consider switching to Docker for all tasks


eu9ene:
It would be easier to maintain the docker images for all tasks in this repo compared to updating the generic worker image elsewhere every time we need to add something. 

Also it would easier to reproduce running the tasks in a different environment, for example on a local machine or on premises cluster.

It's very common to run GPU workloads under Nvidia Docker toolkit https://docs.nvidia.com/ai-enterprise/deployment-guide-vmware/0.1.0/docker.html

bhearsum:
We're already running docker tasks on generic-worker machine for CPU tasks. I did a quick test to see if its trivial to do this, but alas, https://github.com/mozilla/firefox-translations-training/pull/619 ran into some odd issues - it looks like generic-worker is not interpreting the payload as a docker-worker one. My read is that the "insecure" generic-worker engine that the GPU workers run doesn't support d2g, while the CPU workers use the "multi" engine that does support it. (I'm confirming this theory.)

Assuming this is true, we'll need to switch the GPU workers to the "multi" engine before we can proceed here. That will probably require some Release SRE involvement; I don't know how much work it is.

bhearsum:
Taskcluster added support for d2g in the insecure engine: https://github.com/taskcluster/taskcluster/pull/7031. It looks like they may have a follow-up to deal with there, but once it's stablized we'll need to roll out a new GPU worker image with an updated version of generic-worker, and then we can give this a try again.

bhearsum:
This is likely to get fixed as part of #700.

Issue #264, opened 2023-11-17
Consolidate on how we run python


gregtatum:
We have both Conda, which was part of the original project, and I introduced poetry with some of the CI tooling. We should consolidate on some kind of tooling solution so we don't have multiple confusing errors like in #263. I was less familiar with Conda, and had trouble getting it working on my machine when I started working on the project. When implementing some CI features, I used poetry due to both my familiarity with it.

Our python packages management should solve the following issue:

1. Provide a lock file for dependency management.
2. Create a virtual environment so that modules can be installed in a predictable way.
3. Work locally when running our utils.
4. Work in CI and in Taskcluster.
5. Supporting pyproject.toml is a plus.
6. It should _practically_ allow us to install and manage dependencies.

## Conda

1. There is a plugin for this: https://github.com/conda/conda-lock
2. This is supported.
3. I had some trouble getting it setup, but maybe this was just me.
4. This is what's used in Taskcluster, but not CI.
5. I [guess this is a thing](https://docs.conda.io/projects/conda/en/latest/user-guide/concepts/conda-plugins.html#packaging-using-a-pyproject-toml-file). I'm not sure how good it is though.
6. Conda is really used in lots of ML-related work so it is a very practical solution. It's one of the few package managers that support PyTorch workflows, not that we are using PyTorch here.

## Poetry

1. Yes, by default.
2. Yes, with `poetry run`
3. Yes, although it's a little awkward putting things in the `Makefile`, and then having to remember to use the `poetry` command correctly locally when switching between different scripts with different dependencies.
4. CI uses it, but Taskcluster does not.
5. Yes, by default.
6. I'm running into some issues with the dogmatic, standards based approach of poetry. For instance, the `mtdata` dependency requires a specific `requests` version, while other dependencies require different version of `requests`. There is no way to work around this issue other than forking `mtdata`. See https://github.com/python-poetry/poetry/issues/697 . In addition to this issue, when I use poetry on side projects, PyTorch can't be installed as it uses non-standardized module installation. This is somewhat concerning to me since I want to align on the practices in the ML community at large.

## pyenv and pip

It would probably be worth filling in this one to consider.

eu9ene:
We actually use even more ways of managing environment: 

For python packages:
[...]

PR #480, opened 2024-03-15
Fix preflight check util

eu9ene:
- fix config path
- make it compatible with earlier versions of python (we use a variety of python versions and there's no reason to make it incompatible)


gregtatum (approved):
Thanks for the fix! I'm a bit sad I can't use the new syntax here. Question: What's the story on our oldest version of python we support? Is there an opportunity to increase it to get access to newer language features?

eu9ene:
> Thanks for the fix! I'm a bit sad I can't use the new syntax here. Question: What's the story on our oldest version of python we support? Is there an opportunity to increase it to get access to newer language features?

It has to do with inconsistent environments we run this in (see https://github.com/mozilla/firefox-translations-training/issues/264). Ideally we should have just two docker images with everything installed (CPU and GPU ones, see also https://github.com/mozilla/firefox-translations-training/issues/391) that we can also run locally and have just one python version for everything. I see some of the prod tasks run with Python version: 3.10.12.

IMO in general new Python versions come out periodically and usually, they're not worth switching to unless you do some low-level Python optimization. All the new syntactic sugar doesn't help much with our problems. The bottlenecks we have are usually not in python but in IO, CPU/GPU, the amount of processed data etc. If say, there is a groundbreaking optimization in python that improves something like memory allocation 2x+, it's worth considering. For now I would even recommend downgraing poetry env to 3.10 to be consistent with production.

Also, I don't see a big difference in using switch syntax vs standard for python if else one.

Issue #474, opened 2024-03-08
Backward steps are duplicated


eu9ene:
<img width="526" alt="Screenshot 2024-03-08 at 9 20 39 AM" src="https://github.com/mozilla/firefox-translations-training/assets/2486505/ec078564-6d3d-45f7-962d-87a3dd461542">

La0:
@eu9ene We could skip the group re-publication if it already exists when publishing with recursive lookups enabled.

This duplication happens because parts of the same graph are shared amongst some of the runs from your list (which makes sense).

As we run several times the code against different task groups, the only way to avoid duplication is to check if the group already exists on W&B (could be a pre-flight step in the code, to list all available groups)

eu9ene:
I see, ok, then we should check before uploading.

eu9ene:
More duplication examples:

<img width="472" alt="Screenshot 2024-03-18 at 2 29 31 PM" src="https://github.com/mozilla/firefox-translations-training/assets/2486505/f0f7d42d-58cf-4141-bc6f-c821c1596751">

PR #483, opened 2024-03-18
Avoid duplicated backward runs publication

vrigal:
Part of #474


eu9ene:
@vrigal like we discussed there seems to be a wider problem of duplicating other steps. Bastien suggested testing that the group already exists before uploading. We can't close #474 until those issues are addressed too.

vrigal:
I could test this code with task group `FKykM4toScus1wnKQhDn_g`, results are published [here](https://wandb.ai/teklia/483).
Only `backwards` remains (name of the training task, as for finetune).

I opened #484 for the overall publication.

eu9ene (approved):

PR #484, opened 2024-03-19
Avoid duplicated runs during W&B publication

### LINKED ISSUE #474: Backward steps are duplicated

vrigal:
Closes #474


eu9ene (approved):
We should make sure to distinguish the use case when re-uploading is unintended from the use case when we change something in the code and we want to re-upload everything. I guess the new deletion CLI argument will help with that.

PR #489, opened 2024-03-21
Add tests for training

### LINKED ISSUE #488: Add an integration test for training
We're having issues with alignments and it's really hard to debug them. Now when we compile Marian and run tests under Docker it should be possible to run the whole `train-student` task in the tests.

It also requires adding OpenBLAS and compiling Marian with it under Docker to work on Apple Silicon.

eu9ene:
- add a test with mocked Marian to check OpusTrainer's output and validate alignments
- add a full integration test with Marian training (it's quite fast, maybe takes 10 sec)

fixes #488

### DIFF
--- a/pipeline/train/train.sh
+++ b/pipeline/train/train.sh
@@ -93,6 +93,12 @@ paste <(${COMPRESSION_CMD} -dc "${valid_set_prefix}.${src}.${ARTIFACT_EXT}") \
       >"${valid_tsv_dataset}"
 
 
+# we run a CPU version of Marian in tests and it does not work with these arguments
+if [[ -z ${USE_CPU+x} ]]; then
+  extra_params+=('--sharding')
+  extra_params+=('local')
+fi
+
 echo "### Training ${model_dir}"
 # OpusTrainer reads the datasets, shuffles, augments them and feeds to stdin of Marian
 opustrainer-train \
@@ -114,7 +120,6 @@ opustrainer-train \
     --shuffle batches \
     --no-restore-corpus \
     --valid-reset-stalled \
-    --sharding local \
     --sync-sgd \
     --quiet-translation \
     --overwrite \
--- a/taskcluster/kinds/tests/kind.yml
+++ b/taskcluster/kinds/tests/kind.yml
@@ -102,7 +102,6 @@ tasks:
         - >-
             export BIN=$MOZ_FETCHES_DIR &&
             export MARIAN=$MOZ_FETCHES_DIR &&
-            export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:$MOZ_FETCHES_DIR/cuda-toolkit/lib64" &&
             make run-tests
     run-on-tasks-for: ["github-push", "github-pull-request"]
     optimization:
@@ -115,8 +114,7 @@ tasks:
         toolchain:
             - fast-align
             - extract-lex
-            - marian
-            - cuda-toolkit
+            - marian-cpu
 
   taskgraph-definition:
       worker-type: b-cpu
--- a/taskcluster/kinds/tests/kind.yml
+++ b/taskcluster/kinds/tests/kind.yml
@@ -102,6 +102,7 @@ tasks:
         - >-
             export BIN=$MOZ_FETCHES_DIR &&
             export MARIAN=$MOZ_FETCHES_DIR &&
+            export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:$MOZ_FETCHES_DIR/cuda-toolkit/lib64" &&
             make run-tests
     run-on-tasks-for: ["github-push", "github-pull-request"]
     optimization:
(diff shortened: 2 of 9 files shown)


pipeline/train/train.sh
```diff
@@ -93,6 +93,12 @@ paste <(${COMPRESSION_CMD} -dc "${valid_set_prefix}.${src}.${ARTIFACT_EXT}") \
       >"${valid_tsv_dataset}"
 
 
+# we run a CPU version of Marian in tests and it does not work with these arguments
+if [[ -z ${USE_CPU+x} ]]; then
```
bhearsum, line 97: What's the purpose of using the `x` default value if this is set? It seems like you could just use plain `[[ -z ${USE_CPU} ]]` here?

bhearsum (approved):

tests/fixtures/__init__.py
```diff
@@ -263,15 +268,14 @@ def find_pipeline_script(commands: Union[list[str], list[list[str]]]) -> str:
         print(command)
         raise Exception("Unable to find a string in the nested command.")
 
-    match = re.search(r"\/pipeline\/([\w\/-]+)\.(py|sh)", command)
-    #                               ^^^^^^^^^^   ^^^^^
-    #                               |            |
-    #    Match the path to the script            Match the extension
+    # $VCS_PATH/taskcluster/scripts/pipeline/train-taskcluster.sh
+    # $VCS_PATH/pipeline/alignment/generate-alignment-and-shortlist.sh
+    match = re.search(r"[\$\w\/]*\/pipeline\/[\w\/-]+\.(py|sh)", command)
```
gregtatum, line 273: Suggestion: I would prefer to maintain regex documentation on the logical groups of the regex, as this will be easier to maintain. Regexes are a frequent source of bugs and confusion. This is a good candidate for `r"""` multiline regexes with comments.

tests/fixtures/marian
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env python3
+"""
+marian test fixture
+
+Do not rely on marian in tests. This mocks marian by outputing the training corpus
```
gregtatum, line 5: Suggestion: I found this comment a little confusing, as we can now rely on Marian in tests. Maybe renaming it to "Marian can be mocked for tests when the actual binary doesn't need to be run.".

tests/test_training.py
```diff
@@ -0,0 +1,146 @@
...
+    }
+    data_dir.run_task("alignments-en-ru", env=env)
+    shutil.copyfile(
+        os.path.join(data_dir.path, "artifacts", "corpus.aln.zst"), data_dir.join("corpus.aln.zst")
+    )
+
+
+def test_train_student_mocked(alignments, data_dir):
+    """
+    Run training with mocked marian to check OpusTrainer output
```
gregtatum, line 78: Praise: Thanks for the nice explainer on the assumption for what this test is testing.

tests/test_training.py
```diff
@@ -0,0 +1,146 @@
...
+        "TEST_ARTIFACTS": data_dir.path,
+        "BIN": bin_dir,
+        "MARIAN": marian_dir,
+        "COMPRESSION_CMD": "zstd",
+        "ARTIFACT_EXT": "zst",
+        "SRC": "en",
+        "TRG": "ru",
+        "USE_CPU": "true",
+    }
+    marian_args = [
```
gregtatum, line 117: Nitpick (optional): You can disable black formatting for a nicer display of arg lists like so:

```
    marian_args = [
        "--disp-freq", "1",
        "--save-freq", "2",
        "--valid-freq", "2",
        "--after-batches", "2",
        "--dim-vocabs", "1000", "1000",
        "--mini-batch", "10",
        "--maxi-batch", "10",
        "--mini-batch-fit", "false",
        "--log-level", "trace",
    ] # fmt:skip
```

gregtatum (approved):
Thanks! The code changes look good so I'm going to go ahead and approve. I'd like to test this locally still before you merge just to verify it works on my machine. I can do that here in an hour or so.

pipeline/train/train.sh
eu9ene: it fails without the default value if the variable is not set

eu9ene:
Ah, it fails on CI! I need to add a toolchain to compile Marian for CPU to use on a CPU machine that runs tests

eu9ene:
Ok, I addressed the suggestions and added compiling Marian to run on a CPU machine. Now the tests pass.

taskcluster/kinds/tests/kind.yml
```diff
@@ -114,7 +114,7 @@ tasks:
         toolchain:
             - fast-align
             - extract-lex
-            - marian
+            - marian-cpu
```
gregtatum, line 117: 👍

gregtatum (approved):
It's working locally for me too!

PR #491, opened 2024-03-21
Update Marian to v1.12.14 2d067afb 2024-02-16

eu9ene:
A student model that failed before is training now: https://firefox-ci-tc.services.mozilla.com/tasks/C6H0G9F_RrSx0zJokDIb_Q/runs/0/logs/live/public/logs/live.log

I did not test the whole pipeline with the new Marian and it will take time to test, but I think we should merge it anyway because the [issue](https://github.com/mozilla/firefox-translations-training/issues/469) is blocking and the upgrade appears to fix it.

fixes #256 
fixes #469

[skip ci]

PR #492, opened 2024-03-22
Add Marian server for model testing

eu9ene:
You can chat with your model now!

```
$ python utils/marian_client.py
Hey model, how are you doing?
Модель, как дела?
```

Depends on #491 

[skip ci]


utils/marian_client.py
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env python3
+"""
+A client that connects to a Marian server and translates text interactively.
```
sylvestre, line 3: Great feature.
I am wondering if it would be hard to implement it without a local server?

eu9ene: Do you mean as a standalone app? It's not hard, but it must be in C++ since Marian doesn't have Python bindings.

utils/marian_client.py
```diff
@@ -0,0 +1,47 @@
...
+from __future__ import division, print_function, unicode_literals
+
+import argparse
+import sys
+
+from websocket import create_connection
+
+if __name__ == "__main__":
+    # handle command-line options
+    parser = argparse.ArgumentParser()
```
gregtatum, line 18: Suggestion: Passing in the docs here will display them via argparse.

```suggestion
    parser = argparse.ArgumentParser(
        description=__doc__,
        formatter_class=argparse.RawTextHelpFormatter,  # Preserves whitespace in the help text.
    )
```

Makefile
```diff
@@ -5,6 +5,10 @@ SHELL=/bin/bash
 
 # task group id for downloading evals and logs
 LOGS_TASK_GROUP?=
+# An ID of a Taskcluster task with a Marian model in the artifacts
+MODEL_TASK?=
```
gregtatum, line 9: Suggestion (blocking): We may need to discuss this outside of a PR review, but the Makefile approach is not scaling well for what we are adding to it.

Makefiles have the following issues:

1. Our implementation is monolithic, with one file having many concerns.
1. Makefiles provide no built in way to list out the commands that are available.
1. The commands do not have a CLI with documentation and argument parsing.
1. There is no way to pass arguments to the underlying scripts other than grafting on environment variables, which is a second API that isn't documented and not parsed nicely.

I've been feeling the friction of this when trying to interact with our tools. I think it's time we address it. We can make our python utilities all the entry point into our scripts, and provide reasonable defaults for the arguments values.

gregtatum (changes requested):
I'm leaving some blocking feedback on this PR. I think we should probably meet and discuss it in our 1:1 on Monday if you are up for it. I'd like to still test this out locally and finish up a proper review, but since I had early blocking design feedback I haven't done that yet.

Makefile
eu9ene: As we discussed live, since it's not clear yet what's the best way to move forward with restructuring our tooling and we want to land this code, I created a follow-up issue: https://github.com/mozilla/firefox-translations-training/issues/495.

gregtatum: I re-requested review for myself and would like to check things out locally, and read through the implementation one more time.

gregtatum:
I'm getting a build error for Marian both on macOS and on Ubuntu.

```
483.1 [ 98%] Linking CXX executable ../marian-vocab
485.1 [ 98%] Built target marian_vocab
491.4 [ 98%] Linking CXX executable ../marian-decoder
496.3 [ 98%] Built target marian_decoder
496.7 [ 98%] Linking CXX executable ../marian-conv
497.0 [ 98%] Linking CXX executable ../marian-scorer
499.9 [ 98%] Built target marian_conv
501.4 [ 98%] Built target marian_scorer
502.9 cc1plus: all warnings being treated as errors
503.0 make[2]: *** [src/CMakeFiles/marian_server.dir/build.make:76: src/CMakeFiles/marian_server.dir/command/marian_server.cpp.o] Error 1
503.0 make[1]: *** [CMakeFiles/Makefile2:702: src/CMakeFiles/marian_server.dir/all] Error 2
503.0 make[1]: *** Waiting for unfinished jobs....
520.6 [ 98%] Linking CXX executable ../marian
526.2 [ 98%] Built target marian_train
526.2 make: *** [Makefile:156: all] Error 2
------
Dockerfile:55
--------------------
  53 |     # Use the same revision as in taskcluster/kinds/fetch/toolchains.yml
  54 |     RUN cd marian-dev && git checkout e8a1a2530fb84cbff7383302ebca393e5875c441
  55 | >>> RUN ./compile-marian.sh marian-dev/build $(nproc) false
  56 |
  57 |     ENV MARIAN=/builds/worker/tools/marian-dev/build
--------------------
ERROR: failed to solve: process "/bin/sh -c ./compile-marian.sh marian-dev/build $(nproc) false" did not complete successfully: exit code: 2
make: *** [Makefile:69: build-docker] Error 1
```

eu9ene:
> > I'm getting a build error for Marian both on macOS and on Ubuntu. It's the same error on both machines.
> > ```
> > 483.1 [ 98%] Linking CXX executable ../marian-vocab
> > 485.1 [ 98%] Built target marian_vocab
> > 491.4 [ 98%] Linking CXX executable ../marian-decoder
> > 496.3 [ 98%] Built target marian_decoder
> > 496.7 [ 98%] Linking CXX executable ../marian-conv
> > 497.0 [ 98%] Linking CXX executable ../marian-scorer
> > 499.9 [ 98%] Built target marian_conv
> > 501.4 [ 98%] Built target marian_scorer
> > 502.9 cc1plus: all warnings being treated as errors
> > 503.0 make[2]: *** [src/CMakeFiles/marian_server.dir/build.make:76: src/CMakeFiles/marian_server.dir/command/marian_server.cpp.o] Error 1
> > 503.0 make[1]: *** [CMakeFiles/Makefile2:702: src/CMakeFiles/marian_server.dir/all] Error 2
> > 503.0 make[1]: *** Waiting for unfinished jobs....
> > 520.6 [ 98%] Linking CXX executable ../marian
> > 526.2 [ 98%] Built target marian_train
> > 526.2 make: *** [Makefile:156: all] Error 2
> > ------
> > Dockerfile:55
> > --------------------
> >   53 |     # Use the same revision as in taskcluster/kinds/fetch/toolchains.yml
> >   54 |     RUN cd marian-dev && git checkout e8a1a2530fb84cbff7383302ebca393e5875c441
> >   55 | >>> RUN ./compile-marian.sh marian-dev/build $(nproc) false
> >   56 |
> >   57 |     ENV MARIAN=/builds/worker/tools/marian-dev/build
> > --------------------
> > ERROR: failed to solve: process "/bin/sh -c ./compile-marian.sh marian-dev/build $(nproc) false" did not complete successfully: exit code: 2
> > make: *** [Makefile:69: build-docker] Error 1
> > ```
> 
> @gregtatum The output doesn't have enough information to say what's wrong in your case. It has to do with adding the flag to build the server. The docker env is supposed to be reproducible. Make sure the image is being rebuilt. Also, what's your docker memory limit? It's worth increasing if it's less than 16Gb.
> 
> I'll try to rebuild everything from scratch with --no-cache and see if I can reproduce this.

@gregtatum I can reproduce it. Sorry, I was a bit ahead and used the updated Marian for this. It will work with this PR merged https://github.com/mozilla/firefox-translations-training/pull/491/files

gregtatum:
Ah that makes since. I was looking into it for a bit, and it was some deprecation errors. The upgrade should fix it, I'll try it again.

gregtatum (approved):
Ok, I got it working on my machine. Looks good to me, we will follow up on #495.

PR #496, opened 2024-03-26
Skip checking existing runs for new projects

vrigal:
Follow up of #484

Avoids a ValueError raised by the wandb client when a project (e.g. `en-ru`) does not exists yet on the wandb namespace.


tracking/translations_parser/publishers.py
```diff
@@ -103,7 +103,8 @@ def open(self, parser) -> None:
         config.update(self.extra_kwargs.pop("config", {}))
 
         # Ensure no W&B run already exists
-        if name := self.extra_kwargs.get("name"):
+        project = next(filter(lambda p: p.name == self.project, wandb.Api().projects()), None)
```
eu9ene, line 106: Why those w&b API calls are outside of try-except block? Don't we want to catch all the errors here and not propagate them to the calling code?

eu9ene (approved):

PR #498, opened 2024-03-26
Allow overriding results from Taskcluster group on W&B

### LINKED ISSUE #366: Support overriding experiment on Weight&Biases
A CLI argument should allow overriding runs (instead of simply duplicating them). This would make iterations easier too because a manual cleanup is required actually in case of an update.

It is not documented in the [Python client](https://docs.wandb.ai/ref/python/), but seems possible using `wandb.Api().run("<entity>/<project>/<run_id>").delete()` (see https://github.com/wandb/wandb/issues/1745#issuecomment-779185936).

Refs. #164

vrigal:
Closes #366


eu9ene:
@vrigal there is a conflict with the recently merged code

eu9ene (commented):
@vrigal I see that you delete some runs based on the name matching. Maybe I'm not following the logic completely but when we reupload the experiment with overriding, all the runs of the old experiment should be deleted to prevent corrupted data. For example, some steps might be renamed in the future etc.

vrigal:
Alright, I updated code on thois PR directly. Your suggestion makes the implementation much simpler!

eu9ene (approved):

PR #500, opened 2024-03-26
Parallel stats

gregtatum:
This adds the distribution stats to the parallel datasets as well. I initially only did the mono datasets to make the first implementation simpler. Here you will find the graphs are named after the datasets.


tests/fixtures/__init__.py
```diff
@@ -161,23 +161,24 @@ def run_task(
             for part in command_parts
         ]
 
-        if extra_args:
```
gregtatum, line 164: These changes were required to support `&&` in the bash commands for the `run_task` abstraction.

gregtatum:
Ah, I forgot to add this as a CI task.

gregtatum:
Ok, it's working now, here as an example task: [analyze-corpus-opus-ada83_v1-ru-en](https://firefox-ci-tc.services.mozilla.com/tasks/FaBfDMq3S5SGbafX1GN3IA)

taskcluster/kinds/analyze-corpus/kind.yml
```diff
@@ -0,0 +1,83 @@
...
+        run:
+            using: run-task
+            command:
+                - bash
+                - -c
+                - >-
+                    pip3 install --upgrade pip setuptools &&
+                    pip3 install -r $VCS_PATH/pipeline/data/requirements/analyze.txt
+                    &&
+                    python3 $VCS_PATH/pipeline/data/analyze.py
```
eu9ene, line 64: Maybe it works for now but I assume we'll want to do some analysis based on parallel sentences, like the ratio between the number of words etc. We might want to tune filters based on those per language. In this case, the analyze script would need to accept either two files as input.

eu9ene (approved):
Approving with an optional suggestion to accept two files as input for the script to make it more future-proof. This would also make the code related to handling ampersands go away.

bhearsum (approved):
Looks good from a Taskcluster standpoint.

jcristau (approved):

Issue #333, opened 2024-01-03
Publish training charts from Taskcluster


eu9ene:
This includes publishing:
- live training logs to W&B dashboards

I assume we'll have separate publishing scripts for other things.

Let's use [Taskgraph transforms](https://taskcluster-taskgraph.readthedocs.io/en/latest/concepts/transforms.html) not to pollute Taskcluster kinds with even more logic.

La0:
We can start working on this now.

I identified these next steps to discuss together:

1. build a dedicated command in the tracking package to support specific tasks (training & evaluation) and use a Taskcluster secret to get the Weight & Biases token
2. Setup the publication code through a Taskcluster transform
3. Patch training shell scripts to capture output and publish from logs
   - when an env variable is set
   - scripts to patch 
      - `pipeline/train/train.sh`
      - `pipeline/train/spm-vocab.sh`
4. Identify other tasks to track & publish

La0:
After discussion with Evgeny, we can start to build the dedicated script to interact with Taskcluster secret & logs.

eu9ene:
One small correction: `pipeline/train/spm-vocab.sh` trains a vocab and we don't want to track this in W&B. So only `train.sh` should be patched.

La0:
@evabardou will work on [Taskcluster secret support](https://github.com/mozilla/firefox-translations-training/issues/481).

We can start experimenting log parsing from CI using `--in-stream` option too

PR #501, opened 2024-03-27
Taskcluster publication

La0:
Refs #333

This PR add code to run `parse_tc_logs` using live logs from teacher trainer task.

The call to the python script is configured through a Taskcluster transform `use-secret` which brings
- Taskcluster proxy
- Taskcluster secret scope
- Taskcluster secret environment variable used by `parse_tc_logs`

When the secret is available and the `parse_tc_logs` has been installed through `pip install ./training`, the command will run against the logs to track & publish metrics.

If the command crashes (invalid or missing credentials for example), it will cleanly exit with exit code 0 to prevent crashing the training too.

Taskcluster header parsing has been disabled when using `--from-stream` as these headers are NOT present when running from inside the task.

### DIFF
--- a/pipeline/train/train.sh
+++ b/pipeline/train/train.sh
@@ -108,7 +108,7 @@ if ! command -v parse_tc_logs &> /dev/null
 then
   echo "### Weight & Biases publication script is not available."
   PARSER=cat
-elif ! -z "$TEST_ARTIFACTS"
+elif [ ! -z ${TEST_ARTIFACTS+x} ];
 then
   echo "### Weight & Biases publication is disabled for unit tests."
   PARSER=cat
--- a/pipeline/train/train.sh
+++ b/pipeline/train/train.sh
@@ -102,6 +102,16 @@ if [[ -z ${USE_CPU+x} ]]; then
   extra_params+=('local')
 fi
 
+# Enable log & metrics publication only when the tracking script is available
+if ! command -v parse_tc_logs &> /dev/null
+then
+  echo "### Weight & Biases publication is disabled."
+  PARSER=tee
+else
+  echo "### Weight & Biases publication is enabled."
+  PARSER="parse_tc_logs --from-stream --wandb-project=moz-translations"
+fi
+
 echo "### Training ${model_dir}"
 # OpusTrainer reads the datasets, shuffles, augments them and feeds to stdin of Marian
 opustrainer-train \
@@ -129,7 +139,7 @@ opustrainer-train \
     --keep-best \
     --tsv \
     --seed ${seed} \
-    "${extra_params[@]}"
+    "${extra_params[@]}" | $PARSER
 
 cp "${model_dir}/model.npz.best-${best_model_metric}.npz" "${model_dir}/final.model.npz.best-${best_model_metric}.npz"
 cp "${model_dir}/model.npz.best-${best_model_metric}.npz.decoder.yml" "${model_dir}/final.model.npz.best-${best_model_metric}.npz.decoder.yml"
--- a/pipeline/train/train.sh
+++ b/pipeline/train/train.sh
@@ -109,7 +109,7 @@ then
   PARSER=tee
 else
   echo "### Weight & Biases publication is enabled."
-  PARSER="parse_tc_logs --from-stream --wandb-project=moz-translations"
+  PARSER="parse_tc_logs --from-stream --wandb-project=moz-translations -v"
 fi
 
 echo "### Training ${model_dir}"
--- a/pipeline/train/train.sh
+++ b/pipeline/train/train.sh
@@ -106,10 +106,10 @@ fi
 if ! command -v parse_tc_logs &> /dev/null
 then
   echo "### Weight & Biases publication is disabled."
-  PARSER=tee
+  PARSER=cat
 else
   echo "### Weight & Biases publication is enabled."
-  PARSER="parse_tc_logs --from-stream -v"
+  PARSER="parse_tc_logs --from-stream -v --wandb-project=${src}-${trg}"
 fi
 
 echo "### Training ${model_dir}"
--- a/pipeline/train/train.sh
+++ b/pipeline/train/train.sh
@@ -104,7 +104,8 @@ fi
 
 # Enable log & metrics publication only when the tracking script is available
 # and we are not running unit tests
-if [[ ! $(command -v parse_tc_logs &> /dev/null) || ! -z "$TEST_ARTIFACTS" ]];
+# and it's not actively disabled through training configuration
+if [[ ! $(command -v parse_tc_logs &> /dev/null) || ! -z "$TEST_ARTIFACTS" || "${WANDB_PUBLICATION,,}" == "false" ]];
 then
   echo "### Weight & Biases publication is disabled."
   PARSER=cat
--- a/pipeline/train/train.sh
+++ b/pipeline/train/train.sh
@@ -109,7 +109,7 @@ then
   PARSER=tee
 else
   echo "### Weight & Biases publication is enabled."
-  PARSER="parse_tc_logs --from-stream --wandb-project=moz-translations -v"
+  PARSER="parse_tc_logs --from-stream -v"
 fi
 
 echo "### Training ${model_dir}"
--- a/pipeline/train/train.sh
+++ b/pipeline/train/train.sh
@@ -114,7 +114,7 @@ then
   PARSER=cat
 else
   echo "### Weight & Biases publication is available."
-  PARSER="parse_tc_logs --from-stream -v --wandb-project=${src}-${trg} --wandb-group ${model_type}.${training_type} --wandb-run-name $TASK_ID"
+  PARSER="parse_tc_logs --from-stream -v"
 fi
 
 echo "### Training ${model_dir}"
--- a/pipeline/train/train.sh
+++ b/pipeline/train/train.sh
@@ -103,7 +103,8 @@ if [[ -z ${USE_CPU+x} ]]; then
 fi
 
 # Enable log & metrics publication only when the tracking script is available
-if ! command -v parse_tc_logs &> /dev/null
+# and we are not running unit tests
+if [[ ! $(command -v parse_tc_logs &> /dev/null) || ! -z "$TEST_ARTIFACTS" ]];
 then
   echo "### Weight & Biases publication is disabled."
   PARSER=cat
--- a/pipeline/train/train.sh
+++ b/pipeline/train/train.sh
@@ -139,7 +139,7 @@ opustrainer-train \
     --keep-best \
     --tsv \
     --seed ${seed} \
-    "${extra_params[@]}" | $PARSER
+    "${extra_params[@]}" 2>&1 | $PARSER
 
 cp "${model_dir}/model.npz.best-${best_model_metric}.npz" "${model_dir}/final.model.npz.best-${best_model_metric}.npz"
 cp "${model_dir}/model.npz.best-${best_model_metric}.npz.decoder.yml" "${model_dir}/final.model.npz.best-${best_model_metric}.npz.decoder.yml"
--- a/pipeline/train/train.sh
+++ b/pipeline/train/train.sh
@@ -109,7 +109,7 @@ then
   PARSER=cat
 else
   echo "### Weight & Biases publication is enabled."
-  PARSER="parse_tc_logs --from-stream -v --wandb-project=${src}-${trg}"
+  PARSER="parse_tc_logs --from-stream -v --wandb-project=${src}-${trg} --wandb-group ${model_type}.${training_type} --wandb-run-name $TASK_ID"
 fi
 
 echo "### Training ${model_dir}"
--- a/pipeline/train/train.sh
+++ b/pipeline/train/train.sh
@@ -103,14 +103,17 @@ if [[ -z ${USE_CPU+x} ]]; then
 fi
 
 # Enable log & metrics publication only when the tracking script is available
-# and we are not running unit tests
-# and it's not actively disabled through training configuration
-if [[ ! $(command -v parse_tc_logs &> /dev/null) || ! -z "$TEST_ARTIFACTS" || "${WANDB_PUBLICATION,,}" == "false" ]];
+# and not running in unit tests
+if ! command -v parse_tc_logs &> /dev/null
 then
-  echo "### Weight & Biases publication is disabled."
+  echo "### Weight & Biases publication script is not available."
+  PARSER=cat
+elif ! -z "$TEST_ARTIFACTS"
+then
+  echo "### Weight & Biases publication is disabled for unit tests."
   PARSER=cat
 else
-  echo "### Weight & Biases publication is enabled."
+  echo "### Weight & Biases publication is available."
   PARSER="parse_tc_logs --from-stream -v --wandb-project=${src}-${trg} --wandb-group ${model_type}.${training_type} --wandb-run-name $TASK_ID"
 fi
 
--- a/taskcluster/kinds/train-teacher/kind.yml
+++ b/taskcluster/kinds/train-teacher/kind.yml
@@ -82,13 +82,20 @@ tasks:
             env:
                 ARTIFACT_EXT: zst
                 COMPRESSION_CMD: zstdmt
+
+                # Weight & Biases publication token is stored in that secret
+                TASKCLUSTER_SECRET: project/translations/level-1/weights-and-biases
             artifacts:
                 - name: public/build
                   path: artifacts
                   type: directory
 
-        # Use a secret to publish logs on weights and biases
-        use-secret: project/translations/level-1/weights-and-biases
+            # Taskcluster proxy is required to read secrets
+            taskcluster-proxy: true
+
+        # The task needs to be able to read that secret to publish on Weight & Biases
+        scopes:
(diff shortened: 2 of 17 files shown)


La0:
Got a [training task](https://firefox-ci-tc.services.mozilla.com/tasks/OZc_WFtKSc2MaNxq9OUOFg/definition) definition with 
- taskcluster proxy enabled
- tracking code install through pip
- taskcluster secret env variable set

pipeline/train/train.sh
```diff
@@ -129,7 +139,7 @@ opustrainer-train \
     --keep-best \
     --tsv \
     --seed ${seed} \
-    "${extra_params[@]}"
+    "${extra_params[@]}" 2>&1 | $PARSER
```
bhearsum, line 142: question: Do we actually want to combine stderr and stdout? I imagine the parser handles it fine, but I'm not sure if it has downsides for the `tee` case.

taskcluster/translations_taskgraph/parameters.py
```diff
@@ -43,9 +43,9 @@ def get_defaults(_):
                     "dim-vocabs": "1000 1000",
                 },
                 "training-teacher": {
-                    "disp-freq": "1",
-                    "save-freq": "25",
-                    "valid-freq": "50",
+                    "disp-freq": "3",
```
bhearsum, line 46: suggestion: We probably don't want to take this change (I think you needed it for testing though).

(I'll defer to @eu9ene on whether or not this is blocking.)

taskcluster/translations_taskgraph/transforms/worker_env.py
```diff
@@ -52,3 +61,23 @@ def inject_worker_env(config, jobs):
...
+                {
+                    "TASKCLUSTER_SECRET": secret_name,
+                }
+            )
+            job["worker"]["taskcluster-proxy"] = True
+            if "scopes" not in job:
+                job["scopes"] = []
+            job["scopes"] += [f"secrets:get:{secret_name}"]
+
+        yield job
```
bhearsum, line 83: suggestion: please put this in the `kind` instead. We've been trying to keep as much as possible in there to improve readability. It does increase repetition, but it makes it easier to map tasks back to kinds and make tweaks. (If you feel this must go in transform, please use a new one -- this one is _specifically_ for dealing with worker-type specific configuration, which this is not.)

bhearsum (changes requested):
I only reviewed the taskcluster parts. Let me know if you have any questions!

taskcluster/translations_taskgraph/parameters.py
```diff
@@ -45,7 +45,7 @@ def get_defaults(_):
                 "training-teacher": {
                     "disp-freq": "1",
                     "save-freq": "25",
-                    "valid-freq": "50",
+                    "valid-freq": "49",
```
eu9ene, line 48: This should be reverted, validation runs correctly once in a regular CI run

tracking/translations_parser/parser.py
```diff
@@ -177,13 +178,23 @@ def _iter_log_entries(self) -> Iterator[tuple[list[tuple[str]], str]]:
...
             # Record logs depending on Marian headers
             if len(headers) >= 2:
                 # First is task timestamp, second is marian timestamp
                 _, _, *marian_tags = headers
-                tag = "_".join(*marian_tags) if marian_tags else "_"
+                tag = _join(marian_tags)
                 self.indexed_logs[tag].append(text)
 
+            # Display parsed log text on stderr for easier debug
+            print(f"Marian log: {text.strip()}", file=sys.stderr)
```
eu9ene, line 196: Should this be removed? I think the idea was to just propagate the existing logging to stdout as.

eu9ene (changes requested):
I compared the [new train-teacher log](https://firefox-ci-tc.services.mozilla.com/tasks/OCGS_xxgQWSYqbJ09aFw2Q/runs/0/logs/public/logs/live.log) form CI with the[ previous one](https://firefox-ci-tc.services.mozilla.com/tasks/VGzXs7RnQWeaqfp5hhCKeg/runs/0/logs/public/logs/live.log). The wrapper should not modify the original log. It should stream stdout and stderr of the wrapped program as is and optionally add some logging records with the same convention. For example: `[tracking] [INFO] Reading logs stream.`

Also, it complains about missing validation here but in the original log we can see validation steps happing correctly:
```
[task 2024-04-15T15:48:33.948Z] [WARNING] Some validation data is incomplete with the following epoch/up couples:
[task 2024-04-15T15:48:33.948Z] [WARNING] * Ep. 1, Up. 49
[task 2024-04-15T15:48:33.948Z] [WARNING] * Ep. 1, Up. 50
[task 2024-04-15T15:48:33.949Z] [WARNING] No ValidationEpoch entry, skipping.
```

I did not find a test publication in https://wandb.ai/moz-translations/ru-en.

taskcluster/kinds/train-teacher/kind.yml
```diff
@@ -101,6 +104,7 @@ tasks:
                 - >-
                     pip3 install --upgrade pip setuptools &&
                     pip3 install -r $VCS_PATH/pipeline/train/requirements/train.txt &&
+                    pip3 install $VCS_PATH/tracking &&
```
eu9ene, line 107: Let's also add it to all other steps that use `train-taskcluster.sh` (train-*, finetune-*)

bhearsum:
> I compared the [new train-teacher log](https://firefox-ci-tc.services.mozilla.com/tasks/OCGS_xxgQWSYqbJ09aFw2Q/runs/0/logs/public/logs/live.log) form CI with the[ previous one](https://firefox-ci-tc.services.mozilla.com/tasks/VGzXs7RnQWeaqfp5hhCKeg/runs/0/logs/public/logs/live.log). The wrapper should not modify the original log. It should stream stdout and stderr of the wrapped program as is and optionally add some logging records with the same convention. For example: [tracking] [INFO] Reading logs stream.

I wonder if this is related to my comment about stdout & stderr getting combined? I imagine that separating those (and handling both streams) might result in better behaviour. Even so - I'm not certain whether or not a downstream consumer can guarantee that they will process messages from two separate pipes in the same order they would've been printed to the console. (Maybe it is possible...I honestly don't know.)

eu9ene:
> > I compared the [new train-teacher log](https://firefox-ci-tc.services.mozilla.com/tasks/OCGS_xxgQWSYqbJ09aFw2Q/runs/0/logs/public/logs/live.log) form CI with the[ previous one](https://firefox-ci-tc.services.mozilla.com/tasks/VGzXs7RnQWeaqfp5hhCKeg/runs/0/logs/public/logs/live.log). The wrapper should not modify the original log. It should stream stdout and stderr of the wrapped program as is and optionally add some logging records with the same convention. For example: [tracking] [INFO] Reading logs stream.
> 
> I wonder if this is related to my comment about stdout & stderr getting combined? I imagine that separating those (and handling both streams) might result in better behaviour. Even so - I'm not certain whether or not a downstream consumer can guarantee that they will process messages from two separate pipes in the same order they would've been printed to the console. (Maybe it is possible...I honestly don't know.)

We discussed live that it's ok to merge those two streams because they are intentionally used by OpusTrainer to output stuff. I guess it was used to separate Marian log and OpusTrainer's own log. Unless there's a way to easily redirect both separately using bash piping.

eu9ene:
Some other things we discussed should be implemented before merging:
- we should investigate warnings with validation publishing. Indeed I don't see validation graphs on https://wandb.ai/moz-translations/moz-translations
- we should add a parameter to the training config to disable publishing on CI and when we debug something and don't want to pollute the dashboards
- as in the previous comment, preserving log format and adding [tracking] prefix for the parser
- using the proper project/experiment name for publishing right away as we'll start using this when it's merged (should be `ru-en/ci_<task_cluster_id>` for the test CI run when enabled
- making sure all the training and fine-tuning steps are tracked

bhearsum:
> Some other things we discussed should be implemented before merging:
>     * we should add a parameter to the training config to disable publishing on CI and when we debug something and don't want to pollute the dashboards

Apologies that I missed the meeting. One possible alternative here, if you want, might be to have a separate publishing location (some sort of sandbox) that we use for these cases.

eu9ene:
> > Some other things we discussed should be implemented before merging:
> > * we should add a parameter to the training config to disable publishing on CI and when we debug something and don't want to pollute the dashboards
> 
> Apologies that I missed the meeting. One possible alternative here, if you want, might be to have a separate publishing location (some sort of sandbox) that we use for these cases.

It's a good idea. We could create a staging namespace for testing, I just don't know if we want to always publish there. The namespace can be called something like `moz-translations-staging` and then we can specify the namespace in the config. An empty value can mean that we don't want to publish anywhere. @La0 we can use an environment variable for this. It will be easier than propagating the argument through many layers. This namespace will be our sandbox for testing before rolling out to production. If everyone is ok with this approach I'll ask the MLOps team to create it for us. cc @gregtatum

eu9ene:
> > > Some other things we discussed should be implemented before merging:
> > > 
> > > * we should add a parameter to the training config to disable publishing on CI and when we debug something and don't want to pollute the dashboards
> > 
> > 
> > Apologies that I missed the meeting. One possible alternative here, if you want, might be to have a separate publishing location (some sort of sandbox) that we use for these cases.
> 
> It's a good idea. We could create a staging namespace for testing, I just don't know if we want to always publish there. The namespace can be called something like `moz-translations-staging` and then we can specify the namespace in the config. An empty value can mean that we don't want to publish anywhere. This namespace will be our sandbox for testing before rolling out to production. If everyone is ok with this approach I'll ask the MLOps team to create it for us. cc @gregtatum
> 
> I think in the short term we can implement it and put an empty string into the CI config. Then when the new namespace is created we'll be able to retest the full reuploading of the old experiments there instead of Teklia project.

UPD: I looked at the code and I see that what W&B namespace we write to is controlled by the secret token, so to implement this we would need to deal with secrets. Let's leave it out of scope for this PR and just make sure we can disable publishing from the experiment config.

taskcluster/translations_taskgraph/transforms/worker_env.py
La0: I moved these settings to kind in 32f5c4e

taskcluster/translations_taskgraph/parameters.py
La0: I only set this to force-trigger a training task.

I explicitly added a `[TRASHME]` in the commit message to avoid merging the PR with this commit

tracking/translations_parser/parser.py
La0: I replaced this by outputting the raw line unmodified, and prefixing log lines from parser with `[tracking]`

eu9ene:
> @eu9ene Is this expected behaviour ? Can we set `perplexity` as optional ?

@La0 All the metrics that are used for validation are set in the Marian config (--valid-metrics), so they are all kind of optional. We usually don't use perplexity. We set them here https://github.com/mozilla/firefox-translations-training/blob/145a84ace322184b5c5d6e9aca0ececb037c5d08/pipeline/train/train.sh#L40

La0:
Last changes:
- I made `perplexity` optional
- A training parameter `wandb-publication` has been added, disabled by default for CI tasks. You can control it when starting a new training

pipeline/train/train.sh
```diff
@@ -102,6 +102,16 @@ if [[ -z ${USE_CPU+x} ]]; then
...
 fi
 
+# Enable log & metrics publication only when the tracking script is available
+if ! command -v parse_tc_logs &> /dev/null
+then
+  echo "### Weight & Biases publication script is not available."
+  PARSER=cat
+else
+  echo "### Weight & Biases publication is available."
+  PARSER="parse_tc_logs --from-stream -v --wandb-project=${src}-${trg} --wandb-group ${model_type}.${training_type} --wandb-run-name $TASK_ID"
```
eu9ene, line 112: @La0 naming should be consistent with the offline uploading. For group: `<experiment-name-from-training-config>_<taskcluster-group-id>`, for task: I think we parse the task label to get one of the following: finetune-student, student, teacher-1, teacher-2, backwards. Some examples:

Screenshot 2024-04-22 at 2 26 36 PM
Screenshot 2024-04-22 at 2 22 10 PM

eu9ene, line 112: It might be better to move this code to a transform where we can properly retrieve the data like Taskcluster task group id and the experiment ID for the CLI arguments. Ideally, this task should not depend on Taskcluster as we usually separate task implementation and the orchestrator-specific code. This job can potentially be launched locally or using a different orchestrator( snakemake etc.)

La0: @bhearsum was against transform usage (see comment [here](https://github.com/mozilla/firefox-translations-training/pull/501#discussion_r1565928735)).

I'll probably move the naming arguments into the python code, as it will be easier to condition these parameters for taskcluster usage

La0: The naming is now fully handled by the tracking Python code, detecting if we are running in Taskcluster, in CI or normal runs, and producing the names using the group publication algorithm.

I setup default values for CI as the configuration is not exposed through task group in that case

bhearsum (approved):
Looks great from a taskgraph standpoint!

pipeline/train/train.sh
eu9ene: moving to the publisher is a good idea!

eu9ene: - [x] The group name looks good
- [x] The graphs look good
- [ ] The run names are still not consistent with the offline publisher

It's important to make naming consistent to compare the same model across older and newer training. See the names in en-ca: https://wandb.ai/moz-translations/en-ca

eu9ene: When this last issue is fixed, I'll run a train action to test how it displays graphs in real time and that it takes the values properly from the config

La0: Done in [37d227f](https://github.com/mozilla/firefox-translations-training/pull/501/commits/37d227f52bc7339e51d8fe6430357f1e2b47ca9a)

eu9ene:
I launched a test training run from this branch https://firefox-ci-tc.services.mozilla.com/tasks/SngAEYS-RyWnY4GgKJTzSQ

La0:
> I launched a test training run from this branch https://firefox-ci-tc.services.mozilla.com/tasks/SngAEYS-RyWnY4GgKJTzSQ

I do not see any training task in the group, maybe the cache did not allow your tasks to run ?

eu9ene:
> > I launched a test training run from this branch https://firefox-ci-tc.services.mozilla.com/tasks/SngAEYS-RyWnY4GgKJTzSQ
> 
> I do not see any training task in the group, maybe the cache did not allow your tasks to run ?

It's running https://firefox-ci-tc.services.mozilla.com/tasks/PDAfVzlWQMam4uqhndjafg. You're probably looking at the actions group which is different from the tasks group. Also, I ran it from a push task.

It looks working! https://wandb.ai/moz-translations/en-ru/groups/test_tracking_SngAEYS-RyWnY4GgKJTzSQ

La0:
I rebased on latest `main` commits, and removed the extra commit that tweaked training parameters.

tracking/translations_parser/cli/taskcluster.py
```diff
@@ -113,40 +163,60 @@ def main() -> None:
...
+    if project_name:
         publishers.append(
             WandB(
-                project=args.wandb_project,
+                project=project_name,
+                group=group_name,
+                name=run_name,
                 artifacts=args.wandb_artifacts,
-                group=args.wandb_group,
                 tags=["cli"],
```
eu9ene, line 191: I see that the tag "cli" is visible in the workgroup. I see we also use the tag "Taskcluster" in the code. Should we utilize tags to distinguish between different modes of publishing? We have 3 now: GCP, taskcluster-offline and taskcluster online.

eu9ene (approved):
I have a couple of questions/suggestions, otherwise it's good to go

tracking/translations_parser/cli/taskcluster.py
La0: I exposed the tags through a new CLI argument `--tags`:
- on CI uploads, it will use `taskcluster`
- on offline group upload using the dedicated entry point, it will use `taskcluster-offline`
- when uploading a local folder, the operator should set `GCP`

tracking/translations_parser/cli/taskcluster.py
```diff
@@ -96,12 +89,69 @@ def get_args() -> argparse.Namespace:
...
+    task = queue.task(task_id)
+    _, task_name = build_task_name(task)
+    group_id = task["taskGroupId"]
+    task_group = queue.task(group_id)
+    config = task_group.get("extra", {}).get("action", {}).get("context", {}).get("input")
+    if config is None:
+        logger.warn(
+            f"Experiment configuration missing on {group_id} @ extra/action/context/input, fallback to CI values"
+        )
+        experiment = {
```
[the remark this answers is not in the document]
La0: What should be the expected behaviour when running on CI without an explicit configuration ? We still need to produce weight & biases project, group & run names.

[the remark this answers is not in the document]
eu9ene: as discussed we can keep it for testing for now

eu9ene:
Ok, it looks good. If we find more issues or have more ideas on what to improve let's handle them in the follow-up tasks. We'll need to reupload the experiments anyway when it's feature complete. We need to make sure the publishing from CI is disabled though not to pollute the ru-en project.

PR #504, opened 2024-03-27
Update docs for OpusTrainer and alignments

eu9ene:
Depends on #502 

[skip ci]


gregtatum (approved):
Praise: Thank you for taking the time to rewrite the table. This makes it much easier to review and read while in code view mode.

PR #510, opened 2024-04-03
Remove the Makefile and replace it with a Taskfile

### LINKED ISSUE #495: Simplify and restructure the makefile and utilities
We increasingly use the Makefile as an entry point for our local CLI tools and it's getting more complex. We should find a better way to organize the tooling code, make it easier to understand arguments and run the CLI tools.

See also https://github.com/mozilla/firefox-translations-training/pull/492#discussion_r1536162389

A few possible alternatives we discussed with @gregtatum:
- Reimplement some of the scripts as standalone Python scripts (pros: easier to maintain, argument parsing, docs; cons: takes some time to rewrite, more code compared to bash, maybe an overkill for such supplementary code)
- extract bash to separate bash scripts (pros: makefile becomes smaller, easier to inspect arguments in the code, ability to run separately; cons: still no interactive help, no argument parsing)
- bash + something for argument parsing and help (pros: easier to use than just bash; cons: the bash code might become complicated)
- find an alternative to Makefile (here is one after a quick search: https://taskfile.dev/)

gregtatum:
Resolves #495.

It took a bit of work to get all the tasks to work consistently both in and out of docker, which lead to a bit of incidental changes with this work. Since our Docker configuration is so new I was having a bit of trouble getting everything to work on my machine. These changes are all broken out into logical commits.


utils/run_model.py
```diff
@@ -0,0 +1,252 @@
+#!/usr/bin/env python3
+"""
+Run m
```
eu9ene, line 3: incomplete comment

utils/run_model.py
eu9ene, line 1: There's also a use case when we want to test an external model (from older snakemake experiments or from statmt/hplt) so there should be a way to run this on a directory. Optional for this PR, we can add it later.

eu9ene (approved):
Nice, it's all organized a lot better now!

utils/run_model.py
gregtatum: Ah yeah, I agree. I filed a follow-up: https://github.com/mozilla/firefox-translations-training/issues/523

PR #512, opened 2024-04-03
Parse logs with Marian 1.12

### LINKED ISSUE #509: Update parser to work with Marian 1.12

vrigal:
Closes #509

This also allows parsing and publishing `perplexity` values.

Example of publication: https://wandb.ai/teklia/512-fi-en?nw=nwuserrigal


eu9ene (approved):
Looks good! Another idea for an integration test: we now have tests that run Marian under Docker: `test_training::test_train_student`. It's relatively fast. We could add another test that takes the log from this run and parses it. We can do this later.

PR #516, opened 2024-04-05
Update the docs for training continuation to use yaml

gregtatum:
For some reason the example was using `yml`. I assume my changes are correct for the implementation.


bhearsum (approved):
Both work (json is valid yaml), but this is much nicer!

eu9ene (approved):

gregtatum:
> json is valid yaml
🤯 TIL

PR #526, opened 2024-04-11
Fix parameters to use correct target tasks method

bhearsum:
As things are now, the `small` parameters will never generate training tasks. (The `large` params already use the correct target tasks method.)


hneiva (approved):
👍 lgtm

PR #530, opened 2024-04-17
Make downloads more robust

gregtatum:
This changes download function to have timeouts and retries. It also increases the amount of logging the function does so that we can diagnose issues in CI.

### DIFF
--- a/pipeline/common/downloads.py
+++ b/pipeline/common/downloads.py
@@ -134,7 +134,7 @@ class RemoteZstdLineStreamer(RemoteDecodingLineStreamer):
 
     Usage:
 
-        with RemoteGzipLineStreamer(url) as lines:
+        with RemoteZstdLineStreamer(url) as lines:
             for line in lines:
                 print(line)
     """
@@ -160,7 +160,7 @@ class DownloadChunkStreamer(io.IOBase):
              gzip.GzipFile(fileobj=f)
     """
 
-    def __init__(self, url: str, total_retries=20, timeout_sec=10.0, wait_before_retry_sec=1.0):
+    def __init__(self, url: str, total_retries=3, timeout_sec=10.0, wait_before_retry_sec=60.0):
         self.url = url
         self.response = None
 
--- a/pipeline/common/downloads.py
+++ b/pipeline/common/downloads.py
@@ -2,8 +2,9 @@ import gzip
 import io
 import json
 import os
+import time
 from io import BufferedReader
-from typing import Optional
+from typing import Generator, Optional
 
 import requests
 import zstandard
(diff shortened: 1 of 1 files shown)


gregtatum:
This adds logging to the download which appears like so:

```
[download-mono] Dataset: news-crawl_news.2007
[download-mono] Language: en
[download-mono] Max Sentences: 10000
[download-mono] Artifacts: /builds/worker/artifacts
[download-mono] File Destination: /builds/worker/artifacts/news_2007.en.zst
[download-mono] Downloading WMT newscrawl monolingual data
[download-mono] http://data.statmt.org/news-crawl/en/news.2007.en.shuffled.deduped.gz
[download-mono] URL: http://data.statmt.org/news-crawl/en/news.2007.en.shuffled.deduped.gz
[downloads] Download size: 184284260 bytes
[downloads] 5% downloaded (9216000/184284260 bytes)
[downloads] 10% downloaded (18432000/184284260 bytes)
[downloads] 15% downloaded (27648000/184284260 bytes)
[downloads] 20% downloaded (36864000/184284260 bytes)
[downloads] 25% downloaded (46071808/184284260 bytes)
[downloads] 30% downloaded (55287808/184284260 bytes)
[downloads] 35% downloaded (64503808/184284260 bytes)
[downloads] 40% downloaded (73719808/184284260 bytes)
[downloads] 45% downloaded (82935808/184284260 bytes)
[downloads] 50% downloaded (92143616/184284260 bytes)
[downloads] 55% downloaded (101359616/184284260 bytes)
[downloads] 60% downloaded (110575616/184284260 bytes)
[downloads] 65% downloaded (119791616/184284260 bytes)
[downloads] 70% downloaded (128999424/184284260 bytes)
[downloads] 75% downloaded (138215424/184284260 bytes)
[downloads] 80% downloaded (147431424/184284260 bytes)
[downloads] 85% downloaded (156647424/184284260 bytes)
[downloads] 90% downloaded (165863424/184284260 bytes)
[downloads] 95% downloaded (175071232/184284260 bytes)
[downloads] 100% downloaded - Download finished.
```

pipeline/common/downloads.py
```diff
@@ -85,47 +78,233 @@ def get_download_size(url: str) -> int:
...
 
-    def decode(self, response_stream):
-        return zstandard.ZstdDecompressor().stream_reader(response_stream)
+    Or pass it as a file handle:
+
+        with DownloadChunkStreamer(url) as f:
+             gzip.GzipFile(fileobj=f)
+    """
+
+    def __init__(self, url: str, total_retries=20, timeout_sec=10.0, wait_before_retry_sec=1.0):
```
eu9ene, line 163: IMO it's better to reduce the number of retries to say 3 but increase the retry intervals to say 60 sec. Libs like tenacity also allow retry policies to exponentially increase the interval each time but it's optional. With 20 retries with a 1-second interval, you risk exceeding the request rate set by the external resource and being blocked.

eu9ene (commented):
We should discuss the retry frequency.

Overall great job fixing but I would prefer to find off the shelf solution for most of this code, especially lower level to make our code less error-prone and reduce maintenance. For example `urllib.Retry` can help with retries and provides backoff factor functionality https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html#urllib3.util.Retry

pipeline/common/downloads.py
gregtatum: Thanks, I adjusted it.

gregtatum:
> Overall great job fixing but I would prefer to find off the shelf solution for most of this code, especially lower level to make our code less error-prone and reduce maintenance. For example urllib.Retry can help with retries and provides backoff factor functionality https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html#urllib3.util.Retry

So we discussed this on our 1:1, and chatted about the trade offs of getting an off the shelf solution, and using some complicated custom code that we will need to maintain.

In general I think we both agreed that this code is a little unfortunate in how complicated it is, but that it will serve as a good primitive to do downloads with iteration. This solves the particular problem of decoding, and getting a line iterator, all while having robust downloads that shouldn't fail in CI. I did look into the the `urllib` Retry implementation, but was struggling with using it with the way I was using iterators.

I'll re-flag you for review.

eu9ene (approved):

PR #532, opened 2024-04-18
allow runtime selection of worker classes through training config (fixes #300)

### LINKED ISSUE #300: allow for runtime selection of GPU workers
#277 is adding support for easily configuring workers at decision task time. This makes it easy to use different sets of workers on different branches. However, we also want to be able to use the same workers on the same branch, eg: to run multiple experiments for the same language pair at the same time. To do this, we'll need at least few things:
- Worker aliases defined in the `workers` block of `config.yml` for each set of workers
- Support for specifying the workers we want to use in the training config (needs changes to action.py and parameters.py at least)
- A new transform that will apply the correct `worker-type` to each GPU task based on what's in the training config
- Possibly some changes to the kinds that use GPU workers to remove the hardcoded `worker-type` from them (since the transform will just be overriding it anyways...).

bhearsum:
This patch adds a new `worker-classes` section of the training config which allows for selecting the "class" of worker to use (currently GCP spot or GCP standard) by kind, with support for a default. This allows us quite flexible configuration, eg: using spot instances for translation and standard ones for training. We will also be able to add one or more classes for Snakepit machines when we bring those online. The default value is `{"default": "gcp-spot"}` - which means that we'll use all spot machines by default. (We can change the default in `config.prod.yml` if desired.)

Most of this patch is quite a boring addition of the new `worker_selection` transform to the pipeline kinds. The most notable part otherwise is a fairly big rework of most things to do with workers in `config.yml`:
* The taskgraph-required `workers.aliases` is now a very simple, straightforward list of all available worker types.
* A new `local-worker-aliases` has been introduced. This maps the generic names like `b-largegpu` to concrete worker types `by-worker-class`. This removes the need for `-standard` variants for the generic names.
* This necessitated a new transform function that looks up the concrete worker type in each kind before we hand off to the `task` transforms. (Previously, that transform simply looked up things like `b-largegpu` in the `workers.aliases` block. If we had the ability to feed it `worker-class` information we could have kept all our mappings there - but I couldn't come up with a reasonable way to do this upstream.)

I have two relevant training runs here:
* https://firefox-ci-tc.services.mozilla.com/tasks/groups/AB7KJw7NSzi0NM_lRRcrlQ is what was fired when I first opened this PR, and uses the default parameters, ending up with spot instances for all tasks.
* https://firefox-ci-tc.services.mozilla.com/tasks/groups/VuefIANGTQ6-xjoZlrczxA is an additional one I triggered with:
```
        worker-classes:
          default: gcp-spot
          train-student: gcp-standard
          train-teacher: gcp-standard
```

...and as expected, the [student](https://firefox-ci-tc.services.mozilla.com/tasks/QRI0ptNWRcyDxvmcPmeCvw) and [teacher](https://firefox-ci-tc.services.mozilla.com/tasks/FIOLYFHQSgmhSPUzu_tTvg) got standard instances, while other [training](https://firefox-ci-tc.services.mozilla.com/tasks/fpemJEKyTieZlzJtPmvclQ) [tasks](https://firefox-ci-tc.services.mozilla.com/tasks/R2uDi7X3TEirbjXhFG_wtA) got spots.


bhearsum:
@gregtatum and/or @eu9ene - I imagine you'll want to have a look at this, at the very least for the training config bits (but all feedback is welcome, of course!).

taskcluster/configs/config.prod.yml
```diff
@@ -161,6 +161,8 @@ taskcluster:
   # After the parallel corpora are merged and de-duplicated, the combined file is
   # then split into an even number of chunks.
   split-chunks: 10
+  worker-classes:
+    default: gcp-spot
```
gregtatum, line 165: Suggestion: It would be helpful to document the available values here, or point to where they are defined (taskcluster/translations_taskgraph/actions/train.py)

taskcluster/translations_taskgraph/transforms/worker_selection.py
```diff
@@ -0,0 +1,71 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
gregtatum, line 1: I did not review these changes, as I assume a taskcluster peer will.

gregtatum (approved):
This is great! It will be much easier to reason with the setup here. I'll leave the taskcluster transform review for someone else to do, but the other changes look good to me.

taskcluster/configs/config.prod.yml
bhearsum: That's a good idea; will do.

taskcluster/test/params/large-lt-en.yml
```diff
@@ -163,4 +163,9 @@ training_config:
   target-stage: all
   taskcluster:
     split-chunks: 10
+    worker-classes:
+      train-backwards: gcp-standard
```
eu9ene, line 167: not sure if it's important here but there's also finetune-student

eu9ene (approved):
Looks good! I would probably add an example to the prod training config on how to use standard instances for training and spot ones for everything else.

taskcluster/test/params/large-lt-en.yml
bhearsum: It won't really make a difference to the consumers of this, but I may as well add it, as I meant this to be production-like.

bhearsum:
> Looks good! I would probably add an example to the prod training config on how to use standard instances for training and spot ones for everything else.

To be clear - you mean https://github.com/mozilla/firefox-translations-training/blob/main/taskcluster/configs/config.prod.yml ?

eu9ene:
> > Looks good! I would probably add an example to the prod training config on how to use standard instances for training and spot ones for everything else.
> 
> To be clear - you mean https://github.com/mozilla/firefox-translations-training/blob/main/taskcluster/configs/config.prod.yml ?

yes, it's our example config and I need to update it myself by the way

gabrielBusta (approved):

bhearsum:
Thanks for the reviews folks! I'm going to hold off landing this for the moment as it is currently based on the backed out #533. If I can't get that fixed in the next week or so, I'll rebase this against the current `main` and merge it.

PR #540, opened 2024-04-29
Update the training continuation docs

gregtatum:
This fixes the location of the `pretrained-models` and adds information of how to find our models. It also mentions the `vocab.spm` requirement.


docs/using-pretrained-models.md
```diff
@@ -14,20 +14,44 @@ Utilizing pretrained models can reduce training time and resource usage.
...
+    # Re-use an existing backwards model from a Google Cloud Storage bucket. This must
+    # be the original (non-quantized) student model.
+    train-backwards:
+      urls:
+        - https://storage.googleapis.com/releng-translations-dev/models/en-fi/opusmt/student/
+      mode: use
+      type: default
+```
+
+To find models from older training runs see the `gs://releng-translations-dev/models` bucket.
```
bhearsum, line 35: Note: the long term storage location for production training runs will be `moz-fx-translations-data--303e-prod-translations-data`. (I think we'll eventually end up migrating all of the data to this bucket and getting rid of `releng-translations-dev`, which was only intended as a short term workaround.)

PR #541, opened 2024-04-29
Always use vocab.spm from artifacts directory in training steps

bhearsum:
This ensures that it will be uploaded alongside models if we are spot terminated, which makes automatically resuming in a rerun much easier.

### DIFF
--- a/pipeline/train/train.sh
+++ b/pipeline/train/train.sh
@@ -102,6 +102,21 @@ if [[ -z ${USE_CPU+x} ]]; then
   extra_params+=('local')
 fi
 
+# Enable log & metrics publication only when the tracking script is available
+# and not running in unit tests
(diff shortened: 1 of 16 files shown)


taskcluster/scripts/pipeline/train-taskcluster.sh
```diff
@@ -38,11 +38,17 @@ case "$pretrained_model_mode" in
         ;;
     "continue"|"init"|"None")
         if [ "$pretrained_model_mode" == "None" ]; then
-            vocab="$MOZ_FETCHES_DIR/vocab.spm"
-        else
-            vocab="$TASK_WORKDIR/artifacts/vocab.spm"
```
bhearsum, line 43: Now that we're copying, the vocab location is the same for all three of these modes.

Issue #542, opened 2024-04-29
Support training continuation (pretrained models) for teacher ensembles


gregtatum:
The training_continuation.py doesn't correctly handle multiple URLs.

Steps to reproduce:

Set the config

```
experiment:
  pretrained-models:
    train-teacher:
      urls:
        - https://example.com/a
        - https://example.com/b
      mode: use
      type: default
```

Run `task preflight-check`

Open `artifacts/full-task-graph.json`.

Search for both `example.com/a` and `example.com/b` in the generated tasks. `train-teacher-XX-XX-1/2` and `train-teacher-XX-XX-2/2` both contain the same files, and the URLs aren't split between them.

PR #543, opened 2024-04-29
Add file overrides to the training continuation, and refactor the implementation

gregtatum:
Edit: I added file overrides for this, to support vocabs and other file name mismatches. For instance in `en-fi` the final model was not a `best-chrf` but a `best-perplexity`. This allows for working around problems in the config itself.

<hr>

I was looking into how training continuation was working, and was confused by some of the mis-direction in the code with iterators and dict comprehension. I refactored the code a bit to understand how things were working. I added some more validation with type-friendly dataclass and enum. I also wrote a few more docs on what was going on.

I didn't end up finishing a test for this, as I didn't want to spend more time on it, but I manually checked the artifacts/full-task-graph.json after running `task preflight-check`.

The code produces the equivalent mounts as the original code. I also filed #542 as I realized that ensemble training wasn't actually working.

### DIFF
--- a/taskcluster/translations_taskgraph/transforms/training_continuation.py
+++ b/taskcluster/translations_taskgraph/transforms/training_continuation.py
@@ -1,8 +1,17 @@
+from dataclasses import dataclass
+from enum import Enum
+from typing import Literal
 from taskgraph.transforms.base import TransformSequence
 from urllib.parse import urljoin
 import os
 
-CONTINUE_TRAINING_ARTIFACTS = (
+"""
+Transform jobs to be able to use pre-trained models.
+
+See docs/using-pretrained-models.md
+"""
+
+CONTINUE_TRAINING_ARTIFACTS = [
     "devset.out",
     "model.npz",
     "model.npz.best-bleu-detok.npz",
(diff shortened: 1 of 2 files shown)


taskcluster/translations_taskgraph/transforms/training_continuation.py
```diff
@@ -1,8 +1,17 @@
...
 import os
 
-CONTINUE_TRAINING_ARTIFACTS = (
+"""
+Transform jobs to be able to use pre-trained models.
+
+See docs/using-pretrained-models.md
+"""
+
+CONTINUE_TRAINING_ARTIFACTS = [
```
gregtatum, line 14: This was more type friendly as a list, for `get_artifact_names`

taskcluster/translations_taskgraph/transforms/training_continuation.py
```diff
@@ -20,64 +29,121 @@
...
+
+    This will download files such as:
+      "https://example.com/en-ru/model.npz.best-bleu-detok.npz"
+      "https://example.com/en-ru/model.npz.best-ce-mean-words.npz",
+      etc.
+    """
+
+    if len(pretrained_model.urls) != 1:
+        raise Exception(
+            "Multiple URLs are currently not supported for pretrained models. See Issue #542"
```
gregtatum, line 85: I filed a follow-up rather than fix it here.

eu9ene (commented):
Would it be possible to write a small unit test for the artifact selection logic as we added more functionality to it?

As I wrote in another issue the OPUS-MT anomaly was a one-time thing and we don't plan to support it in the immediate future, but the vocabs are usually in a separate folder for older models so the ability to explicitly specify the path adds some convenience. I just copied things to have the expected structure in the past.

Also, I just remembered that OPUS-MT models required some special preprocessing, but I guess our student should be fine. We should double-check this [PR](https://github.com/mozilla/firefox-translations-training/pull/192/files#diff-28708fbc337269dea5331ccd8e73fa8ad3318c5d7d8d4f9c9cdd4acefe94b09a).

gregtatum:
Yeah, I'll add a test. Originally I hadn't since it was just a refactor, but this now also adds functionality.

gabrielBusta (approved):
Looks good to me. FWIW - Ben is refactoring this `mounts` stuff on #546

bhearsum:
> Looks good to me. FWIW - Ben is refactoring this `mounts` stuff on #546

We'll see! I'm not sure if I will be refactoring that in advance, or following up later. In any case, this can land and I will deal with any rebasing needed in my patch(es).

gregtatum:
Rebasing this may be hard, and I'm not planning on doing it in the short term. It's still valid, and I may re-open it in the future.

PR #545, opened 2024-04-29
Always use vocab.spm from artifacts directory in training steps

bhearsum:
This ensures that it will be uploaded alongside models if we are spot terminated, which makes automatically resuming in a rerun much easier.


gabrielBusta (approved):

gregtatum (approved):
Seems reasonable to me.

PR #546, opened 2024-04-30
Wrap train-taskcluster.sh in train_taskcluster.py

bhearsum:
In order to support automatic continuation of spot terminated training runs (#270) I'll need to do some non-trivial things before handing off to `train.sh`. Rather than try to do them in bash, I'd like to do them in Python. This PR should have no behavioru differences, but it lays a foundation for this upcoming work. It would also be a good basis for replacing `train-taskcluster.sh` with Python entirely in the future.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -101,7 +101,7 @@ tasks:
         task test -- tests/test_alignments.py
     deps: [poetry-install-tests]
     cmds:
-      - PYTHONPATH=$(pwd) poetry run pytest -vv {{.CLI_ARGS}}
+      - PYTHONPATH="$(pwd):$(pwd)/taskcluster/scripts/pipeline" poetry run pytest -vv {{.CLI_ARGS}}
 
   test-docker:
     desc: Run the unit tests in the docker image. Some tests require the pre-built Linux executables.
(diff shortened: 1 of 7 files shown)


bhearsum:
> I'm concerned about the number of wrappings in our training. We have: Marian wrapped by OpusTrainer wrapped by `train.sh` wrapped by `train-taskcluster.sh` that will be wrapped by `train_taskcluster.py`...
> 
> I strongly recommend rewriting `train-taskcluster.sh` in python instead of adding another wrapping level to reduce complexity.

I'm a bit hesistant to do take that on as something that blocks turning spot instances back on, since we're on a tight timeline, and it's difficult to test all of the training continuation scenarios, but I can if you'd like.

Alternatively, I can follow up with it later, after we've branched for the upcoming trainings.

eu9ene:
> > I'm concerned about the number of wrappings in our training. We have: Marian wrapped by OpusTrainer wrapped by `train.sh` wrapped by `train-taskcluster.sh` that will be wrapped by `train_taskcluster.py`...
> > I strongly recommend rewriting `train-taskcluster.sh` in python instead of adding another wrapping level to reduce complexity.
> 
> I'm a bit hesistant to do take that on as something that blocks turning spot instances back on, since we're on a tight timeline, and it's difficult to test all of the training continuation scenarios, but I can if you'd like.
> 
> Alternatively, I can follow up with it later, after we've branched for the upcoming trainings.

Sure, it's up to you, fixing this as a follow-up works for me.

jcristau (approved):

eu9ene (approved):
I'm ok with merging this to unblock spot instances but please add a TODO and an issue to rewrite this in Python to reduce wrapping and complexity.

PR #547, opened 2024-04-30
Custom cleaning

### LINKED ISSUE #210: Add full support of the custom filters for OpusCleaner
As a follow-up for the [initial integration](https://github.com/mozilla/firefox-translations-training/pull/163) we should:
- Use the OpusCleaner UI and produce the full training config for all datasets for a language pair
- Figure out where to store all the filter configs (either this repo and being able to add a directory to TC resources or probably better using a separate repo). We can also discuss this with HPLT folks.
- Test training with the full config
- Compare quality to training with the default config
- Refactor and document things if needed (see comments for the [PR](https://github.com/mozilla/firefox-translations-training/pull/163))

eu9ene:
- update default OpusCleaner filters
- add per dataset custom filters (harmful filters were removed after inspection in OpusCleaner UI)
- pre-download the fast_text model to prevent errors on downloading through OpusCleaner
- add a config option to switch between using custom and default settings.

I'm running another baseline en-ru teacher training with these adjustments and I see that the defaults outperform the custom filters that I tuned on backward models and where they showed positive results. I didn't use NLLB for this other experiment though to make it faster. So I'm adding an option to use the defaults for now until we can show better results with custom filters on the experiments.

fixes #210


eu9ene:
CI looks good. finetune-student fails because of the #549

taskcluster/translations_taskgraph/actions/train.py
```diff
@@ -143,6 +143,13 @@ def validate_pretrained_models(params):
                     "use-opuscleaner": {
                         "type": "string",
                         "description": "use OpusCleaner to clean corpus",
+                        "enum": ["true", "false"],
+                    },
+                    "opuscleaner-mode": {
```
bhearsum, line 148: Thanks for remembering to update this & parameters.py.

bhearsum (approved):

taskcluster/translations_taskgraph/actions/train.py
eu9ene: sure, I filed #597 though :)

gregtatum (approved):
The changes all seem reasonable to me. 👍

PR #548, opened 2024-04-30
Add a preflight check for URL mounts

gregtatum:
I had my training fail because of this, so I thought it would be nice to add a preflight check for bad URLs for pretrained models.

I included a drive-by that disables the PLW0603 ruff rule, as I found it was mostly causing false positives where I was using global scopes in a way that I felt was reasonable. We had a few suppressions in the codebase, and everything felt like valid code to me. If you disagree with this I can remove the commit and add a suppression to my global statement.


gregtatum:
Here is an example of the check:

image

eu9ene (approved):

PR #555, opened 2024-05-02
Add missing wandb_publication parameter on finetune-student task.

### LINKED ISSUE #553: CI pollutes W&B
See https://wandb.ai/moz-translations/ru-en, the runs appear to be fresh.

La0:
Closes #553

### DIFF
--- a/taskcluster/kinds/finetune-student/kind.yml
+++ b/taskcluster/kinds/finetune-student/kind.yml
@@ -38,6 +38,7 @@ tasks:
                 - fetches.train-student
                 - run.command
                 - attributes
+                - worker.env
         attributes:
             stage: finetune-student
             src_locale: "{src_locale}"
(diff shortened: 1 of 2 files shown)


bhearsum (approved):
Should we be default this to False instead, or is the risk of pollution better than the risk of not publishing at all? (@eu9ene or @gregtatum - that's probably a question for you.)

eu9ene:
> Should we be default this to False instead, or is the risk of pollution better than the risk of not publishing at all? (@eu9ene or @gregtatum - that's probably a question for you.)

Yes, pollution is less risky than losing data :)

eu9ene:
However, I think we should not publish anything if the env variable is not set. I can imagine us adding more steps that publish to W&B and forgetting to set it.

eu9ene (changes requested):
Let's not publish anything if the variable is not explicitly set. This would minimize encountering this issue when adding more publishing from other steps.

La0:
I disabled the publication by default, and moved the logic to a CLI args, as all other parameters.

eu9ene (approved):

PR #556, opened 2024-05-02
Fix the opus test

gregtatum:
I thought I had fixed this on the #551, but apparently I hadn't committed it to the right place and pushed it. It looks like the tests didn't run in CI so still had a green check mark. Anyway, this should fix main, which currently is broken.


gregtatum:
Ok, my local stuff was messed up, nevermind.

Issue #503, opened 2024-03-27
Add COMET to the evaluation steps


gregtatum:
We already use COMET in models repo, but don't have it in the training steps. This makes it hard to cross-compare with our release criteria while training models.

PR #557, opened 2024-05-02
Add a docker marker

### LINKED ISSUE #506: Add a `binaries` marker to pytests
We have tests that require external binaries like Marian and others. These are all available in our local Docker images. However this image is quite slow on non-x86 systems. In this case it would be nice to mark tests that are requiring external binaries.

We can do this through pytest markers:

https://docs.pytest.org/en/7.1.x/example/markers.html

`@pytest.mark.binaries`

Then you can run them using `-m "not binaries"`.

gregtatum:
Fixes #506.

I opted for the name `docker_amd64`, as I'm experimenting with multiple local docker images right now while working on #503

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -86,3 +86,8 @@ build-backend = "poetry.core.masonry.api"
 
 [tool.pytest.ini_options]
 testpaths = ["tests"]
+markers = [
+  # Run tests outside of docker:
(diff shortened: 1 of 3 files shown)


eu9ene (approved):

gregtatum:
I'm merging even though tests are failing due to #564.

Issue #538, opened 2024-04-26
docker tasks on generic worker sometimes hit issues with caches


bhearsum:
For example: https://firefox-ci-tc.services.mozilla.com/tasks/IvbeCQBuRuKIOaeOIGEfHg

We had an initial run of this task which got spot killed. Subsequent runs failed with:
```
error: cache /builds/worker/checkouts is not empty and is missing a .cacherequires file; the cache names for this task are likely mis-configured or TASKCLUSTER_CACHES is not set properly
```

bhearsum:
@matt-boris and I dug into this a bunch today. Here's what we found:
- The task linked above got into a strange loop where it was quickly spot terminated, then the next run failed with the error above, and then repeated that over and over.
- There was [_another_ task in the same graph](https://firefox-ci-tc.services.mozilla.com/tasks/Z9yD0ywmSdqYM7ViOtW0vg/runs/0) that had many exceptions - but _all_ of those were spot terminations. The first rerun after the first spot termination had no issue with caches - they end up in the "cache is empty" code path in `run-task` just like the first run.
- The most notable differences between the docker-worker machines and the new workers that are hitting this issue are: the new ones are generic-worker, and that they handle spot terminations correctly. This _ought_ to mean that the cache is preserved between runs.
- The error message we're seeing [comes from run-task](https://github.com/taskcluster/taskgraph/blob/f05edebfd9c30f02d84db1725d1c8d0949c1f21b/src/taskgraph/run-task/run-task#L511-L519), and happens when we have a non-empty cache with no `.cacherequires` file on it.

There are two very confusing parts of this still:
1) Why does one task that's been spot killed end up with this error, while another one runs fine?
2) How exactly could get into this state. The runs that get spot killed get far enough that they [have already written a .cacherequires file](https://github.com/taskcluster/taskgraph/blob/f05edebfd9c30f02d84db1725d1c8d0949c1f21b/src/taskgraph/run-task/run-task#L415-L416), which suggests that we should end up in [the code path in run-task that finds a .cacherequires file](https://github.com/taskcluster/taskgraph/blob/f05edebfd9c30f02d84db1725d1c8d0949c1f21b/src/taskgraph/run-task/run-task#L431)

I think the next step here is to add a bunch of additional logging to `run-task` to help inspect the state of the cache when we hit a failure like this. It _should_be fairly easy to achieve this by simulating host maintance events on workers.

bhearsum:
[...]

PR #561, opened 2024-05-03
Re-enable generic-worker for CPU tasks

bhearsum:
Analysis done in https://github.com/mozilla/firefox-translations-training/issues/538#issuecomment-2079685137 shows that the problems are intermittent, and largely related to spot terminations. We're seeing the latter in the existing workers anyways, so unless we find more serious issues with generic-worker for CPU tasks, we may as well go ahead with this.

### DIFF
--- a/taskcluster/config.yml
+++ b/taskcluster/config.yml
@@ -65,28 +65,28 @@ workers:
             provisioner: '{trust-domain}-{level}'
             implementation: docker-worker
             os: linux
-            worker-type: 'b-linux-large-gcp'
+            worker-type: 'b-linux-large-gcp-d2g'
         # Use for tasks that don't require GPUs, but need lots of disk space
         # eg: dataset cleaning & merging
         b-cpu-largedisk:
             provisioner: '{trust-domain}-{level}'
             implementation: docker-worker
             os: linux
-            worker-type: 'b-linux-large-gcp-300gb'
+            worker-type: 'b-linux-large-gcp-d2g-300gb'
         # Use for tasks that don't require GPUs, but need immense amounts of disk space
         # eg: alignments
         b-cpu-xlargedisk:
             provisioner: '{trust-domain}-{level}'
             implementation: docker-worker
             os: linux
-            worker-type: 'b-linux-large-gcp-1tb'
+            worker-type: 'b-linux-large-gcp-d2g-1tb'
         # Use for tasks that don't require GPUs, but need immense amounts of disk space
         # and higher reliability
         b-cpu-xlargedisk-standard:
             provisioner: '{trust-domain}-{level}'
             implementation: docker-worker
             os: linux
-            worker-type: 'b-linux-large-gcp-1tb-standard'
+            worker-type: 'b-linux-large-gcp-d2g-1tb-standard'
         # Use for quick tasks that need a GPU, eg: evaluate
         b-gpu:
             provisioner: '{trust-domain}-{level}'


bhearsum:
This was previously reviewed by RelEng; doesn't need review by them again IMO. @eu9ene and/or @gregtatum it's up to you if/when we re-land this as far as I'm concerned.

eu9ene:
The two issues I reported to you were OOMs and the third one was due to preemptions I think because the task has finished successfully after a couple of restarts. 

The risk of landing this PR is running into "cache something" issue again and we want to stabilize the pipeline now. It would be definitely nice to enable monitoring for CPU machines to look at resource utilization. Should we try to reproduce the issues with caching again before landing this?

bhearsum:
> The two issues I reported to you were OOMs and the third one was due to preemptions I think because the task has finished successfully after a couple of restarts.
> 
> The risk of landing this PR is running into "cache something" issue again and we want to stabilize the pipeline now. It would be definitely nice to enable monitoring for CPU machines to look at resource utilization. Should we try to reproduce the issues with caching again before landing this?

To be honest, I don't think I'll have time in the next week weeks to debug that, as both bringing the Snakepit machines online and getting spot instances turned back on are higher priority.

However, I don't think what we saw in the run linked from #538 is representative of what we'll see. It's _definitely_ possible that we'll get spot killed and get those strange cache failures, but those _do_ fix themselves in the next run by clearing out the cache. What happened in that run was that we had:
- spot kill
- cache issue
- spot kill
- cache issue
- etc.

So - we do lose a little bit of time when we have the cache issue, but it's not quite as serious as it first seemed IMO.

In any case - I'm happy to live with this status quo for now, or try to move forward this again - totally your / @gregtatum's call.

eu9ene:
Let's hold for now and wait until we're in a more stable state with the training recipe and infra. Then we can give it another try. We can try training something from this branch and check that the cache issues don't impact training. It's important to merge this soon though because we want to monitor CPU resources to tune machine configs.

bhearsum:
@eu9ene - Here's a Task that demonstrates successfully recovering from the cache issue: https://firefox-ci-tc.services.mozilla.com/tasks/WM8ahtXGQSCuvlljOU3O0Q/runs/0

Obviously this is very annoying, and if you don't want to land until we get to the bottom of it, that's fine, but hopefully this gives some confidence that it won't burn a tremendous amount of time. (We don't even download upstream artifacts before the cache fails the job, so these failures typically only cost us a few minutes.)

eu9ene:
> @eu9ene - Here's a Task that demonstrates successfully recovering from the cache issue: https://firefox-ci-tc.services.mozilla.com/tasks/WM8ahtXGQSCuvlljOU3O0Q/runs/0
> 
> Obviously this is very annoying, and if you don't want to land until we get to the bottom of it, that's fine, but hopefully this gives some confidence that it won't burn a tremendous amount of time. (We don't even download upstream artifacts before the cache fails the job, so these failures typically only cost us a few minutes.)

@bhearsum ok, if the risk is minimal and we just lose some retry attempts with this (should we increase them then?), then I guess being able to monitor CPU machine outweighs the extra restarts on pre-emption.

eu9ene (approved):

bhearsum:
> > @eu9ene - Here's a Task that demonstrates successfully recovering from the cache issue: https://firefox-ci-tc.services.mozilla.com/tasks/WM8ahtXGQSCuvlljOU3O0Q/runs/0
> > Obviously this is very annoying, and if you don't want to land until we get to the bottom of it, that's fine, but hopefully this gives some confidence that it won't burn a tremendous amount of time. (We don't even download upstream artifacts before the cache fails the job, so these failures typically only cost us a few minutes.)
> 
> @bhearsum ok, if the risk is minimal and we just lose some retry attempts with this (should we increase them then?), then I guess being able to monitor CPU machine outweighs the extra restarts on pre-emption.

We can always back out again if it's problematic!

PR #568, opened 2024-05-06
Use a virtual environment per requirements.txt file in run_task

gregtatum:
Edit: See [my comment](https://github.com/mozilla/firefox-translations-training/pull/568#issuecomment-2105297536) below about my removal of the docker changes.

---

The PR does essentially 2 things.

1. It splits the docker image into 2. The first is the base testing layer with all of the testing dependencies. The second is the training layer, which requires the amd64 build image.

2. The second is that it adds a new virtual environment layer to the run_task abstraction. This fixes the issues around dependency management conflicts for python dependencies coming from the ML python community. For instance, eflomal will not install on my machine, but is required for tasks. The virtual environments are unique per requirements file, and are re-used between test runs. They are keyed off of the docker execution environment, requirements file, and hash of the key files they depend on to be created.

The commits should be well-ordered and split in a way that makes the overall code changes easier to read. A lot of the code diff size is related to poetry.lock changes due to dependency management.

This should unblock me so that I can use COMET in run_task tests for #503.


utils/run_model.py
```diff
@@ -165,11 +165,14 @@ def translate_over_websocket(port: int):
...
 def main() -> None:
-    if not os.environ.get("IS_DOCKER"):
+    dockerfile = os.environ.get("DOCKERFILE")
+    if not dockerfile:
         # Re-run the command in docker if it wasn't started.
         args = sys.argv[1:]
-        subprocess.check_call(["task", "docker-run", "--", "task", "run-model", "--", *args])
+        subprocess.check_call(["task", "docker-run-amd64", "--", "task", "run-model", "--", *args])
         return
+    if dockerfile != "local-train":
```
bhearsum, line 174: AFAICT this is unreachable because if `dockerfile == "local-train"` we're in the block above that returns, or `dockerfile is None` (in which case, does the rest of this function work correctly?).

Was this meant to go above `if not Dockerfile`?

bhearsum (commented):
I haven't dove into most of what this is touching much until now, so just to make sure I understand: the files in `docker` are used for local dev (and only local dev?)

If so, I think this makes a lot of sense. I'm very much on board with dockerizing tests as much as possible to increase reproducability.

This patch largely seems correct, but I have a couple of things inline, which may very well just be me misunderstanding something.

utils/tasks/docker-run-amd64.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+set -x
+
+# Run the local docker image, see the Taskfile.yml for usage.
+
+if [ $(uname -m) == 'arm64' ]; then
```
eu9ene, line 6: FYI I tried to run this from main and it did not detect the platform properly here:
```
task test-docker
task: [test-docker] task docker-run -- task test
task: [docker-build] ./utils/tasks/docker-build.sh
+ source utils/tasks/docker-setup.sh
++ set -x
+++ uname -m
++ '[' x86_64 == arm64 ']'
```
I remember it was running under Rosetta when launched from pip environment. Not sure if it got fixed here.

Taskfile.yml
```diff
@@ -101,35 +106,49 @@ tasks:
         task test -- tests/test_alignments.py
     deps: [poetry-install-tests]
     cmds:
+      - poetry install --only tests --only utils --no-root
       - PYTHONPATH=$(pwd) poetry run pytest -vv {{.CLI_ARGS}}
 
+  test-native:
```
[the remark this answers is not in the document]
gregtatum: > Rename this to something like test-non-amd64

I don't think this is accurate, as the base docker image is multi-platform. It will use the "native" architecture for your OS, whether it is ARM, AMD64. So here native would be more accurate, while using `test-non-amd64` could be running on native amd64 system.

How about I make it more specific and use "multiplatform", as that's really the distinction here.

gregtatum:
> I haven't dove into most of what this is touching much until now, so just to make sure I understand: the files in docker are used for local dev (and only local dev?)

Yes, those files are only local dev.

Taskfile.yml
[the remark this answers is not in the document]
bhearsum: Ah, apologies - I misunderstood what the naming here was for. I'm happy with the current version, or the new suggestion.

bhearsum (approved):

Taskfile.yml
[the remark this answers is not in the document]
gregtatum: I actually like the `multiplatform` name better, as it's more accurate as to what's going on.

gregtatum:
I pulled out the docker work. It turns out my Docker Desktop was stuck on a version that had an update bug and I was stuck on an old version. That old version did not support Rosetta, and running the amd64 image on this old version was horrifically slow. After the update, things work pretty quickly. Most of my motivations for the docker work were to provide a multiplatform image that was fast. At this point, I don't think that is necessary, so I'm pulling that out and making this just about run_task.

I will probably follow-up with a few tweaks to our local docker image, but now things are working much better for me.

tests/fixtures/__init__.py
```diff
@@ -301,16 +342,37 @@ def find_pipeline_script(commands: Union[list[str], list[list[str]]]) -> str:
...
-def get_task_command_and_env(task_name: str, script=None) -> tuple[str, dict[str, str]]:
+def find_requirements(commands: Commands) -> Optional[str]:
+    command = get_command(commands)
+
+    # Match the following:
+    # pip install -r $VCS_PATH/pipeline/eval/requirements/eval.txt && ...
+    #                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+    match = re.search(
+        r"""
+        pip3?\ install\ -r\ \$VCS_PATH\/  # Find the pip install.
```
eu9ene, line 353: I suggest we double-check that our requirements are installed with this exact pattern everywhere. It shouldn't be a problem if we migrate to a separate field in the kind in the future.

eu9ene (approved):
I think it's a great way to deal with the complexity of handling multiple environments in the tests. One thing I would suggest is to separate all the tests that run as a taskcluster task into a separate directory because they diverge more and more from regular unit tests in how they are launched.

tests/fixtures/__init__.py
gregtatum: I did another manually search for .txt in all the kinds, and they all seem to match.

gregtatum:
I filed #591 as a follow-up.

Issue #584, opened 2024-05-09
Config setting wandb-publication doesn't work


eu9ene:
It seems the issue is a bool value in YAML.

```
wandb-publication: True
```
becomes:
```
"WANDB_PUBLICATION": "True"
```
in task graph, but the script expects  a string "true".

A workaround is to use a string "true" but the type in YAML schema is bool.

We should also add an example to the reference production config: `taskcluster/configs/config.prod.yml`

PR #585, opened 2024-05-09
Add W&B publication setting

eu9ene:
I'm updating the config with the setting that should work but there's also an issue #584 that we ran into with @bhearsum 

[skip ci]


bhearsum (changes requested):
Can we get this added to https://github.com/mozilla/firefox-translations-training/blob/main/taskcluster/translations_taskgraph/actions/train.py at the same time, please? That way it will show up when we kick off training runs.

eu9ene:
I tested the train action [here](https://firefox-ci-tc.services.mozilla.com/tasks/OdXGyFXVRc2b4QjJ8DXTjQ) and it looks working.

bhearsum (approved):

PR #587, opened 2024-05-10
Add COMET to the evaluation

### LINKED ISSUE #503: Add COMET to the evaluation steps

gregtatum:
Add COMET to the evaluation task. I would like to follow-up with pre-downloading the model using some kind of fetch task rather than downloading it from the source every time. Downloading it every time is probably less reliable, but it at least unblocks things.

Resolves #503


pipeline/eval/requirements/eval.txt
```diff
@@ -1,221 +1,157 @@
...
+# This file is autogenerated by pip-compile with Python 3.10
 # by the following command:
 #
-#    pip-compile --generate-hashes eval-requirements.in
+#    pip-compile pipeline/eval/requirements/eval.in
 #
-colorama==0.4.6 \
-    --hash=sha256:08695f5cb7ed6e0531a20572697297273c47b8cae5a63ffc6d6ed5c201be6e44 \
-    --hash=sha256:4f1d9991f5acc0ca119f9d443620b77f9d6b33703e51011c16baf57afb285fc6
+aiohttp==3.9.5
```
bhearsum, line 7: It's difficult to argue that there's any real security risks to not pinning hashes, but could you please regenerate this with `--generate-hashes` anyways, to be safe? (I suppose the worst thing that could happen is pulling in cryptomining code or some such...)

bhearsum (changes requested):
This looks fine to me with the requirements regenerated. Does it need @eu9ene's review as well?

pipeline/eval/requirements/eval.txt
gregtatum: This wasn't an intentional change on my part. To guard against this in the future, I added a task to update the requirements.

https://github.com/mozilla/firefox-translations-training/pull/587/commits/27d130afd43f4c696eb80fa860517703ffb27a93

Taskfile.yml
```diff
@@ -207,3 +207,30 @@ tasks:
...
+    cmds:
+      # Make sure a file was given to update.
+      - >-
+        if [[ -z "{{.CLI_ARGS}}" ]]; then
+          echo "Provide a path to the .in file";
+          echo "For example:"
+          echo "task update-requirements -- pipeline/eval/requirements/eval.in";
+          exit 1
+        fi
+      # Make sure the command is being run for docker
```
gregtatum, line 225: Python 3.12 breaks on my machine for pip-compile on our repo, this is why I am enforcing docker.

Taskfile.yml
```diff
@@ -207,3 +207,30 @@ tasks:
     cmds:
       - >-
         PYTHONPATH=$(pwd) poetry run python -W ignore utils/run_model.py {{.CLI_ARGS}}
+
```
gregtatum, line 210: This addition is in addition piece of code to address @bhearsum's feedback.

Taskfile.yml
bhearsum: Yup, that makes sense. (We do this in a bunch of other places as well for the same reason.)

Taskfile.yml
```diff
@@ -207,3 +207,30 @@ tasks:
...
+    summary: |
+      Example usage:
+      task update-requirements -- pipeline/eval/requirements/eval.in
+    cmds:
+      # Make sure a file was given to update.
+      - >-
+        if [[ -z "{{.CLI_ARGS}}" ]]; then
+          echo "Provide a path to the .in file";
+          echo "For example:"
+          echo "task update-requirements -- pipeline/eval/requirements/eval.in";
```
eu9ene, line 222: I assume the path should not be hardcoded here

pipeline/eval/eval.py
```diff
@@ -158,11 +164,11 @@ def main(args_list: Optional[list[str]] = None) -> None:
     elif args.model_variant == "gpu":
         if not args.workspace:
             raise Exception("The workspace size was not provided")
-        if not args.gpus:
-            raise Exception("The number of GPUs was not provided")
+        # if not args.gpus:
```
eu9ene, line 167: Uncomment or remove?

pipeline/eval/eval.py
```diff
@@ -225,23 +231,43 @@ def main(args_list: Optional[list[str]] = None) -> None:
...
+    comet_checkpoint = comet.download_model(
+        comet_model_name, saving_directory=os.environ.get("COMET_MODEL_DIR")
+    )
+    comet_model = comet.load_from_checkpoint(comet_checkpoint)
+    comet_data = []
+    for source, target, target_ref in zip(source_lines, target_lines, target_ref_lines):
+        comet_data.append({"src": source, "mt": target, "ref": target_ref})
+    comet_results = comet_model.predict(comet_data, gpus=args.gpus)
+    # Reduce the precision.
+    comet_score = float(round(comet_results.system_score * 1e4)) / 1e4
```
eu9ene, line 268:
```suggestion
    comet_score = round(comet_results.system_score, 4)
```

tests/test_eval.py
```diff
@@ -41,24 +44,30 @@ def get_quantized_marian_args(data_dir: DataDir, model_name: str):
...
+    ("evaluate-student-sacrebleu-wmt09-en-ru",                    "base",      "final.model.npz.best-chrf.npz", comet_forwards),
+    ("evaluate-teacher-ensemble-sacrebleu-sacrebleu_wmt09-en-ru", "base",      "model*/*.npz",                  comet_forwards),
+    ("evaluate-quantized-sacrebleu-wmt09-en-ru",                  "quantized", "model.intgemm.alphas.bin",      comet_forwards)
 ]  # fmt:skip
 
 
 @pytest.mark.parametrize("params", test_data, ids=[d[0] for d in test_data])
 def test_evaluate(params) -> None:
-    (task_name, model_type, model_name) = params
+    (task_name, model_type, model_name, comet) = params
```
eu9ene, line 62: should we separate running comet to another test so that we at least can easily run the fast ones?

pipeline/eval/eval.py
```diff
@@ -225,23 +231,43 @@ def main(args_list: Optional[list[str]] = None) -> None:
...
+
+    # COMET_MODEL_DIR allows tests to place the model in a data directory
+    comet_checkpoint = comet.download_model(
+        comet_model_name, saving_directory=os.environ.get("COMET_MODEL_DIR")
+    )
+    comet_model = comet.load_from_checkpoint(comet_checkpoint)
+    comet_data = []
+    for source, target, target_ref in zip(source_lines, target_lines, target_ref_lines):
+        comet_data.append({"src": source, "mt": target, "ref": target_ref})
+    comet_results = comet_model.predict(comet_data, gpus=args.gpus)
```
eu9ene, line 266: You can also control `batch_size` here. The default one might be fine (16 I think), but ideally we should check how it utilizes GPU memory and bump it if it's underutilized.

eu9ene (commented):
Nice! I see we've been running evals on a GPU machine so we should double-check that COMET actually uses it. It outputs some warnings: https://firefox-ci-tc.services.mozilla.com/tasks/BoXF3JpuQeOfjfp485c-2Q/runs/0/logs/public/logs/live.log. The scoring with comet took 15 minutes and it's just 1000 examples I think.

Taskfile.yml
gregtatum: The path is just an example path.

pipeline/eval/eval.py
gregtatum: Thanks, that's old code. I'll remove it.

pipeline/eval/eval.py
gregtatum: I removed the parameter from it since I figured the defaults would be fine.

> ideally we should check how it utilizes GPU memory

Ah, I haven't used the dashboards yet to do any tuning like this yet.

pipeline/eval/eval.py
gregtatum: That's much nicer!

bhearsum:
re: the test failure, I'm very confused by it. It's looking for `nvidia-cuda-nvrtc-cu12`, which apparently comes in via `torch` (which comes in via `unbabel-comet`), but that package doesn't explicitly (at least obviously...) depend on it. https://github.com/pytorch/pytorch/issues/100974#issuecomment-2016303514 has some discussion of a similar issue that suggests that bringing torch in as a direct dependency may make `pip-compile` do the right thing.

tests/test_eval.py
gregtatum: If we do maybe we should follow-up. I'm a bit annoyed how long this takes to run. I spent a bit of time trying to train my own comet model that was much smaller so that it could potentially run faster.

Splitting up the test would be hard since it's a run_task and not a unit test where it's easy to run just parts of it. I'd have to add a bunch of environment variables that artificially not do work, and then the assertions would get a bit messy to run.

gregtatum: Of course I figured it out after typing this. I split up the test.

Taskfile.yml
```diff
@@ -207,3 +207,30 @@ tasks:
...
+        if [[ -z "$IS_DOCKER" ]]; then
+          task docker-run -- task update-requirements -- {{.CLI_ARGS}} && exit
+        fi
+      # Make sure pip-tools are available in docker.
+      - >-
+        if ! command -v pip-compile &> /dev/null; then
+          pip install pip-tools
+        fi
+      # Finally generate the hashes.
+      - pip-compile --generate-hashes {{.CLI_ARGS}} --allow-unsafe
```
gregtatum, line 236: --allow-unsafe fixed my issue for installing torch, and it's the recommended config.

```
  --allow-unsafe / --no-allow-unsafe
                                  Pin packages considered unsafe: distribute,
                                  pip, setuptools.

                                  WARNING: Future versions of pip-tools will
                                  enable this behavior by default. Use --no-
                                  allow-unsafe to keep the old behavior. It is
                                  recommended to pass the --allow-unsafe now
                                  to adapt to the upcoming change.
```

eu9ene (approved):

bhearsum (approved):

Issue #519, opened 2024-04-09
Publish evals from Taskcluster


eu9ene:

La0:
@eu9ene Could you link here all the Taskcluster kind files you want to track ?

eu9ene:
Everything that starts from "evaluate":
https://github.com/mozilla/firefox-translations-training/blob/main/taskcluster/kinds/evaluate-quantized/kind.yml
https://github.com/mozilla/firefox-translations-training/blob/main/taskcluster/kinds/evaluate-teacher-ensemble/kind.yml
https://github.com/mozilla/firefox-translations-training/blob/main/taskcluster/kinds/evaluate/kind.yml

The goal is to integrate everything we had in the offline publisher.

Also, we should reuse the code from the offline publisher that maps the task label to the training steps to keep the naming consistent.

Issue #574, opened 2024-05-07
Actualize moz-translations when online uploading is feature complete


eu9ene:
We should:
-  reupload the online experiments with added functionality directly from Taskcluster
-  upload the new ones from the [spreadsheet](https://docs.google.com/spreadsheets/d/1iyKKWVYXV68sAKwzU9Y6f6KNRnvQQ7HJ7431TSfr5BU/edit#gid=0).

From this point, we won't need to reupload the old experiments unless there are some bugfixes or functionality to add.

We will also consider `moz-translation` our production namespace. All testing should happen in another namespace.

eu9ene:
Also, we'll need to wait until the big training is finished (ETA - end of August).

eu9ene:
I will prepare a list with group IDs.

Issue #544, opened 2024-04-29
Improve caching of teacher ensembles


eu9ene:
Currently, the name of the teacher training task includes the number of models in the ensemble:
```
"{src_locale}-{trg_locale}-{{this_chunk}}/{{total_chunks}}"
```

Our usual use case is to experiment with one teacher and if we have good results to run the whole pipeline. This means `total_chunks` will be switched from 1 to 2. Ideally, we want not to retrain one of the models and use the cached one instead.

The fix seems to be removing `total_chunks` from the name. Maybe bere there's something else. cc @bhearsum

bhearsum:
That seems sensible to me, yeah. The only reason I put the total chunks in the name was because it's what we do in chunked jobs on other projects.

If you care about having `total_chunks` in the name you could also fiddle with it in https://github.com/mozilla/firefox-translations-training/blob/b7e0a2732c45781626f2f7f7b291edae3e4e97e4/taskcluster/translations_taskgraph/transforms/cached_tasks.py#L98, although that might be a bit more hacky, and you're going to bust existing caches either way.

gregtatum:
Resolved by #606.

Issue #449, opened 2024-02-15
Inconsistent dataset naming publishing Taskcluster group


vrigal:
Follow-up of #406.
 
Label used from taskcluster is inconsistent among evaluations for runs and evaluations that does not have a run (thus displaying data in different graph):
* evaluate-student-flores-aug-typos_devtest-lt-en becomes "Flores-aug-typos_devtest-lt-en"
* evaluate-teacher-ensemble-flores-flores_aug-typos_devtest-lt-en becomes "Flores-flores_aug-typos_devtest-lt-en"

This can be fixed either by retrieving dataset name from the task, or handling the duplicate (e.g. `flores-flores…`).

eu9ene:
Check out [this](https://github.com/mozilla/firefox-translations-training/blob/190358a9231c5b4e057bf465fcbcef783ddbcc34/utils/taskcluster_downloader.py#L23) regex, it might handle some problems with inconsistency.

vrigal:
This regex has been used already to support those cases (from https://github.com/mozilla/firefox-translations-training/pull/406#issuecomment-1946702488) :
https://github.com/mozilla/firefox-translations-training/blob/190358a9231c5b4e057bf465fcbcef783ddbcc34/utils/taskcluster_downloader.py#L23-L39

eu9ene:
Where? I don't see it in the tracking code, only in the utils. If it's being used it might miss some cases then so feel free to improve it.

eu9ene:
Here is an example names from configs with the format we should use: 

`<importer>_<augmentation?>_<dataset>`

```
    - flores_devtest
    - flores_aug-title_devtest
    - flores_aug-upper_devtest
    - flores_aug-typos_devtest
    - flores_aug-noise_devtest
    - flores_aug-inline-noise_devtest
    - sacrebleu_wmt19
    - sacrebleu_aug-title_wmt19
    - sacrebleu_aug-upper_wmt19
    - sacrebleu_aug-typos_wmt19
    - sacrebleu_aug-noise_wmt19
    - sacrebleu_aug-inline-noise_wmt19
    - mtdata_Neulab-tedtalks_test-1-eng-lit
```

For the last one `mtdata` is importer and `Neulab-tedtalks_test-1-eng-lit` is a dataset name.

See data import [docs](https://mozilla.github.io/firefox-translations-training/data.html)  and augmentation [docs](https://mozilla.github.io/firefox-translations-training/opus-trainer.html#data-augmentation).

PR #589, opened 2024-05-13
Generic Taskcluster task naming

### LINKED ISSUE #573: Ensure naming for runs is consistent
I noticed that the teacher training run is called `teacher` in some runs and  `teacher-1` in others. It will make it hard to compare them.

Please check all names for both offline and online Taskcluster runs and ensure they are consistent. We should also try to make them consistent with old GCP experiments but the teacher training recipe was different back then and it's ok that it's called differently (`teacher-all`, `teacher-finetuned`). Except that all other runs should have the same naming.

Examples:
recent fix: https://wandb.ai/teklia/565-lt-en
online: https://wandb.ai/moz-translations/en-ru/runs/rry2yzfq
taskcluster offline: https://wandb.ai/moz-translations/en-ca
old Snakemake experiment: https://wandb.ai/moz-translations/nl-en

vrigal:
Required by #519 (making `build_task_name` supports evaluation tasks labels tags from Taskcluster)

Closes #573


tracking/translations_parser/utils.py
```diff
@@ -50,14 +50,16 @@
...
-    # Else try to parse an evaluation label with importer, dataset and auugmentation
-    match = EVAL_REGEX.match(tag)
+    if match is None:
+        # Else try to parse an evaluation label with importer, dataset and auugmentation
+        match = EVAL_REGEX.match(tag)
     if not match:
         raise ValueError(tag)
     groups = match.groupdict()
-    return groups["model"], groups["importer"], groups["dataset"], groups["aug"]
+    # Always keep the index in the model name (teacher-1/2 → teacher-1)
```
eu9ene, line 59: FYI we're going to rename teacher 1/2 to teacher 1 not to invalidate the caches when we want to train the second part of the teacher ensemble after an experiment with only one teacher. See #544

eu9ene: So maybe it's worth adding support for the names without `/2` right away. @gregtatum We'll need to sync the work on #544 with this.

eu9ene (approved):
Approving but let's discuss if we want to implement support of renaming from #544 right away.

tracking/translations_parser/utils.py
vrigal: Hopefully this is already handled by the parser from #449 (covered [here](https://github.com/mozilla/firefox-translations-training/pull/589/files#diff-cc6b34cc2db0bc83dd40f5ec44f20ef4d2d81b9d01c8f919a845b00b78151b2bL22)).

|full name | corresponding run |
|-|-|
| `teacher_flores_devtest` | `teacher-1` |
| `teacher-1_flores_devtest`| `teacher-1`|
| `teacher-2_flores_devtest`| `teacher-1`|
| `teacher-flores_devtest_1/2`| `teacher-1`|
| `teacher-flores_devtest_2/2`| `teacher-2`|

vrigal:
Closing in favor of upstream patch.

eu9ene (commented):
@vrigal Do I understand correctly that now a suffix `-1` is added to all the runs? If yes, we do not want that. We train multiple models only for the teacher. This `-1` makes names longer. To be more explicit, we want:

```
(for older models)
teacher-base-0
teacher-base-1
teacher-parallel-0
teacher-parallel-1
teacher-all-0
teacher-all-1
finetune-teacher-1
finetune-teacher-2
etc.

(for newer Taskcluster models)
teacher-1
teacher-2
backwards
student
student-finetuned
quantized
teacher-ensemble
```

Naming of the teacher models will differ for the older experiments but everything else should be exactly the same across all the experiments and publishers. We might want to store a map of renamings.

vrigal:
@eu9ene I pushed a [commit](https://github.com/mozilla/firefox-translations-training/pull/589/commits/469781d0533a713f9b794006002e81e329c1ec9d) trying to match your example.

Custom renaming could be a follow-up of this I suppose.

eu9ene (approved):
Ok, I see the values in unit tests but to verify this we would need to reupload everything from scratch. We can do that in #574

PR #590, opened 2024-05-13
[wip] add support for automatic uploading of artifacts (fixes #466)

### LINKED ISSUE #466: automatically upload important artifacts to a GCP bucket
Everything we run in Taskcluster expires eventually. Anything we care about retaining must be uploaded elsewhere, and to ensure it happens we should automate it. We have existing tools for doing this pretty easily in Taskcluster.

@eu9ene or @gregtatum - can we get a list of artifacts from each pipeline step that we should be uploading? Models coming out of training tasks are the obvious thing - but I imagine there might be other things too.

bhearsum:


taskcluster/kinds/beetmover/kind.yml
```diff
@@ -0,0 +1,135 @@
...
+                by-tasks-for:
+                    github-pull-request: dep
+                    pr-action: dep
+                    default: "{training_config_bucket}"
+            # upstream artifacts defines which artifacts from upstream tasks we want to upload
+            upstream-artifacts:
+                by-upstream-kind:
+                    # Tasks from kinds matching these patterns will have all artifacts
+                    # uploaded; other steps will only have logs uploaded.
+                    (distillation-student.*|evaluate.*|export|train-teacher.*):
```
evgenyrp: + train backwards

I would also include artifacts for corpus continuation here right away, but we should double-check with @gregtatum if he wants to do it separately.

The steps would be:
- corpus-align*
- distillation-corpus-final-filtering
- Maybe some other corpus producing steps if we want to rerun alignments (dependencies of the corpus-align* steps)

taskcluster/translations_taskgraph/transforms/beetmover.py
```diff
@@ -0,0 +1,144 @@
...
+        n = task_label[-3]
+        return f"distillation-parallel-src-translate{n}"
+    elif task_label.startswith("distillation-parallel-src-translations-score"):
+        return "distillation-parallel-src-translations-score"
+    elif task_label.startswith("distillation-student-model-finetune"):
+        return "student-finetuned"
+    elif task_label.startswith("distillation-student-model-quantize"):
+        return "quantized"
+    elif task_label.startswith("distillation-student-model-train"):
+        return "student"
```
evgenyrp: I agree it's a good idea to make the naming compatible with our existing structure

bhearsum: Are there any other adjustments you'd like to make to this function, or is the current form good?

evgenyrp: I would use the full task label names where possible for simplicity, like `backtranslations-mono-trg-dechunk-translations` instead of the shortened `backtranslations-mono-trg-dechunk`. Also, it's possible to simplify the code by defining groups of names for which we need different logic and processing them as lists.

bhearsum: I just pushed up a change that addresses most, if not all, of this. I'm assuming you want to keep things like the short "quantized" and "student" - let me know if that's not the case. The effective changes to the taskgraph are:
```
--- /home/bhearsum/tmp/2025-06-20/before.yml	2025-06-20 15:50:37.266864990 -0400
+++ /home/bhearsum/tmp/2025-06-20/after.yml	2025-06-20 15:54:13.876980011 -0400
@@ -678,7 +678,7 @@
       - paths:
           '*.log':
             destinations:
-            - task-reference: logs/ru-en/ci_<decision>/backtranslation-mono-trg/
+            - task-reference: logs/ru-en/ci_<decision>/backtranslations-mono-trg-chunk/
         taskId:
           task-reference: <backtranslations-mono-trg-chunk>
       releaseProperties:
@@ -748,7 +748,7 @@
       - paths:
           '*.log':
             destinations:
-            - task-reference: logs/ru-en/ci_<decision>/backtranslations-mono-trg-dechunk/
+            - task-reference: logs/ru-en/ci_<decision>/backtranslations-mono-trg-dechunk-translations/
         taskId:
           task-reference: <backtranslations-mono-trg-dechunk-translations>
       releaseProperties:
@@ -3556,7 +3556,7 @@
       - paths:
           '*.log':
             destinations:
-            - task-reference: logs/ru-en/ci_<decision>/distillation-mono-src-dechunk/
+            - task-reference: logs/ru-en/ci_<decision>/distillation-mono-src-dechunk-translations/
         taskId:
           task-reference: <distillation-mono-src-dechunk-translations>
       releaseProperties:
@@ -3859,7 +3859,7 @@
       - paths:
           '*.log':
             destinations:
-            - task-reference: logs/ru-en/ci_<decision>/distillation-parallel-src-dechunk/
+            - task-reference: logs/ru-en/ci_<decision>/distillation-parallel-src-dechunk-translations/
         taskId:
           task-reference: <distillation-parallel-src-dechunk-translations>
       releaseProperties:
```

taskcluster/kinds/beetmover/kind.yml
bhearsum: @gregtatum - just reflagging you here: do we want non-log artifacts uploaded for any of the steps listed above? (Or any other ones, for that matter.)

taskcluster/configs/config.ci.yml
```diff
@@ -106,5 +106,6 @@ wandb-publication: true
 target-stage: all-pipeline
 taskcluster:
   split-chunks: 2
+  upload-bucket: dep
```
gregtatum: Question: What does "dep" mean here?

Suggestion: Could we use a full name here so that it's a bit more explicit.

taskcluster/kinds/all-pipeline/kind.yml
```diff
@@ -21,6 +21,8 @@ kind-dependencies:
     - evaluate-teacher-ensemble
     - corpus-analyze-parallel
     - corpus-analyze-mono
+    - beetmover
```
gregtatum: > (This is the name we typically give to upload tasks like this - we can, of course, rename it to upload-artifacts or such to be more in-line with the recent renaming that happened.

Yeah, I would really prefer to use something less cryptic, like "upload" to be clearer what this step does.

taskcluster/kinds/beetmover-experiment-config/kind.yml
```diff
@@ -0,0 +1,57 @@
...
+transforms:
+    - taskgraph.transforms.task_context
+    - translations_taskgraph.transforms.beetmover_experiment_config
+    - taskgraph.transforms.cached_tasks
+    - taskgraph.transforms.task
+
+tasks:
+    beetmover:
+        label: beetmover-experiment-config
+        description: upload experiment config
```
gregtatum: Suggestion: In all the other kinds I added more in-depth documentation so that the docs can be auto-generated, and really explain what each task does. Here this is just repeating the task name without explaining what or why it does. It would be nice to have this have a full description written in full sentences. It's been super helpful working on the big training run having more detailed docs everywhere in the Taskcluster UI.

taskcluster/kinds/beetmover-experiment-config/kind.yml
```diff
@@ -0,0 +1,57 @@
...
+                    github-pull-request: dep
+                    pr-action: dep
+                    default: "{training_config_bucket}"
+            data-map:
+                # `data` key is also required, and added by the
+                # `beetmover_experiment_config` transform
+                - content-type: application/json
+                  destinations:
+                      - {
+                          "task-reference": "experiments/{src_locale}-{trg_locale}/{experiment_name}_<decision>/config.yml"
```
gregtatum: suggestion (clarify): `<decision>` is new here, can you add a short bit of documentation on how this key gets filled in. I'm assuming there is some new transform mechanism somewhere.

taskcluster/kinds/beetmover/kind.yml
```diff
@@ -0,0 +1,136 @@
...
+    - distillation-student-model-train
+    - evaluate
+    - evaluate-quantized
+    - evaluate-teacher-ensemble
+    - export
+    - train-teacher-model
+
+tasks:
+    beetmover:
+        label: beetmover-{name}
```
gregtatum: suggestion (clarify): I'm unclear on how `name` gets populated here. Is this some kind of transform that decides on it? Can we give some examples of what types of labels will be produced here or how they are produced?

taskcluster/kinds/beetmover/kind.yml
```diff
@@ -0,0 +1,136 @@
...
+        description: upload artifacts
```
gregtatum: suggestion: Same comment as above with a more full piece of documentation here.

taskcluster/kinds/beetmover/kind.yml
```diff
@@ -0,0 +1,136 @@
...
+        from-deps:
+            group-by: single
+            # This will pull in stage, src_locale, trg_locale, etc. attributes
+            copy-attributes: true
```
gregtatum: Thought: So I guess this is where `{name}` comes from.

taskcluster/kinds/beetmover/kind.yml
```diff
@@ -0,0 +1,136 @@
...
+                by-tasks-for:
+                    github-pull-request: dep
+                    pr-action: dep
+                    default: "{training_config_bucket}"
+            # upstream artifacts defines which artifacts from upstream tasks we want to upload
+            upstream-artifacts:
+                by-upstream-kind:
+                    # Tasks from kinds matching these patterns will have all artifacts
+                    # uploaded; other steps will only have logs uploaded.
+                    (backtranslations-train-backwards.*|distillation-student.*|evaluate.*|export|train-teacher.*):
```
gregtatum: Question: Do the upload tasks get immediately run after each task runs, or at the end of the pipelline?

taskcluster/translations_taskgraph/transforms/beetmover.py
```diff
@@ -0,0 +1,132 @@
...
+    if prefixes := [
+        prefix for prefix in TASKS_WITH_MATCHING_STEP_DIR_PREFIXES if task_label.startswith(prefix)
+    ]:
+        # there should only be only be one match!
+        assert len(prefixes) == 1
+        return prefixes[0]
+    elif any([task_label.startswith(prefix) for prefix in FULL_TASK_LABEL_STEP_DIR_TASKS]):
+        return task_label
+    elif task_label.startswith("backtranslations-mono-trg-translate"):
+        n = task_label[-3]
```
gregtatum: Suggestion: This seems a bit brittle, can you give an example of what you are expecting here? Same as below. Perhaps it's worth adding some kind of assertion that the label has a certain structure.

taskcluster/translations_taskgraph/transforms/beetmover_experiment_config.py
```diff
@@ -0,0 +1,31 @@
...
+import json
+
+from taskgraph.transforms.base import TransformSequence
+from taskgraph.util.schema import resolve_keyed_by
+
+transforms = TransformSequence()
+
+
+@transforms.add
+def add_training_config(config, jobs):
```
gregtatum: Suggestion: This file doesn't have any documentation, it would be nice to document each of these transforms.

taskcluster/translations_taskgraph/transforms/continuation.py
```diff
@@ -198,10 +198,13 @@ def apply_continuation(config: TransformConfig, jobs: Iterable[Job]):
             continue
 
         if corpus_parallel:
-            if stage == "corpus-merge-parallel":
+            if stage == "corpus-merge-parallel" or config.kind == "beetmover":
```
gregtatum: Question: Did my existing tests catch this, or did you just think to add this here? I'd be curious if we need to add beetmover tasks the the continuation tests to verify the tasks aren't being produced.

taskcluster/kinds/beetmover/kind.yml
gregtatum: It seems correct to me.

taskcluster/translations_taskgraph/transforms/beetmover.py
gregtatum: > I agree it's a good idea to make the naming compatible with our existing structure

It would probably be nice to have everything match the new naming structure at some point, but I'm fine checking this in as it is.

gregtatum (approved):
Thanks! This code looks good to me. I left a few comments around adding some more documentation where things were unclear. I'd also  like to add the full taskgraph.json to the uploads as well, but I think that can be a follow-up, and I might be able to do that work myself after this lands.

My only bigger piece of feedback is preferring not to use jargon like `beetmover` if possible, and prefer the simpler word `upload-`.

Thanks for the work on this!

bhearsum:
> I'd also like to add the full taskgraph.json to the uploads as well, but I think that can be a follow-up, and I might be able to do that work myself after this lands.

This should be pretty straightforward. I'm happy to do it, but if you'd rather use it as, eg: an opportunity to get some experience with this, that sounds sensible to me.

taskcluster/configs/config.ci.yml
bhearsum: I'm glad you called it out...this is another RelEng-ism that means "not production" (which nobody else should be expected to know...). I'm happy to use some other name for it (and perhaps for the `release` bucket too)? Maybe "nonprod" and "prod"? I'm happy to use whatever you prefer.

taskcluster/kinds/beetmover-experiment-config/kind.yml
bhearsum: Noted, will do!

taskcluster/kinds/beetmover/kind.yml
bhearsum: Upload tasks are run immediately after each individual step; the only dependency they have is on the task they are uploading files from. (Except for the rare occasion when the step has run, but the upload has not, in which case they additionally depend on the decision task.)

taskcluster/translations_taskgraph/transforms/continuation.py
bhearsum: Your existing tests caught it. I think you're right that we ought to add the beetmover/upload tasks though.

(I also think the current fix here might actually be causing problems in continuation mode -- I'll double check that, and perhaps trouble you to help me do a real world continuation run when it's ready.)

taskcluster/configs/config.ci.yml
gregtatum: `prod` and `dev` works for me, or `production` and `development`.

taskcluster/translations_taskgraph/transforms/beetmover.py
bhearsum: That's up to you and @evgenyrp as far as I'm concerned. It's easy enough to change it (either now or later).

bhearsum:
> > I'd also like to add the full taskgraph.json to the uploads as well, but I think that can be a follow-up, and I might be able to do that work myself after this lands.
> 
> This should be pretty straightforward. I'm happy to do it, but if you'd rather use it as, eg: an opportunity to get some experience with this, that sounds sensible to me.

Just to clarify: I'm guessing you want `task-graph.json` (eg: [this one](https://firefoxci.taskcluster-artifacts.net/Fdqr6u7uTR-n20bGbTKubQ/0/public/task-graph.json)) that contains resolved taskids of all tasks scheduled, and not `full-task-graph.json` (eg: [this](https://firefoxci.taskcluster-artifacts.net/Fdqr6u7uTR-n20bGbTKubQ/0/public/full-task-graph.json)) that contains no taskids, and all tasks (even those that weren't scheduled)?

bhearsum:
I'll move this back to draft while I finish implementing / testing the changes we've discussed.

taskcluster/kinds/beetmover/kind.yml
bhearsum: This is handled by `task-context`. `name` is one of the variables that is [implicitly available for substitution](https://taskcluster-taskgraph.readthedocs.io/en/latest/reference/transforms/task_context.html#implicit-context). (I'll add a comment to this effect -- it's certainly not obvious.)

taskcluster/kinds/beetmover-experiment-config/kind.yml
bhearsum: This is actually deep taskgraph stuff; I'll add some comments. (Unless you meant something in https://github.com/mozilla/translations/tree/main/docs?)

taskcluster/kinds/beetmover/kind.yml
bhearsum: I'm adding an explanatory comment about this.

taskcluster/translations_taskgraph/transforms/continuation.py
bhearsum: I think I've got this sorted now -- at the very least, the tests pass (with the beetmover tasks added to them), and some sanity checks I did locally show the same tasks being generated as before, except with `beetmover` tasks added for the pipeline tasks that were already running.

I would like to do at least one actual training run with continuation enabled still.

bhearsum:
There's a small amount of finalization that needs to be done here, which is waiting on a deployment to allow for us to use the new `development` and `production` bucket names. Once that's done, I'll be removing the `temp` commit currently here, and running another set of uploads. This blocks merging, but it should be ready for re-review at this point.

https://firefox-ci-tc.services.mozilla.com/tasks/groups/EPJtfrVDRNGx6c-B87FEXA is a task group that uploaded artifacts for everything except the `quantize` step and things downstream of it. [This gist](https://gist.github.com/bhearsum/deb418d219e9ffe7ed8de8a57f399906) contains a list of all files that were uploaded as part of it.

https://firefox-ci-tc.services.mozilla.com/tasks/groups/Wfv1pK31SGqYOXyZNTKV1Q is a task group that uses the following continuation configuration:
```
continuation:
  corpora:
    distillation:
      src: https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/RTlM_ntMQjyPqNCx4qh0SA/runs/0/artifacts/public%2Fbuild%2Fcorpus.ru.zst
      trg: https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/RTlM_ntMQjyPqNCx4qh0SA/runs/0/artifacts/public%2Fbuild%2Fcorpus.en.zst
  vocab:
    src: https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/DdYFEhmNRYyseo0Z9NpInA/runs/0/artifacts/public%2Fbuild%2Fvocab.ru.spm
    trg: https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/DdYFEhmNRYyseo0Z9NpInA/runs/0/artifacts/public%2Fbuild%2Fvocab.en.spm
```

Based on what I see in [the continuation tests](https://github.com/mozilla/translations/blob/16c257ab60c192ff28dc646a4135e90632f500cd/tests/test_continuation.py#L212-L246) this appears to have run the correct pipeline and upload tasks. (Note that the `analyze-mono` upload tasks were run for caching reasons...a [subsequent task group](https://firefox-ci-tc.services.mozilla.com/tasks/groups/c30RuzljQk-W910T5Qye0w) from the same config did not run them.)

taskcluster/kinds/upload-artifacts/kind.yml
```diff
@@ -0,0 +1,171 @@
...
+    - distillation-parallel-src-translate
+    - distillation-parallel-src-translations-score
+    - distillation-student-model-finetune
+    - distillation-student-model-quantize
+    - distillation-student-model-train
+    - evaluate
+    - evaluate-quantized
+    - evaluate-teacher-ensemble
+    - export
+    - train-teacher-model
```
hneiva, line 61: this might be the largest amount of dependencies I've seen in a kind

taskcluster/kinds/upload-artifacts/kind.yml
```diff
@@ -0,0 +1,171 @@
...
+                by-upstream-kind:
+                    (distillation-student.*|evaluate.*|export|train-teacher.*):
+                        # task-reference's are a special notation that allow us to
+                        # indicate to taskgraph that we want <FOO> to be replaced with
+                        # the task id of FOO, which is not known until very late in
+                        # the taskgraph generation process
+                        # https://firefox-source-docs.mozilla.org/taskcluster/taskgraph.html#task-parameterization
+                        "*.log":
+                            - {
+                                "task-reference": "logs/{src_locale}-{trg_locale}/{experiment_name}_<decision>/{step_dir}/"
```
hneiva, line 161: I didn't know `<decision>` was an option in task-references. Is this new?

taskcluster/translations_taskgraph/transforms/upload_artifacts.py
```diff
@@ -0,0 +1,153 @@
...
+    ]:
+        # there should only be only be one match!
+        assert len(prefixes) == 1
+        return prefixes[0]
+    elif any([task_label.startswith(prefix) for prefix in FULL_TASK_LABEL_STEP_DIR_TASKS]):
+        return task_label
+    elif task_label.startswith("backtranslations-mono-trg-translate"):
+        n = get_chunk_number(task_label)
+        return f"backtranslations-mono-trg-translate{n}"
+    elif task_label.startswith("backtranslations-train-backwards-model"):
```
hneiva, line 106: nit: Having 2 map dicts up top might make this a bit more readable
Something like:
```
CHUNKED_TASK_LABELS = {"backtranslations-mono-trg-translate": "backtranslations-mono-trg-translate{n}"...}
TASK_LABELS = {"build-vocab": "vocab"...}
...
```
Then you could just lookup the 2 maps without having all these elifs

taskcluster/translations_taskgraph/util/dependencies.py
```diff
@@ -0,0 +1,33 @@
+def get_upload_artifacts_upstream_dependency(dependencies):
```
hneiva, line 6: Missing file header

taskcluster/translations_taskgraph/worker_types.py
```diff
@@ -0,0 +1,48 @@
...
+def build_upload_artifacts_payload(_, task, task_def):
+    """Converts the simplified `worker` from an upload-artifacts task into a valid,
+    runnable payload, and adds the appropriate scopes and tags."""
+
+    upstream_task = get_upload_artifacts_upstream_dependency(task["dependencies"])[0]
+
+    worker = task["worker"]
+    task_def["tags"]["worker-implementation"] = "scriptworker"
+    task_def["payload"] = {
+        "releaseProperties": {"appName": "translations"},
```
jcristau: shouldn't this take app-name from its input?

taskcluster/kinds/upload-artifacts/kind.yml
bhearsum: Nope, that was introduced all the way back in 2020: https://phabricator.services.mozilla.com/D82595 (it's not a common need though, as most tasks don't fetch anything from decision tasks).

taskcluster/translations_taskgraph/actions/train.py
```diff
@@ -490,6 +490,11 @@ def get_descendants(task_id, queue, res=None):
                         "type": "number",
                         "description": "The number of chunks (parallel jobs) to use in `split` steps",
                     },
+                    "upload-bucket": {
```
jcristau, line 559: Should this be required or is there a sane default if it's not set?

taskcluster/translations_taskgraph/transforms/upload_artifacts.py
```diff
@@ -0,0 +1,153 @@
...
+def substitute_step_dir(_, jobs):
+    for job in jobs:
+        upstream_task = get_upload_artifacts_upstream_dependency(job["dependencies"])[1]
+        step_dir = task_to_step_dir(upstream_task)
+        substitute(job["worker"]["artifact-map"], step_dir=step_dir)
+        yield job
+
+
+# tasks for where the full task label is needed to avoid overwriting logs
+# from different, eg: datasets
```
jcristau, line 53: How is that list built, and is there a risk it'll go out of sync with reality, causing overwrites and/or failures?

taskcluster/translations_taskgraph/transforms/upload_artifacts.py
bhearsum: That's a nice little improvement; will do.

taskcluster/translations_taskgraph/transforms/upload_artifacts.py
```diff
@@ -0,0 +1,153 @@
...
+    "distillation-corpus-final-filtering",
+    "distillation-mono-src-chunk",
+    "distillation-mono-src-dechunk-translations",
+    "distillation-parallel-src-chunk",
+    "distillation-parallel-src-dechunk-translations",
+    "distillation-parallel-src-translations-score",
+}
+
+
+def task_to_step_dir(task_label: str) -> str:
```
jcristau, line 89: ick.  I don't know how to make it better but this function is not pretty :)

taskcluster/translations_taskgraph/worker_types.py
bhearsum: Indeed, it's there, we may as well. (I think an earlier iteration of this didn't have it the kinds, but it clearly does now.)

taskcluster/translations_taskgraph/actions/train.py
bhearsum: Hm, I thought this was handled by the `default` being set a few lines up for the `taskcluster` object...but that probably doesn't work if that's present but doesn't contain this.

Yes - we should set a default here as we do for many other things.

taskcluster/translations_taskgraph/transforms/upload_artifacts.py
bhearsum: It was built by inspection, so yes, this is a risk. We do have [beetmover-level protections for overwrites](https://github.com/mozilla-releng/scriptworker-scripts/pull/1230), but obviously this could still cause errors.

This is probably a job for a Verification; I'll look into it. (I don't think there's any way we can verify this until the entire graph is built...)

tests/test_continuation.py
```diff
@@ -329,6 +367,16 @@ class TestParams:
...
+            "upload-artifacts-distillation-corpus-final-filtering-ru-en",
+            "upload-artifacts-corpus-merge-parallel-ru-en",
+            "upload-artifacts-corpus-merge-mono-src-ru",
+            "upload-artifacts-corpus-merge-mono-trg-en",
+            "upload-artifacts-corpus-align-backtranslations-ru-en",
+            "upload-artifacts-corpus-align-parallel-ru-en",
+            "upload-artifacts-corpus-align-distillation-ru-en",
+            "upload-artifacts-train-backwards-ru-en",
+            "upload-artifacts-train-teacher-model-ru-en-1",
+            "upload-artifacts-build-vocab-ru-en",
```
gregtatum, line 379: Thanks! This should be sufficient to keep the continuation code working with uploads.

Issue #408, opened 2024-01-29
Tune workspace dashboard to enable comparison across models and experiments


eu9ene:
<img width="2545" alt="Screenshot 2024-01-29 at 9 43 07 AM" src="https://github.com/mozilla/firefox-translations-training/assets/2486505/549e3b27-eb8f-4446-b26e-220a8cff234d">

We have a great bar chart to compare across the models for one experiment (runs group).  We should figure out how to tune this dashboard or rename the steps so that we can also compare metrics as well as charts across the experiments. The expectation was that this tool will replace Tensorboard which does a good job visualizing lines from different runs on one dashboard.

<img width="2297" alt="Screenshot 2024-01-29 at 11 19 11 AM" src="https://github.com/mozilla/firefox-translations-training/assets/2486505/d445fada-e1e4-446e-979c-2cd3761ba824">

eu9ene:
If W&B doesn't allow this, the workaround would be to rename runs with a suffix with experiment name so that they are unique. For example `student-finetuned_opustrainer`.

eu9ene:
We discussed that W&B uses display names to show charts and since those names are not unique it doesn't show the runs properly on the charts. We discussed the following potential solutions:
1. We should figure out if there is a workaround on the W&B side, like using an internal ID that also includes a group name. 
2. Add a short suffix to each run name composed of a hash of the group name. We don't have a lot of groups per language, so the chance of a collision is pretty low and we can shorten the suffix. For example: `teacher-1_s7g3j`.
3. Just use a counter for the groups per project and add a suffix with the current value. For example `teacher-1_13`. We would need to load the project groups to calculate the value each time we need to publish something. There might be race conditions possible in this case when we run multiple tasks at the same time, for example, if we run many evaluations all at once as the only tasks in the group. So the hashing approach seems to be more robust.

vrigal:
If we use a unique name, it could also be used as the ID for the run (making #610 trivial, which simplifies all the "resume" code).

PR #595, opened 2024-05-14
Override W&B data on a resumed training

### LINKED ISSUE #594: weights and biases reporting doesn't work correctly when resuming training after a spot termination
Over in #580 we've been working on getting spot instances re-enabled. That part works correctly now, and training is continued from the last checkpoint, but I noticed in our latest run that reporting is turned off when resuming.

In [run #1](https://firefox-ci-tc.services.mozilla.com/tasks/U9n3MagKRfWfZgTf4Qu6fw/runs/1/logs/public/logs/live.log) we had:
```
[task 2024-05-13T23:07:22.900Z] wandb: Currently logged in as: moz-translations-wandb-bot (moz-translations). Use `wandb login --relogin` to force relogin
[task 2024-05-13T23:07:23.803Z] wandb: wandb version 0.17.0 is available!  To upgrade, please run:
[task 2024-05-13T23:07:23.803Z] wandb:  $ pip install wandb --upgrade
[task 2024-05-13T23:07:23.803Z] wandb: Tracking run with wandb version 0.16.1
[task 2024-05-13T23:07:23.803Z] wandb: Run data is saved locally in /home/ubuntu/tasks/task_171564135626786/checkouts/vcs/pipeline/train/wandb/run-20240513_230722-eeqsrpg6
[task 2024-05-13T23:07:23.803Z] wandb: Run `wandb offline` to turn off syncing.
[task 2024-05-13T23:07:23.805Z] wandb: Syncing run backwards
[task 2024-05-13T23:07:23.805Z] wandb: ⭐️ View project at https://wandb.ai/moz-translations/en-ru
[task 2024-05-13T23:07:23.805Z] wandb: 🚀 View run at https://wandb.ai/moz-translations/en-ru/runs/eeqsrpg6
```

But in [run #2](https://firefox-ci-tc.services.mozilla.com/tasks/U9n3MagKRfWfZgTf4Qu6fw/runs/2/logs/public/logs/live.log) we got:
```
[task 2024-05-14T00:51:23.329Z] [tracking WARNING] This run already exists on W&B: [<Run moz-translations/en-ru/eeqsrpg6 (running)>]. No data will be published.
```

Note that when resuming we usually will end up redoing some work that happens after the last checkpoint, but before the termination happened -- I'm not sure the best way to handle this in W&B - I imagine others will have a better idea.

vrigal:
Closes #594

### DIFF
--- a/tracking/translations_parser/publishers.py
+++ b/tracking/translations_parser/publishers.py
@@ -1,5 +1,6 @@
 import csv
 import logging
+import os
 from abc import ABC
 from collections import defaultdict
 from pathlib import Path
@@ -99,8 +100,9 @@ class WandB(Publisher):
         config.update(self.extra_kwargs.pop("config", {}))
 
         try:
-            # Check if a W&B run already exists with this name
             project = next(filter(lambda p: p.name == self.project, wandb.Api().projects()), None)
+            # Check if a W&B run already exists with this name
+            existing_runs = []
             if project and (name := self.extra_kwargs.get("name")):
                 existing_runs = list(
                     wandb.Api().runs(
@@ -108,7 +110,15 @@ class WandB(Publisher):
                         filters={"display_name": name, "group": self.extra_kwargs.get("group")},
                     )
                 )
-                if len(existing_runs) > 0:
+            if len(existing_runs) > 0:
+                if os.environ.get("RUN_ID", 0) > 0:
+                    logger.info(
+                        "Training has been resumed from an earlier run, data will be overwritten on W&B."
+                    )
+                    for run in existing_runs:
+                        logger.warning(f"Deleting existing run {run.display_name}.")
+                        run.delete()
+                else:
                     logger.warning(
                         f"This run already exists on W&B: {existing_runs}. No data will be published."
                     )
--- a/tracking/translations_parser/publishers.py
+++ b/tracking/translations_parser/publishers.py
@@ -110,26 +110,39 @@ class WandB(Publisher):
                         filters={"display_name": name, "group": self.extra_kwargs.get("group")},
                     )
                 )
-            if len(existing_runs) > 0:
-                if os.environ.get("RUN_ID", 0) > 0:
-                    logger.info(
-                        "Training has been resumed from an earlier run, data will be overwritten on W&B."
-                    )
-                    for run in existing_runs:
-                        logger.warning(f"Deleting existing run {run.display_name}.")
-                        run.delete()
-                else:
+            if len(existing_runs) == 0:
+                # Start a new W&B run
+                self.wandb = wandb.init(
+                    project=self.project,
+                    config=config,
+                    **self.extra_kwargs,
+                )
+                return
+            elif len(existing_runs) == 1:
+                run = existing_runs[0]
+                # Avoid overriding an existing run on a first training, this should not happen
+                if int(os.environ.get("RUN_ID", 0)) < 1:
                     logger.warning(
-                        f"This run already exists on W&B: {existing_runs}. No data will be published."
+                        f"A W&B run already exists with name '{name}': {run}. No data will be published."
                     )
                     return
-
-            # Start a W&B run
-            self.wandb = wandb.init(
-                project=self.project,
-                config=config,
-                **self.extra_kwargs,
-            )
+                # Resume an existing run
+                logger.info(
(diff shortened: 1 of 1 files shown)


tracking/translations_parser/publishers.py
```diff
@@ -99,16 +99,25 @@ def open(self, parser) -> None:
...
                 )
-                if len(existing_runs) > 0:
+            if len(existing_runs) > 0:
+                if os.environ.get("RUN_ID", 0) > 0:
+                    logger.info(
+                        "Training has been resumed from an earlier run, data will be overwritten on W&B."
+                    )
+                    for run in existing_runs:
+                        logger.warning(f"Deleting existing run {run.display_name}.")
+                        run.delete()
```
eu9ene, line 119: @vrigal it's not supposed to delete the data from the previous run. It should continue tracking like the restart never happened.

eu9ene (changes requested):

tracking/translations_parser/publishers.py
eu9ene: See https://docs.wandb.ai/guides/runs/resuming#resuming-guidance

vrigal:
I applied the changes (to resume an existing run on W&B) and tested here: https://wandb.ai/teklia/test/runs/tv2vtdha

> we can just pass the artifacts dir to the script parse_tc_logs --from-stream -v --wandb-artifacts ${model_dir}.

Its probably worth opening an issue for this. W&B seems to sync (override) logs & artifacts automatically. (https://wandb.ai/teklia/test/groups/test/files/output.log).

> One complication here is that we continue training from the last checkpoint so there will be small overlap with previously written data

W&B should handle this automatically, as it ignores data with a step inferior the last data written.
I think it should simply trigger some warnings from the W&B client.

eu9ene:
@vrigal one thing I don't understand is who sets "RUN_ID" env?

eu9ene (approved):
Ok, this can work but from the code perspective there are some issues:
- w&b publisher should not know anything about Taskcluster environment variables
- why don't we just use resume="allow" so that if the run already exists, it continues automatically? I don't think there is a use case where we have a run with the same name when running things on Taskcluster except the restart of the same task. I guess this logic was implemented in the publisher for publishing offline experiments to prevent republishing the same ones. In this case, the publisher should accept an argument set by a cli that indicates what to do if the run already exists.
 
Other known issues:
- W&B drops overlapped data from the new run instead of overwriting (`wandb: WARNING (User provided step: 15000 is less than current step: 15001. Dropping entry: {'gnorm': 0.7161, '_timestamp': 1715790811.6654584}). `)

I think since it kind of works we can merge it to unblock enabling spot instances but we should address those issues later.

vrigal:
> why don't we just use resume="allow" so that if the run already exists, it continues automatically? I don't think there is a use case where we have a run with the same name when running things on Taskcluster except the restart of the same task. I guess this logic was implemented in the publisher for publishing offline experiments to prevent republishing the same ones. In this case, the publisher should accept an argument set by a cli that indicates what to do if the run already exists.

This is an old issue. The simpler way to handle this would be to use `<group>-<model>` as a unique ID in W&B.
It should be possible to keep  resume="allow" then. It would be compatible with the override option and would probably work in most case (W&B drops overlapped data, as you mentioned), but requires some important changes in the code.

eu9ene:
> > why don't we just use resume="allow" so that if the run already exists, it continues automatically? I don't think there is a use case where we have a run with the same name when running things on Taskcluster except the restart of the same task. I guess this logic was implemented in the publisher for publishing offline experiments to prevent republishing the same ones. In this case, the publisher should accept an argument set by a cli that indicates what to do if the run already exists.
> 
> This is an old issue. The simpler way to handle this would be to use `<group>-<model>` as a unique ID in W&B. It should be possible to keep resume="allow" then. It would be compatible with the override option and would probably work in most case (W&B drops overlapped data, as you mentioned), but requires some important changes in the code.

We can rethink all that in #408, but I would use UID in model names as a last resort because they would clutter the dashboards.

vrigal:
@eu9ene to be clear, run ID (used to identify a run) is different that run name (used to display graphs). For now we do not use an ID, it is automatically set by W&B (e.g. `brmhnekj`) which guarantees unicity. But if there is a unique way to identify a run (e.g. `<group>-<model_name>`) it could help with deletion(override) or resuming a run. It can be a separate issue than #408. I'll write an issue for this.

PR #596, opened 2024-05-14
Add ability to switch to a one-stage teacher training

### LINKED ISSUE #472: Teacher model does not continue training on original corpus
Language pair: en-ru

Screenshot 2024-03-06 at 11 16 41 AM

After training on a mix with back-translated data, it doesn't continue training on the original corpus. This is the same problem I investigated a couple of years ago and I don't think it's caused by anything new. It's likely related to the quality of the datasets and cleaning.

At the same time for lt-en:

Screenshot 2024-03-06 at 11 25 48 AM

eu9ene:
fixes #472 

[skip ci]


taskcluster/scripts/pipeline/train-taskcluster.sh
```diff
@@ -62,6 +63,7 @@ case "$pretrained_model_mode" in
         "$best_model_metric" \
         "$alignments" \
         "$seed" \
+        "$teacher_mode" \
```
gregtatum, line 66: Question: Do you know if we have any tests for training continuation? It would be nice to have coverage here. (no action needed in this PR)

taskcluster/kinds/finetune-student/kind.yml
```diff
@@ -108,6 +108,7 @@ tasks:
                     0
                     None
                     None
+                    None
```
gregtatum, line 111: Thought: I really want to rewrite the training scripts to all use a consistent python training script... but after our big training run.

taskcluster/configs/config.prod.yml
```diff
@@ -52,6 +52,9 @@ experiment:
   # Determine how many teachers to train.
   # See: docs/teacher-ensemble.md
   teacher-ensemble: 2
+  # Switch to "one-stage" training if back-translations are produced by a high quality model or
+  # the model stops too early on the fine-tuning stage
+  teacher-mode: "two-stage"
```
gregtatum, line 57: Praise: This is a nice solution!

pipeline/train/configs/opustrainer/teacher.one-stage.yml
```diff
@@ -0,0 +1,39 @@
...
+# Train on a mix until early stopping
+# (useful for clean back-translated data produced by a strong model)
+train:
+  - original 0.7
+  - backtranslated 0.3
+  - until original inf
+
+# The default values of the modifiers are taken from the paper https://arxiv.org/pdf/2311.14838.pdf
+# Please refer to docs/opus-trainer.md for further details
+modifiers:
```
gregtatum, line 17: Suggestion (optional): The configs have some duplicated information, it would be a bit more maintainable to have 3 files here, one-stage, two-stage, and a shared config with things like the modifiers seed, etc.

gregtatum (approved):
Thanks, this looks like a good solve for the issue.

pipeline/train/configs/opustrainer/teacher.one-stage.yml
eu9ene: I agree but I'll leave this one for the refactoring of the training script. The final config will likely be composed dynamically from multiple configs. I'll file an issue for the refactoring.

taskcluster/scripts/pipeline/train-taskcluster.sh
eu9ene: We don't have any. It's tricky to test. But it has the same entry point as regular training. It might be easier to test with your recent refactoring. I found only a test that checks that taskcluster-train.py propagates arguments.

bhearsum (approved):

PR #558, opened 2024-05-02
Publish evaluation metrics

La0:
Closes #519 

- Setup `evaluate-*` taskcluster kinds with TC proxy, secret access, publication trigger
- Generic W&B bootstrap code from training publication so it's usable from evaluation code

PR #598, opened 2024-05-15
 Publish evaluation metrics

### LINKED ISSUE #519: Publish evals from Taskcluster

vrigal:
Based on #589 (required for valid evaluation task naming).

Closes #519

Based on Bastien's work (#558)

[skip ci]


vrigal:
I tried testing by setting `{"training_config": { "wandb-publication": True }}` in the training configuration (`taskcluster.translations_taskgraph.parameters`) and change parameters to avoid caching, all in a temporary commit (TRASHME).

However no train nor evaluation task is started in Taskcluster (marked unscheduled). @eu9ene do you have right to start this task for example https://firefox-ci-tc.services.mozilla.com/tasks/ce0zK1kgSxC_cwZesDXPLQ ? (if it works it should publish that eval metric to a new group `ci_b_lxggt6TXCnpi1qsjeIUQ` on `moz-translations` workspace project `ru-en`).

eu9ene:
> I tried testing by setting `{"training_config": { "wandb-publication": True }}` in the training configuration (`taskcluster.translations_taskgraph.parameters`) and change parameters to avoid caching, all in a temporary commit (TRASHME).
> 
> However no train nor evaluation task is started in Taskcluster (marked unscheduled). @eu9ene do you have right to start this task for example https://firefox-ci-tc.services.mozilla.com/tasks/ce0zK1kgSxC_cwZesDXPLQ ? (if it works it should publish that eval metric to a new group `ci_b_lxggt6TXCnpi1qsjeIUQ` on `moz-translations` workspace project `ru-en`).

The [train-backwards](https://firefox-ci-tc.services.mozilla.com/tasks/QsxjhxZDTPiMqPzZTHazOg/runs/0/logs/public/logs/live.log) task is failed, that's why it doesn't start. It's something about: `Publication failed: argument --wandb-project: conflicting option string: --wandb-project`

eu9ene:
Depends on #611 to run CI

eu9ene:
We should not forget to check that the newly added COMET metric is being published.

taskcluster/kinds/evaluate-quantized/kind.yml
```diff
@@ -69,9 +85,11 @@ tasks:
                 - bash
                 - -c
                 - >-
-                    pip install -r $VCS_PATH/pipeline/eval/requirements/eval.txt &&
-                    export PYTHONPATH=$PYTHONPATH:$VCS_PATH &&
                     export PATH=$PATH:~/.local/bin &&
+                    export PYTHONPATH=$PYTHONPATH:$VCS_PATH &&
+                    pip install --upgrade pip
```
eu9ene, line 90: &&

taskcluster/kinds/evaluate-teacher-ensemble/kind.yml
```diff
@@ -43,10 +43,14 @@ tasks:
             substitution-fields:
                 - fetches
                 - run.command
+                - worker.env
+                - dataset
```
eu9ene, line 47: there's no dataset field. substitution will work for run.command

eu9ene: You can try running `task preflight-check` locally to validate this.

pipeline/eval/eval.py
```diff
@@ -136,6 +145,16 @@ def main(args_list: Optional[list[str]] = None) -> None:
...
     )
+    parser.add_argument(
+        "--dataset",
+        type=str,
+        help="The name of the dataset (e.g. 'flores-aug-inline-noise_devtest')",
+    )
+
+    # Add Weight & Biases CLI args when module is loaded
+    if WANDB_AVAILABLE:
+        add_wandb_arguments(parser)
```
eu9ene, line 156: Why are you extending the command line args of another script? Keep in mind that W&B publishing will be used in a bunch of other places, for example for dataset analysis. I understand that the idea is to reuse the code but this is way too hacky and unnecessary. All you need for the publisher class is to accept optional parameters and read the environment variables by default. Also, this makes the `cli/taskcluster.py` less readable.

eu9ene (changes requested):
This will likely work after the bug fix but we should refactor the code reuse without adding arguments to another script. Also please double check that the recently added COMET metric is working. We'll need green CI and everything published on W&B to verify that this works. Thanks for pushing on this one, we really need it for the release!

vrigal:
> Also please double check that the recently added COMET metric is working.

This metric was not specified. I will open a new Issue/PR for this.

It is actually ignored by the online publication. Actual support for `.metrics` files will not work (from old experiments and offline Taskcluster uploads). It will raise `AssertionError: file must contain exactly 2 float values`.

vrigal:
Evaluation data has been published to W&B: https://wandb.ai/moz-translations/ru-en/groups/ci_bVp4cnGJSgCeY2pK5aKK0A/workspace

I found two other things:
* Teacher-1 is not correctly parsed by the regex (label `evaluate-teacher-flores-devtest-ru-en-1`), because it detects `ru-en-1` as part of the dataset (instead of the language at the end of the label). In case that extra language part is always present, it will be a 1 character patch.
* I noticed that simple can be displayed as bar charts once the grouping option is disabled (see our [custom workspace](https://wandb.ai/moz-translations/ru-en/groups/ci_bVp4cnGJSgCeY2pK5aKK0A/workspace?nw=nwuserbabadiemoz)). That would be a great option in order to remove Tables/custom charts for evaluation metrics and homogenize all charts.

pipeline/eval/eval.py
eu9ene: @vrigal I don't think this was resolved. The idea is not to use `parser` of `eval.py` but call  `get_wandb_publisher` with the arguments either manually loaded from the environment variables or maybe using another method from `tracking`.

eu9ene:
@vrigal things don't look right here:  https://wandb.ai/moz-translations/ru-en/groups/ci_bVp4cnGJSgCeY2pK5aKK0A/workspace. Evals use different run names than training:
Screenshot 2024-05-21 at 3 49 46 PM

> Teacher-1 is not correctly parsed by the regex (label evaluate-teacher-flores-devtest-ru-en-1), because it detects ru-en-1 as part of the dataset (instead of the language at the end of the label). In case that extra language part is always present, it will be a 1 character patch.

Yes, I see, this is a must-fix before merging. Please also add `"sacrebleu_aug-upper_wmt19"` to https://github.com/mozilla/firefox-translations-training/blob/f5a6af83849f08cce3e7b8b9d06cc2a0456f2c68/taskcluster/translations_taskgraph/parameters.py#L91 so that we can tests that the augmentation part of the dataset is parsed correctly.

vrigal:
Publication should be fine now (flores & sacrebleu): https://wandb.ai/moz-translations/ru-en/groups/ci_Z45R3e22TWCzyqQ5UFdWdQ/workspace

I updated the regex so it supports labels ending with `-1`, merged run names in the label parser (`backward -> backwards` and `finetuned -> finetune`) and dropped the temporary commit used to trigger the complete CI.

eu9ene (approved):

eu9ene:
The issue in CI is #549

PR #611, opened 2024-05-17
Use shorter names for dataset URLs

gregtatum:
```
evaluate-teacher-url-gcp_pytest-dataset_a0017e-en-ru-1/2
evaluate-teacher-url-gcp_pytest-dataset_a0017e-en-ru-2/2
dataset-url-gcp_pytest-dataset_a0017e-en-ru
clean-mono-url-en-gcp_pytest-dataset_en_cdd0d7-mono-src
```

This PR also includes #606 since the CI failed due to this issue.

Resolves #527.

PR #606, opened 2024-05-16
Change the caching strategy for teacher ensembles

gregtatum:
I haven't validated this change yet in a production run.

PR #612, opened 2024-05-17
Ensure all of the task labels can be parsed in the task graph

gregtatum:
This is a break out from #611 and #606. It reworks the task label parsing, renames the mechanism to use "task label" instead of "tag", and includes other fixes and cleanups to the mechanisms. The commits should be logically ordered for the work.

### DIFF
--- a/tests/test_tracking_utils.py
+++ b/tests/test_tracking_utils.py
@@ -1,10 +1,11 @@
 import pytest
+from fixtures import get_full_taskgraph
 
-from tracking.translations_parser.utils import build_task_name, parse_task_label
+from tracking.translations_parser.utils import ParsedTaskLabel, build_task_name, parse_task_label
 
 
 @pytest.mark.parametrize(
-    "example, parsed_values",
+    "task_label, parsed_values",
     [
         (
             "evaluate-teacher-flores-flores_aug-title_devtest-lt-en-1_2",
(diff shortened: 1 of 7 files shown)


tests/test_tracking_utils.py
```diff
@@ -58,10 +59,42 @@
...
-def test_parse_tag(example, parsed_values):
-    assert parse_tag(example) == parsed_values
+def test_parse_task_label(task_label, parsed_values):
+    assert parse_task_label(task_label) == ParsedTaskLabel(*parsed_values)
+
+
+def test_parse_labels_on_full_taskgraph():
+    """Ensure that all the taskgraph task labels parse."""
+    for task in get_full_taskgraph():
+        if not task.startswith("train-") and not task.startswith("evaluate-"):
```
eu9ene, line 88: nice, we'll catch all the issues in the tests now. Why are you skipping train- and evaluate-? Can we just enable this test for everything?

tracking/translations_parser/utils.py
```diff
@@ -12,49 +13,104 @@
...
+    #
+    # Capture any augmentations
+    #   evaluate-teacher-flores-flores_aug-title_devtest-lt-en-1_2
+    #                                  ^^^^^^^^^
     r"(?P<aug>aug-[^_]+)?"
-    r"_?(?P<dataset>[-\w_]*?(-[a-z]{3}-[a-z]{3})?)?"
-    r"-?(?P<lang>[a-z]{2}-[a-z]{2})?"
+    #
+    # Capture the dataset.
+    #   evaluate-quantized-mtdata_aug-mix_Neulab-tedtalks_eng-lit-lt-en
```
eu9ene, line 79: nice docs!

eu9ene (approved):
lgtm, let's unblock other PRs but we should wait for CI to pass. FYI @La0 @vrigal

tests/test_tracking_utils.py
gregtatum: The logic is inverted here, so if it's not train and not evaluate, it's skipping. Maybe I should rewrite it as:

`if not (task.startswith("train-") or task.startswith("evaluate-")):`

PR #620, opened 2024-05-21
Automatically generate training config files with the `task config-generator`

gregtatum:
This PR adds automatic generation of training configs based on the production config.


configs/en-sl-spring-2024.yml
```diff
@@ -0,0 +1,163 @@
...
+  name: spring-2024
+  src: en
+  trg: sl
+  best-model: chrf
+  use-opuscleaner: 'true'
+  opuscleaner-mode: defaults
+  bicleaner:
+    default-threshold: 0.5
+    dataset-thresholds: {}
+  mono-max-sentences-src: 500_000_000
```
eu9ene, line 16: We need to figure out a policy for monolingual data. We have full control of the mix in OpusTrainer now. but if it's say just 30M of original data I don't know if it makes sense to use 200M mono. Let's discuss live.

configs/en-sl-spring-2024.yml
```diff
@@ -0,0 +1,163 @@
...
+  train:
+  - opus_NLLB/v1  #                                       27,406,782 sentences
+  - opus_OpenSubtitles/v2018 #                           19,641,457 sentences
+  - opus_ELRC-EMEA/v1 #                                  13,288,992 sentences
+  - opus_ParaCrawl/v9 #                                   9,516,259 sentences
+  - opus_DGT/v2019 #                                      5,125,455 sentences
+  - opus_CCAligned/v1 #                                   4,366,555 sentences
+  - opus_TildeMODEL/v2018 #                               2,048,216 sentences
+  - opus_MaCoCu/v2 #                                      1,875,518 sentences
+  - opus_MultiMaCoCu/v2 #                                 1,875,517 sentences
```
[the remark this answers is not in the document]
eu9ene: Here are more recommendations from the Edinburgh group on what to skip: https://github.com/mozilla/firefox-translations-training/blob/2df0a3a905a26fed7e6a6e48ccb2156f29282b4a/utils/find-corpus.py#L23. I think OPUS100 was also some multilingual stuff. KDE, Gnome and Ubuntu were considered too technical I think.

configs/bs-en-spring-2024.yml
```diff
@@ -0,0 +1,114 @@
...
+  training-student:
+    early-stopping: '20'
+  training-student-finetuned:
+    early-stopping: '20'
+target-stage: all
+wandb-publication: true
+taskcluster:
+  split-chunks: 20
+  worker-classes:
+    default: gcp-spot
```
eu9ene, line 114: Like we discussed with @bhearsum we might want to switch bicleaner to gcp-standard if a single parallel dataset is too large to prevent restarts on preemptions that will slow down our training significantly.

utils/config_generator.py
```diff
@@ -0,0 +1,383 @@
...
+
+    # Remove numbers anything that is not a letter. This is a little aggressive, but should help
+    # deduplicate more datasets. For example:
+    #   opus: 725-Hallituskausi_2011_2
+    #   mtdata: hallituskausi_2011_2015-1-eng-fin
+    corpus_name = re.sub(r"[^a-z]", "", corpus_name.lower())
+
+    # Ignore train/test/dev splits.
+    #   opus: NeuLab-TedTalks/v1
+    #   mtdata: Neulab-tedtalks_train-1-eng-fin
```
eu9ene, line 176: Not sure if it's relevant here, but if such a dataset with 3 parts is present we usually use it in `train/devtest/test`

utils/config_generator.py
```diff
@@ -0,0 +1,383 @@
...
+
+        # Add augmented datasets to check performance for the specific cases
+        test_datasets.append("flores_aug-mix_devtest")
+        test_datasets.append("flores_aug-title_devtest")
+        test_datasets.append("flores_aug-upper_devtest")
+        test_datasets.append("flores_aug-typos_devtest")
+        test_datasets.append("flores_aug-noise_devtest")
+        test_datasets.append("flores_aug-inline-noise_devtest")
+
+        devtest_datasets.append("flores_aug-mix_dev")
```
eu9ene, line 206: All devtest datasets should be with `aug-mix` augmentation

utils/config_generator.py
```diff
@@ -0,0 +1,383 @@
...
+"""
+Generate a training config for a language pair based on the latest production
+training config, taskcluster/configs/config.prod.yml.
+"""
+
+root_dir = Path(__file__).parent.parent
+prod_config_path = root_dir / "taskcluster/configs/config.prod.yml"
+
+pretrained_student_models = {
+    ("ru", "en"): "https://storage.googleapis.com/releng-translations-dev/models/ru-en/better-teacher/student"
```
eu9ene, line 27: This is a great idea btw! We just need to prepare those models (copy vocab etc)

configs/en-ru-spring-2024.yml
```diff
@@ -0,0 +1,166 @@
...
+  # The training data contains:
+  #   250,048,886 sentences
+  # 
+  # Skipped datasets:
+  #  - opus_CCMatrix/v1 - ignored datasets
+  #  - opus_bible-uedin/v1 - ignored datasets
+  #  - opus_GNOME/v1 - not enough data
+  #  - opus_ELRC-3855-SWPS_University_Soci/v1 - not enough data
+  #  - opus_Ubuntu/v14.10 - not enough data
+  #  - opus_WikiTitles/v3 - not enough data
```
eu9ene, line 59: WikiTitles has an issue with direction for en-ru/ru-en

utils/config_generator.py
eu9ene: @eu9ene should ask @bhearsum where to find recently trained models

bhearsum: If you're talking about the ones I uploaded by hand, these are in https://console.cloud.google.com/storage/browser/moz-fx-translations-data--303e-prod-translations-data;tab=objects?forceOnBucketsSortingFiltering=true&project=moz-fx-translations-data--303e&prefix=&forceOnObjectsSortingFiltering=false.

Anything after I did those uploads only exists in Taskcluster artifacts.

eu9ene: ok, so the most recent ones are in the tasks and slightly less recent are in `https://console.cloud.google.com/storage/browser/moz-fx-translations-data--303e-prod-translations-data/models/en-hu/`. I see that the model like https://console.cloud.google.com/storage/browser/moz-fx-translations-data--303e-prod-translations-data/models/en-hu/baseline_enhu_aY25-4fXTcuJNuMcWXUYtQ/student-finetuned?pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))&project=moz-fx-translations-data--303e needs a vocab copied to the folder to use for training continuation as we started copying it in the pipeline only recently.

configs/bs-en-spring-2024.yml
eu9ene: a large nllb dataset seems to be the case only for en-ru so let's see how it performs first

configs/en-sl-spring-2024.yml
eu9ene: We decided to use the current policy as news-crawl for English is ~200M sentences and it's fine to have some overhead for back-translations

configs/en-sl-spring-2024.yml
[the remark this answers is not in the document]
eu9ene: We decided not to exclude anything from the baseline training. @gregtatum we still should remove the ones with Multi prefix though.

configs/en-ru-spring-2024.yml
gregtatum: I added it to the ignore list.

utils/config_generator.py
bhearsum: Your GCP accounts has credentials to do this, FYI. That's going to the be quickest way to unblock this.

configs/en-sl-spring-2024.yml
eu9ene: @gregtatum I remember now why I wanted to limit back-translations more for smaller original corpus. With the two-stage approach, we train for two epochs of back-translated first and it will take too long with 200M sentences and reach the second stage quite late. Maybe it's fine for prod training, but it wasn't convenient for the experiments

configs/cs-en-spring-2024.yml
```diff
@@ -0,0 +1,236 @@
...
+  - opus_Tanzil/v1 #                                        233,399 sentences
+  - opus_News-Commentary/v16 #                              218,509 sentences
+  - opus_TED2020/v1 #                                       170,611 sentences
+  - opus_wikimedia/v20230407 #                              146,717 sentences
+  - opus_KDE4/v2 #                                          134,071 sentences
+  - opus_ELRC-presscorner_covid/v1 #                        129,652 sentences
+  - opus_NeuLab-TedTalks/v1 #                               111,107 sentences
+  - opus_ECB/v1 #                                            63,716 sentences
+  - opus_bible-uedin/v1 #                                    62,151 sentences
+  - opus_WMT-News/v2019 #                                    44,859 sentences
```
eu9ene, line 119: We should remove https://opus.nlpl.eu/WMT-News This is a dev/test set

eu9ene: Actually, for some languages, it's pretty big so maybe it's not the test set but it's not clear from https://www.statmt.org/wmt19/translation-task.html that there was an extra dataset released that can be used for training. \

eu9ene: sacrebleu wmt19 for en-lt is 1000 lines long and this one is 5,998 so it's likely something else

utils/config_generator.py
```diff
@@ -0,0 +1,392 @@
...
+    experiment["trg"] = target
+    experiment["bicleaner"]["dataset-thresholds"] = {}
+
+    pretrained_model = pretrained_student_models.get((source, target))
+    if pretrained_model:
+        experiment["teacher-mode"] = "two-stage"
+        experiment["pretrained-models"]["train-backwards"]["urls"] = [pretrained_model]
+    else:
+        experiment["pretrained-models"] = {}
+
```
eu9ene, line 75: There should be also this here:
```
        experiment["teacher-mode"] = "one-stage"
```

But I found another use case when we don't want that. If the amount of mono-trg data is too small (for example for en-lt) we still want to use two-stage. We don't want to loop over 5M back-translated sentences.

configs/en-lt-spring-2024.yml
```diff
@@ -0,0 +1,195 @@
...
+  spm-vocab-size: 32000
+  teacher-ensemble: 2
+  teacher-mode: two-stage
+  pretrained-models: {}
+datasets:
+
+  # Skipped test/devtest datasets:
+  #  - wmt19/dev - variant dataset
+  devtest:
+  - flores_aug-mix_dev
```
eu9ene, line 28: I had to add `  - sacrebleu_aug-mix_wmt19/dev` here manually. Let's output everything from `find-cropus` that's being skipped.
Here's my config:
```
  devtest:
  - flores_aug-mix_dev
  - sacrebleu_aug-mix_wmt19/dev
  - mtdata_aug-mix_Neulab-tedtalks_dev-1-eng-lit

  test:
  - flores_devtest
  - flores_aug-mix_devtest
  - flores_aug-title_devtest
  - flores_aug-upper_devtest
  - flores_aug-typos_devtest
  - flores_aug-noise_devtest
  - flores_aug-inline-noise_devtest
  - sacrebleu_wmt19
  - mtdata_Neulab-tedtalks_test-1-eng-lit
```

utils/config_generator.py
gregtatum: As far as I can tell they all do have `aug-mix` specified here.

configs/cs-en-spring-2024.yml
gregtatum: I wonder if we should have a step the removes any devtest or test from the training data, so we make sure we're not lying to ourselves.

utils/config_generator.py
gregtatum: Let's follow-up with rules for this, or manually configure the configs after they are generated.

configs/en-ru-spring-2024.yml
```diff
@@ -0,0 +1,174 @@
...
+  spm-vocab-size: 32000
+  teacher-ensemble: 2
+  teacher-mode: two-stage
+  pretrained-models: {}
+datasets:
+  devtest:
+  - mtdata_Neulab-tedtalks_dev-1-eng-rus
+  - mtdata_UN-un_dev-1-eng-rus
+  - flores_aug-mix_dev
+  - sacrebleu_aug-mix_mtedx/test
```
eu9ene, line 28: this dataset was not found for en-ru... Can it be because we use different version of sacrebleu in find-corpus? https://firefox-ci-tc.services.mozilla.com/tasks/dKusTBI1Rg-UeXIJz_0amQ/runs/0/logs/public/logs/live.log

eu9ene (approved):
As discussed, we'll continue iterating in follow-up PRs including manual modification of the training configs. 

IMPORTANT!!! The generated configs are not ready for production as some issues around dev/test datasets are not addressed and require manual modification.

gregtatum:
I made a meta to track issues.

https://github.com/mozilla/firefox-translations-training/issues/633

PR #624, opened 2024-05-22
Retry tasks on exit code 128 to work around intermittent issues with DNS failures

bhearsum:
This should help make #549 less painful. I suggest we back it out once we get to the bottom of that.


bhearsum:
It doesn't look like we're going to hit any dns issues in the CI tasks, but I'm reasonably confident this will work.

eu9ene (approved):
lgtm, but let's wait until the CI is green

bhearsum:
(Doesn't really need RelEng review....)

PR #629, opened 2024-05-24
Revert CPU generic workers

eu9ene:


bhearsum (approved):
For what it's worth, the failures in https://firefox-ci-tc.services.mozilla.com/tasks/groups/BFgoyXuvRHe0kK4J9d-Lgw are caused by exceeding the max run time. Bumping https://github.com/mozilla/firefox-translations-training/blob/56040c94b9a745a9594491b57a373ba231739cc9/taskcluster/kinds/clean-corpus/kind.yml#L66 seems like it would fix them.

Nevertheless, I understand if you want to be cautious here and revert anyways. I'll call out that landing this will lose memory monitoring for CPU workers in case that matters at the moment.

Approving this as it's 100% your call as far as I'm concerned.

eu9ene:
@bhearsum I clearly see a lot of random weird errors unrelated to max run time when using those new workers and all of those errors were immediately gone after switching to the old ones. I think there's an infrastructure issue, probably a network-related one.

bhearsum:
I filed https://github.com/mozilla/firefox-translations-training/issues/640 as a followup.

PR #637, opened 2024-05-27
Parse stalled validation data

### LINKED ISSUE #534: Support extra metrics from Tensorboard
I noticed that I have some extra metrics generated by marian tensorboard that are missing on W&B dashboards. They are all useful.

Screenshot 2024-04-24 at 9 04 55 AM

The missing ones:
- valid/bleu-detok_stalled
- valid/ce-mean-words_stalled
- valid/chrf_stalled
- train/words_per_second

vrigal:
Closes #534 

Publication example: https://wandb.ai/teklia/534-lt-en

Note:
* I added support for `perplexity_stalled`.
* `train/words_per_second` is already handled but named `rate` on Weight & Biases. It can be renamed easily in case it is confusing.


vrigal:
I added support for Learning rate, as discussed during the meeting. Results updated on https://wandb.ai/teklia/534-lt-en.

tracking/translations_parser/parser.py
```diff
@@ -128,16 +138,17 @@ def parse_validation_log(
...
+        epoch, up = int(results["ep"]), int(results["up"])
         entry = self._validation_entries[(epoch, up)]
-        entry[key] = val
+        # Transform values to match output types
+        entry[key] = ValidationEpoch.__annotations__[key](results["value"])
+        if results["stalled"] is not None:
+            entry[f"{key}_stalled"] = float(results["stalled"])
         # Build a validation epoch from multiple lines
-        expected_keys = set(ValidationEpoch.__annotations__.keys()) - {"epoch", "up", "perplexity"}
+        expected_keys = {"ce_mean_words", "chrf", "bleu_detok"}
```
eu9ene, line 151: Just to double check: those metrics are specified here and can be changed to any metrics supported by Marian `--valid-metrics` [argument](https://marian-nmt.github.io/docs/cmd/marian/):
https://github.com/mozilla/firefox-translations-training/blob/fac7511f0f84af352c90502132a162a52009cee3/pipeline/train/train.sh#L41

Does this code support arbitrary metrics?

vrigal: Parser's code is not generic, as it uses a regex to match lines with validation data.
However I can keep the original code, so updating new metrics would be simpler (i.e. updating the regex and the ValidationEpoch dataclass). I'll add a bit of documentation about this.

eu9ene: Ok, it's fine. I don't think we'll want to change this soon.

eu9ene (approved):

PR #643, opened 2024-05-28
Update the config generator to remove 404s, and improve the Content-Length lookup

### LINKED ISSUE #642: Exclude mtdata datasets from config generation that return a 404
This will help tasks not fail when training.

gregtatum:
Resolves #642.

* Skips mtdata datasets that have a 404
* Improves the Content-Length lookup, by performing a GET fallback.
* Improves the de-duplication of mtdata datasets.

### DIFF
--- a/configs/en-lt-spring-2024.yml
+++ b/configs/en-lt-spring-2024.yml
@@ -2,7 +2,7 @@
 # task config-generator -- en lt --name spring-2024
 #
 # The documentation for this config can be found here:
-# https://github.com/mozilla/firefox-translations-training/blob/fac7511f0f84af352c90502132a162a52009cee3/taskcluster/configs/config.prod.yml
+# https://github.com/mozilla/firefox-translations-training/blob/b5ee6eeced6c113aaa413fa2f693ccfc340e1d2c/taskcluster/configs/config.prod.yml
 experiment:
   name: spring-2024
   src: en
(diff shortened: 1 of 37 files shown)


configs/en-lt-spring-2024.yml
```diff
@@ -125,10 +126,9 @@ datasets:
   - opus_ELRC-2740-vaccination/v1 #                             546 sentences
   - opus_ELRC-vaccination/v1 #                                  546 sentences
   - opus_ELRC_2923/v1 #                                         384 sentences
-  - mtdata_ELRC-lithuanian_legislation_seimas_lithuania-1-eng-lit
-  - mtdata_ELRC-eu_publications_medical_v2-1-eng-lit
-  - mtdata_ELRC-wikipedia_health-1-eng-lit
-  - mtdata_ELRC-nteu_tierb-1-eng-lit
+  - mtdata_ELRC-lithuanian_legislation_seimas_lithuania-1-eng-lit # ~611,715 sentences (69.1 MB)
```
eu9ene, line 129: This one failed in all retries so I had to remove it. Maybe the  resource was temporarily unavailable

eu9ene (approved):
Looks good, but it seems the current checks still don't guarantee that a dataset won't work on downloading.

gregtatum:
> but it seems the current checks still don't guarantee

Hmm... that's annoying. I'm not sure what else we can do there. mtdata does processing on the downloaded data, so it could also be failing there if the data isn't shaped as expected.

PR #645, opened 2024-05-29
Add HPLT mono bulk importer

eu9ene:
This is an util that I used to download and filter monolingual data from HPLT. This is not a production importer that will be a part of the pipeline. It should be more efficient than this script as it's quite memory-intensive.

[skip ci]

### DIFF
--- a/utils/download_hplt.py
+++ b/utils/download_hplt.py
@@ -1,3 +1,8 @@
+"""
+This script was used to download and filter monolingual data from HPLT.
+This is not a production importer that will be a part of the pipeline.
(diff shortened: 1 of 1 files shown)


utils/download_hplt.py
```diff
@@ -0,0 +1,85 @@
+import argparse
```
gregtatum, line 1: Suggestion: can you add a docstring explaining the state of this script, e.g. something like what you wrote in the PR.

gregtatum (approved):

PR #646, opened 2024-05-29
Add HPLT mono data to the configs

eu9ene:
I regenerated the configs in the fast mode for now.

[skip ci]


gregtatum (approved):
Looks good to me!

eu9ene:
Ok, it was too slow to regenerate in slow mode, so I just reverted the commit part that removed stats for the datasets

PR #647, opened 2024-05-29
Add mono nllb to the config generation

gregtatum:
This is faking it somewhat as I'm still generating the datasets and uploading them.

### DIFF
--- a/configs/sk-en-spring-2024.yml
+++ b/configs/sk-en-spring-2024.yml
@@ -1,8 +1,8 @@
 # The initial configuration was generated using:
-# task config-generator -- --name=spring-2024 --fast sk en
+# task config-generator -- sk en --name spring-2024
 #
 # The documentation for this config can be found here:
-# https://github.com/mozilla/firefox-translations-training/blob/4a86accaaea086b57ff730c25c642dd711e11586/taskcluster/configs/config.prod.yml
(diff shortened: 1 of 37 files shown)


configs/sk-en-spring-2024.yml
```diff
@@ -136,8 +136,9 @@ datasets:
   # The monolingual data contains:
   #   ~37,969,053 sentences
   mono-src:
-  - url_https://storage.googleapis.com/releng-translations-dev/data/mono-hplt/08/hplt_filtered_sk_1.txt.zst  # 31,032,575 sentences 
-  - url_https://storage.googleapis.com/releng-translations-dev/data/mono-hplt/08/hplt_filtered_sk_2.txt.zst # 6,936,478 sentences 
+  - url_https://storage.googleapis.com/releng-translations-dev/data/mono-hplt/08/hplt_filtered_sk_1.txt.zst  # 31,032,575 sentences
+  - url_https://storage.googleapis.com/releng-translations-dev/data/mono-hplt/08/hplt_filtered_sk_2.txt.zst # 6,936,478 sentences
+  - url_https://storage.googleapis.com/releng-translations-dev/data/mono-nllb/nllb-mono-sk.txt.gz
```
eu9ene, line 141: I see that the url importer supports only `RemoteZstdLineStreamer`, not gzip

gregtatum: Nice catch!

eu9ene (approved):
Ok, I think it's good enough to start our review, and we can fix the compression and add more stats in a follow-up PR.

PR #650, opened 2024-05-30
Fix the monolingual nllb dataset to use zst and report the sentence count

gregtatum:
This is the fast follow after merging #647.

### DIFF
--- a/configs/en-el-spring-2024.yml
+++ b/configs/en-el-spring-2024.yml
@@ -2,7 +2,7 @@
 # task config-generator -- en el --name spring-2024
 #
 # The documentation for this config can be found here:
-# https://github.com/mozilla/firefox-translations-training/blob/3d592e570b8919dff511cd8af238335ed84c8d63/taskcluster/configs/config.prod.yml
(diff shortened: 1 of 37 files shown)


configs/en-el-spring-2024.yml
```diff
@@ -182,6 +182,7 @@ datasets:
   - mtdata_ELRC-greek_legislation_anticorruption_plan-1-ell-eng # ~4,428 sentences (500.4 kB)
   - mtdata_ELRC-convention_transfer_sentenced_persons-1-ell-eng # ~131 sentences (14.8 kB)
   - mtdata_ELRC-international_judicial_cooperation_civil_matters-1-ell-eng # ~427 sentences (48.3 kB)
+  - mtdata_ELRC-letter_rights_persons_arrested_or_detained-1-ell-eng # ~2,078 sentences (234.8 kB)
```
eu9ene, line 185: hm, some of those datasets are changing places

eu9ene (approved):
Looks good! We have a lot of monolingual data, maybe even too much for some languages like RU. Also, I still have a concert of processing big shards. Would it be possible to split NLLB data to the fixed size shards? Say 50M lines each

configs/en-el-spring-2024.yml
gregtatum: I guess we should sort the results for more stable re-generation.

PR #657, opened 2024-05-31
Fix sacrebleu importing for datasets with the language pair flipped.

### LINKED ISSUE #74: tedx download via sacrebleu fails due to directionality of the data set
When training a French system, we ran into a failed "download" by sacrebleu because the training pipeline presumes it can pass en-fr to sacrebleu when the data set is only fr-en (indicating the direction in which it was translated).  

https://github.com/mjpost/sacrebleu/issues/177

This will also impact other languages with the same data set el es it pt it ru depending on direction.

gregtatum:
Resolves #74. This was the cause of this failure: https://firefox-ci-tc.services.mozilla.com/tasks/E5VIGOY_SYCfqfDHROtTBg


gregtatum:
Re-opening from a different remote.

PR #658, opened 2024-05-31
Fix data importing for sacrebleu

### LINKED ISSUE #74: tedx download via sacrebleu fails due to directionality of the data set

gregtatum:
Resolves #74. The directionality of the language pairs is not specified in sacrebleu, and it can fail when we get it wrong. [skip ci]

### DIFF
--- a/pipeline/data/importers/corpus/sacrebleu.sh
+++ b/pipeline/data/importers/corpus/sacrebleu.sh
@@ -16,7 +16,41 @@ dataset=$4
 COMPRESSION_CMD="${COMPRESSION_CMD:-pigz}"
 ARTIFACT_EXT="${ARTIFACT_EXT:-gz}"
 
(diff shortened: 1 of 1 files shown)


gregtatum:
I've validated this fix here: https://firefox-ci-tc.services.mozilla.com/tasks/groups/PzLlsmuDQ7eMfm3pxYgxDA#sacrebleu

I'd like to merge this in release, and we can move fixes back into main. I used this push to kick off cs-en again.

gabrielBusta (approved):

PR #662, opened 2024-06-03
Close resources in alignments for each step

gregtatum:
I was debugging alignment failures, and noticed that file handles are only closed at the end of the script running. It would be better to close things after each step is run. I also expanded the docs a bit more. For the review I would recommend hiding whitespace changes, as a lot of indentation changed.

screenshot of the hide whitespace widget

This PR is lower priority, and doesn't fix anything in the current training run. It was just a drive-by as I was looking into the code.


eu9ene:
I think I did that while doing the optimization work, so we can close this one

PR #664, opened 2024-06-05
Upgrade to Taskgraph 8

bhearsum:
There's nothing specifically we need here at the moment, but my work in #466 will require it, and it's generally better to stay up to date.


bhearsum:
Unfortunately, the `taskgraph-diff` failure is expected across major versions (we try to run both the before and after version with the same taskgraph version, which obviously doesn't work). I checked locally, and the task definitions end up identical with and without this patch.

taskcluster/translations_taskgraph/transforms/skip_unless_pipeline_changed.py
```diff
@@ -36,9 +36,17 @@
 transforms = TransformSequence()
 
 
+def check(files_changed, patterns):
```
gabrielBusta, line 39: Is this check something re-usable?

bhearsum: I guess ideally we would implement this as an optimization, since that's where this code moved: https://github.com/taskcluster/taskgraph/blob/eefda885ea20bb146e703282eb6e80247e651472/src/taskgraph/optimize/strategies.py#L70

Let me give that a go; it's a much better solution!

gabrielBusta (approved):

PR #677, opened 2024-06-13
Make mono shuffling deterministic to fix caching issues

### LINKED ISSUE #669: Teacher training regression
I see a super weird major regression in teacher training quality after the [fix](https://github.com/mozilla/firefox-translations-training/pull/658) of the Sacrebleu importer in the release branch.

The configs for uk-en were identical, moreover sacrebleu is not used in dev/test datasets, so there's no reason for it to regress. I see the same for the other restarted models: cs-en, el-en, ro-en, sk-en. COMET evals are ~0.5 (meaning garbage) for all of them.

[Proper training uk-en train action](https://firefox-ci-tc.services.mozilla.com/tasks/YKY97vZVTdODzmzG0q6rAA)
[Regressed train action](https://firefox-ci-tc.services.mozilla.com/tasks/Tat9yj5WT3mxcPNgI8PW5Q)

[Proper training run](https://firefox-ci-tc.services.mozilla.com/tasks/fENJrS54TlO8d1A8FXZKyQ/runs/0/logs/public/logs/live.log)
[Regressed run](https://firefox-ci-tc.services.mozilla.com/tasks/Rmx1-mEcTRKBxNEYZaA1ww/runs/0/logs/public/logs/live.log)

Screenshot 2024-06-11 at 1 45 24 PM

@bhearsum @gabrielBusta I wonder if it can be somehow related to infrastructure. Did we implement any changes on infra level  recently that might affect the release branch?

eu9ene:
We use the same code here, so it should work:

https://github.com/mozilla/firefox-translations-training/blob/fd2f7da7a47eaeb9dde92a47f250afb000edb465/pipeline/translate/merge-corpus.sh#L57

I also checked randomization in other places of the pipeline and we have seeds set there.

fixes #669

### DIFF
--- a/pipeline/clean/merge-mono.sh
+++ b/pipeline/clean/merge-mono.sh
@@ -32,6 +32,15 @@ datasets=( "${@:3}" )
 
 COMPRESSION_CMD="${COMPRESSION_CMD:-pigz}"
 
+# https://stackoverflow.com/questions/41962359/shuffling-numbers-in-bash-using-seed
+# Deterministic shuffling
+get_seeded_random()
+{
+  seed="$1"
+  openssl enc -aes-256-ctr -pass pass:"$seed" -nosalt \
+    </dev/zero 2>/dev/null
+}
+
 echo "### Merging the following datasets:"
 ls $datasets
 
@@ -40,7 +49,7 @@ mkdir -p "${dir}"
 
 ${COMPRESSION_CMD} -dc "${datasets[@]}" |
   ${BIN}/dedupe |
-  shuf -n "${max_sentences}" |
+  shuf -n "${max_sentences}" --random-source=<(get_seeded_random 42) |
   ${COMPRESSION_CMD} >"${output}"
 
 
--- a/taskcluster/translations_taskgraph/parameters.py
+++ b/taskcluster/translations_taskgraph/parameters.py
@@ -38,7 +38,7 @@ def get_defaults(_):
             },
             "marian-args": {
                 "training-backward": {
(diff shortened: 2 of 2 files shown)


bhearsum (approved):
As I understand it, this will ensure that all merge mono jobs will use the same random seed, which will produce the same results as long as their inputs are the same.

It will not prevent `merge-mono` jobs from one training from using `merge-mono` jobs from another, but that _should_ be okay because part of the cache digest of those jobs is their upstream dependencies. This means that each `merge-mono` task that has a distinct set of upstream datasets should end up with a different digest.

(This actually might have an interesting side effect of reducing the amount of `mono` tasks we run if we let one training run finish that chain before kicking off the rest.)

eu9ene:
@bhearsum thanks for the detailed analysis! Yes, your understanding is correct. The goal is to make this step deterministic and still reuse the same tasks across runs which will save some resources when training at scale. It's indeed a good idea to run one model first and then use the cached mono branch by the others.

We'll need to run many more training pairs in this training package, so we should implement a proper fix regardless of our approach to the 5 failed teachers. My current idea is to just reuse the backward model for them and rerun everything else with the busted caches which should guarantee the correct behaviour, otherwise, it would be hard to sync the two different paths of merge-mono+split+collect. You can see the configs for the affected languages that follow this idea [here](https://github.com/mozilla/firefox-translations-training/pull/678/commits/a95f672547afe2b898c450f3a2536d688a41eb75). The downside is that it will rerun the decoding with the backwards model. I'm open to exploring other options too.

PR #678, opened 2024-06-13
Configs stage2

eu9ene:
Latest config updates.


bhearsum:
@eu9ene - does this still need our review? If so, can you clarify what you're looking for? We can obviously help make sure things like `existing_tasks` are set up correctly, but changes to the `teacher-ensemble` and such are out of our scope AFAIK.

eu9ene:
> @eu9ene - does this still need our review? If so, can you clarify what you're looking for? We can obviously help make sure things like `existing_tasks` are set up correctly, but changes to the `teacher-ensemble` and such are out of our scope AFAIK.

No, I just wanted to notify everyone what I'm doing with the configs. We can merge this later when we're done with training.

eu9ene:
The training is now completed and we can merge the final configs.

gregtatum (approved):

PR #682, opened 2024-06-18
allow for more than 99 training datasets (fixes #653)

### LINKED ISSUE #653: Task has too many dependencies
https://firefox-ci-tc.services.mozilla.com/tasks/ZqlokLMTQG-pZPJtK9UnOw/runs/0/logs/public/logs/live.log

```Exception: task merge-corpus/merge-corpus-da-en has too many dependencies (105 > 99)```

### PR DESCRIPTION
**Pull request by bhearsum**
The most visible change here is the addition of `post-bicleaner-dummy` tasks that sit between `bicleaner` and `merge-corpus`. These tasks depend on up to 98 bicleaner tasks themselves, and republish the artifacts from each. `merge-corpus` pulls artifacts directly from these dummy tasks.

Ideally we would still pull artifacts from the `bicleaner` tasks, but doing so is not a trivial matter. I did some math on the cost of republishing and my best estimate is in the 3 figures per year if we had 100 full sized training runs - so IMO it's not worth the cost or complexity to take another path.

We can depend on up to 98 `post-bicleaner-dummy` tasks, so we can have 98*98 training datasets now (~9600), which is more than enough for the forseeable future.

There was some necessary refactoring and preamble before fixing this which is described in the individual commits.

### DIFF
--- a/taskcluster/kinds/merge-corpus/kind.yml
+++ b/taskcluster/kinds/merge-corpus/kind.yml
@@ -17,66 +17,6 @@ kind-dependencies:
     - bicleaner
     - toolchain
 
-task-defaults:
-    attributes:
-        src_locale: "{src_locale}"
-        trg_locale: "{trg_locale}"
-        cache:
-            resources:
-                - pipeline/clean/merge-corpus.sh
-    task-context:
-        from-parameters:
-            src_locale: training_config.experiment.src
-            trg_locale: training_config.experiment.trg
-        substitution-fields:
-            - name
-            - label
-            - description
-            - worker.env
-            - dependencies
-            - fetches
-            - attributes
-            - run.command
-    upstreams-config:
-        upstream-artifacts:
-            - "{dataset_sanitized}.{src_locale}.zst"
-            - "{dataset_sanitized}.{trg_locale}.zst"
-    worker-type: b-cpu-largedisk
-    worker:
-        docker-image: {"in-tree": "train"}
-        max-run-time: 86400
-        artifacts:
-            - name: public/build
-              path: /builds/worker/artifacts
-              type: directory
-        env:
-            SRC: "{src_locale}"
-            TRG: "{trg_locale}"
-            COMPRESSION_CMD: zstdmt
-            ARTIFACT_EXT: zst
-        # 128 happens when cloning this repository fails
-        retry-exit-status: [128]
-
-    # Don't run unless explicitly scheduled
-    run-on-tasks-for: []
-
-    run:
-        using: run-task
-        command:
-            - bash
-            - -c
-            # Arguments are:
-            # 1) output prefix
-            # 2) input files
-            - >-
-                export BIN=$MOZ_FETCHES_DIR &&
-                $VCS_PATH/pipeline/clean/merge-corpus.sh
-                artifacts/{artifact_prefix}
-                $MOZ_FETCHES_DIR/*.zst
-    fetches:
-        toolchain:
-            - preprocess
-
 tasks:
     merge-corpus:
         label: merge-corpus-{src_locale}-{trg_locale}
@@ -84,15 +24,70 @@ tasks:
         attributes:
             dataset-category: train
             stage: merge-corpus
+            src_locale: "{src_locale}"
+            trg_locale: "{trg_locale}"
             cache:
                 type: merge-corpus
+                resources:
+                    - pipeline/clean/merge-corpus.sh
+
+        task-context:
+            from-parameters:
+                src_locale: training_config.experiment.src
+                trg_locale: training_config.experiment.trg
+            from-object:
+                artifact_prefix: corpus
+            substitution-fields:
+                - name
+                - label
+                - description
+                - worker.env
+                - dependencies
+                - fetches
+                - attributes
+                - run.command
 
         upstreams-config:
+            upstream-artifacts:
+                - "{dataset_sanitized}.{src_locale}.zst"
+                - "{dataset_sanitized}.{trg_locale}.zst"
             upstream-task-attributes:
                 cleaning-type:
                     by-cleaning-type:
                         bicleaner-ai: bicleaner-ai
 
-        task-context:
-            from-object:
-                artifact_prefix: corpus
+        worker-type: b-cpu-largedisk
+        worker:
+            docker-image: {"in-tree": "train"}
+            max-run-time: 86400
+            artifacts:
+                - name: public/build
+                  path: /builds/worker/artifacts
+                  type: directory
+            env:
+                SRC: "{src_locale}"
+                TRG: "{trg_locale}"
+                COMPRESSION_CMD: zstdmt
+                ARTIFACT_EXT: zst
+            # 128 happens when cloning this repository fails
+            retry-exit-status: [128]
+
+        # Don't run unless explicitly scheduled
+        run-on-tasks-for: []
+
+        run:
+            using: run-task
+            command:
+                - bash
+                - -c
+                # Arguments are:
+                # 1) output prefix
+                # 2) input files
+                - >-
+                    export BIN=$MOZ_FETCHES_DIR &&
+                    $VCS_PATH/pipeline/clean/merge-corpus.sh
+                    artifacts/{artifact_prefix}
+                    $MOZ_FETCHES_DIR/*.zst
+        fetches:
+            toolchain:
+                - preprocess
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: taskcluster/kinds/merge-corpus/kind.yml [hunk 1]
```diff
@@ -7,92 +7,93 @@ loader: taskgraph.loader.transform:loader
...
     - taskgraph.transforms.run:transforms
     - translations_taskgraph.transforms.cached_tasks:transforms
     - taskgraph.transforms.task:transforms
 
 kind-dependencies:
-    - bicleaner
+    - post-bicleaner-dummy
     - toolchain
 
-task-defaults:
```
gabrielBusta, line 20: Just curious, why move these defaults to tasks?

taskcluster/translations_taskgraph/transforms/dependency_dummies.py
gabrielBusta, line 1: I wonder if this is something that could be re-used (like "from deps.")

gabrielBusta (approved):
The generated Taskgraph looks reasonable to me.

taskcluster/kinds/merge-corpus/kind.yml
bhearsum: Simply because there's only one entry in `task`. There's no benefit to splitting the entry. (In a much earlier iteration there was only one `merge` kind, and this was copied from that, which is why it had a `task-defaults` section.)

taskcluster/translations_taskgraph/transforms/dependency_dummies.py
bhearsum: I think it's plausible to build a more generic version of this, yeah. I don't know that this is it though. (I'm also secretly hoping that we can find a way to make Taskcluster more flexible here, and avoid the need for dummies altogether...)

gabrielBusta: If we could tweak the value per project (instead of at the instance level) that would allow for some experimentation.

eu9ene (commented):
This adds even more complexity to the already convoluted graph... Why can't we remove this limitation from the Taskcluster itself or at least change this constant to, say, 1000? Maybe we should go this way and while waiting for the fix just change the config generator to cut the smallest datasets from the training config.

taskcluster/translations_taskgraph/transforms/fetches_from_upstream.py
```diff
@@ -0,0 +1,49 @@
...
+            Optional("extra"): dict,
+        },
+    },
+    extra=ALLOW_EXTRA,
+)
+transforms = TransformSequence()
+
+
+@transforms.add
+def fetches_from_upstream(config, jobs):
```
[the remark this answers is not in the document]
gregtatum: Maybe some of your commit messages could be persisted better here in code docs.

gregtatum (approved):
> This adds even more complexity to the already convoluted graph... Why can't we remove this limitation from the Taskcluster itself or at least change this constant to, say, 1000? Maybe we should go this way and while waiting for the fix just change the config generator to cut the smallest datasets from the training config.

A mix of questions and thoughts:

I'd be curious about this as well, is there a technical reason we wouldn't just increase the limit? Looking at the solution I don't really mind the fix here, even if it does add some complexity to the graph structure.

I don't understand how the dummy chunking happens. I generated a taskgraph locally with a couple hundred fake datasets. The graph produced 3 `all-` tasks, `all-en-ru-1`, `all-en-ru-2`, `all-en-ru-3`. The chunking for post bicleaner step was named a bit weirder `post-bicleaner-dummy-ai-opus-ELRC-5067-SciPar_v1-en-ru-2` as it included a dataset name. I'm unsure where this splitting is happening.

All that said, I tested this locally, and it appears to fix the problem.

bhearsum:
On the topic of increasing the limit: I agree that is the ideal fix (and would be useful in a whole lot of other places as well). I've been chatting with the Taskcluster folks about it. The crux of the issue is that there's some inner loops that are O(dependencies) that have cause reliability issues in the past. https://bugzilla.mozilla.org/show_bug.cgi?id=1316593 speaks to this a bit, and there's a thread in #taskcluster on Matrix from June 19th with a bit more background.

It's not _impossible_ to do this, but the Taskcluster folks are hesistant because of issues in the past. I'm going to continue that conversation, and see if we can either prove that we won't hurt reliability by increasing or removing the limit, or see if we can find a way to change or rearchitect the code to remove the problematic inner loops.

I don't expect either of these things to be quick though. I'm happy to simply forego this "fix" in the meantime if we can live with the 99 training dataset limit - totally up to you folks.

bhearsum:
> I don't understand how the dummy chunking happens. I generated a taskgraph locally with a couple hundred fake datasets. The graph produced 3 all- tasks, all-en-ru-1, all-en-ru-2, all-en-ru-3. The chunking for post bicleaner step was named a bit weirder post-bicleaner-dummy-ai-opus-ELRC-5067-SciPar_v1-en-ru-2 as it included a dataset name. I'm unsure where this splitting is happening.

Hm; I may have missed a needed `from-deps` setting in `post-bicleaner-dummy` (`from-deps` is what's dealing with the chunking: https://taskcluster-taskgraph.readthedocs.io/en/latest/reference/transforms/from_deps.html). I'll look into this - thanks for pointing it out.

taskcluster/translations_taskgraph/transforms/fetches_from_upstream.py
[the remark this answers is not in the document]
bhearsum: That's a good point; will do.

bhearsum:
Just to crosslink, https://github.com/mozilla/firefox-translations-training/issues/653#issuecomment-2203472524 contains another possible solution for this problem.

bhearsum:
> I don't understand how the dummy chunking happens. I generated a taskgraph locally with a couple hundred fake datasets. The graph produced 3 `all-` tasks, `all-en-ru-1`, `all-en-ru-2`, `all-en-ru-3`. The chunking for post bicleaner step was named a bit weirder `post-bicleaner-dummy-ai-opus-ELRC-5067-SciPar_v1-en-ru-2` as it included a dataset name. I'm unsure where this splitting is happening.

To explain the naming a bit more, it's done in a few stages. In taskgraph terms, `label` is actually the thing we're talking about.

First, the default behaviour of `from-deps` will [derive the name for generated tasks](https://taskcluster-taskgraph.readthedocs.io/en/latest/reference/transforms/from_deps.html#generated-task-names). The default for this (which we use) is to strip off the `kind` from the `primary-dependency`, which is the first one listed in `kind-dependencies` when not explicitly specified. For `all` this is `export`, which means we end up with, eg: `en-ru` (from `export-en-ru`).

Second, `dependency_dummies` comes along and potentially splits this one task into multiple tasks (dependening on the number of dependencies). This gives us things like `en-ru-1`.

Finally, taskgraph sets the `label` based on the `name` [over here](https://github.com/taskcluster/taskgraph/blob/9567ad16897098983796a92be576e4ca87b3c9e3/src/taskgraph/transforms/task.py#L918-L921), which is how the kind name (`all` in this example) gets prepended.

(All of this can be bypassed by setting an explicit `label`, but that doesn't work very well when you're generating more than 1 task from a single `kind` entry.)

eu9ene:
I'm ok with merging this if there's no other way based on https://bugzilla.mozilla.org/show_bug.cgi?id=1316593.

bhearsum:
> I'm ok with merging this if there's no other way based on https://bugzilla.mozilla.org/show_bug.cgi?id=1316593.

As it turns out, we're making some headway in adjusting the maximum number of dependencies in the past few days! See https://github.com/taskcluster/taskcluster/issues/7151. It's not a guarantee that it will happen, but recent investigations are encouraging.

If you'd like to hold off on this until we see the outcome of that I'm happy to let it sit.

eu9ene:
> > I'm ok with merging this if there's no other way based on https://bugzilla.mozilla.org/show_bug.cgi?id=1316593.
> 
> As it turns out, we're making some headway in adjusting the maximum number of dependencies in the past few days! See [taskcluster/taskcluster#7151](https://github.com/taskcluster/taskcluster/issues/7151). It's not a guarantee that it will happen, but recent investigations are encouraging.
> 
> If you'd like to hold off on this until we see the outcome of that I'm happy to let it sit.

I think we're not in a rush to land this since our workaround with removing small datasets works fine for now.

bhearsum:
Taskcluster has gained support for 10,000 dependencies per task. This is no longer needed. After #858 merges we'll be able to take advantage of this.

PR #683, opened 2024-06-18
Override the existing_tasks explicitly provided in the action's input

gabrielBusta:
This is a way to explicitly replace a task in a group with another task of the same kind from another group.


eu9ene (approved):
Look working! https://firefox-ci-tc.services.mozilla.com/tasks/Gy029KR8RSiVt4OxvuSP7

bhearsum (changes requested):
I'm not sure this is ideal, but I'm fine with it, and don't have any better ideas at the moment.

Please update this to explicitly log the labels and task ids of any `existing_tasks` that end up overridden to aid future debugability.

gabrielBusta:
Done @bhearsum :)

bhearsum (approved):

PR #696, opened 2024-06-24
Publish OpusTrainer and Marian logs to W&B logs tab

### LINKED ISSUE #522: Publish task logs from Taskcluster
We are especially interested in publishing the full training live.log as a file to W&B artifacts or logs (wherever it will be more convenient to view it). 

This can be published when the run is finished. We can view the live log directly in Taskcluster. I assume we can publish the live log not only for training but also for evaluation steps since it will be a very similar mechanism (likely transform).

vrigal:
Closes #522

Tested from task group `N1O85rIASLmCwKfUpCvTlw`: https://wandb.ai/teklia/696-da-en

WandB client handles logging from stdout.
The inconvenient of this method is that the parser will output all parsed lines at the end of the script, but it is probably the less "hacky" way to do it.


eu9ene (approved):

vrigal:
@eu9ene I simplified parser logs' storage, so we should have all the lines now. I updated results at https://wandb.ai/teklia/696-da-en.

eu9ene:
@vrigal please resolve conflicts so I can merge. It's likely `poetry lock`

Issue #618, opened 2024-05-21
bring snakepit machines online with taskcluster


bhearsum:
We started doing this in https://mozilla-hub.atlassian.net/browse/RELOPS-575 but hit issues because the host OS is too old. I think our best path forward here is likely to be:
1) Upgrade or reimage to get a more modern Ubuntu version on the host
2) Wait for #391 so we run tasks in a docker container

cc @aerickson for visibility or other ideas; it's likely going to be a month or two before we get back around to this though.

PR #700, opened 2024-06-25
switch GPU workers to generic-worker & dockerize GPU tasks

### LINKED ISSUE #710: switch to generic worker for all tasks
docker-worker is deprecated, and causes us numerous issues, eg: #582, lack of memory monitoring, and others.

We need to switch to generic-worker for all tasks.

Some (but maybe not all) things we need to sort out before we can do this are: 
- https://github.com/mozilla/firefox-translations-training/issues/640
- https://github.com/mozilla/firefox-translations-training/issues/630
- https://github.com/mozilla/firefox-translations-training/issues/538

bhearsum:
This will be necessary for #466 due to the "simple" engine not supporting chain of trust. It also has some other benefits:
* Moves us onto a common worker image that is more well maintained than the existing one
* Dockerizes GPU tasks

Overall, this work fixes #710, #538, and #391. It unblocks #466 and #618.


bhearsum:
It looks like these new workers largely work fine. We are hitting https://github.com/taskcluster/taskcluster/issues/7128, which will need to be fixed before this is landable, but all of the GPU tasks seem to work well within Docker on this new image.

bhearsum:
> It looks like these new workers largely work fine. We are hitting [taskcluster/taskcluster#7128](https://github.com/taskcluster/taskcluster/issues/7128), which will need to be fixed before this is landable, but all of the GPU tasks seem to work well within Docker on this new image.

https://github.com/taskcluster/taskcluster/pull/7428 has this patch. We'll need a new worker image when that's done to pick it up. I also want to wait on https://github.com/mozilla-releng/fxci-config/pull/244 to land to avoid some unnecessary confusion in this repo.

bhearsum:
The full pipeline has been run with the most up-to-date version of this PR. https://firefox-ci-tc.services.mozilla.com/tasks/groups/ZZChI5CvQB6g98JAjsEGzw has the first half, while https://firefox-ci-tc.services.mozilla.com/tasks/groups/VSlpveRyTQq5sN6oYNb4ng has the rest. (The first one didn't complete because I rebased this while it was running, which cancelled some of the tasks.)

taskcluster/kinds/evaluate-quantized/kind.yml
```diff
@@ -84,12 +85,14 @@ tasks:
...
         run:
             using: run-task
+            use-caches: [checkout, pip]
             command:
                 - bash
                 - -c
                 - >-
                     export PATH=$PATH:~/.local/bin &&
                     export PYTHONPATH=$PYTHONPATH:$VCS_PATH &&
+                    export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:$MOZ_FETCHES_DIR/cuda-toolkit/lib64" &&
```
eu9ene, line 95: I don't know if you've looked at it but we should make sure the GPUs are being utilized. Sometimes there are issues with that under Docker.

taskcluster/kinds/translate-corpus/kind.yml
eu9ene, line 1: `translate-*` tasks also use GPUs, do we need `export LD_LIBRARY_PATH...` there?

eu9ene (commented):
I'm happy to see us dockerizing more and more things! :) 

Looks good overall, but I have a question about the `translate` tasks that I'd like to clarify before approval.

taskcluster/kinds/evaluate-quantized/kind.yml
bhearsum: I had a look at the graphs in GCP. For the translate tasks, the run _so_ quickly on the PR pipeline that barely anything shows up there. For a `train-student` task we do see an interesting difference. The logs certainly show the GPUs being found, but the task in this PR seems to have very low GPU utilization:
image

vs. [a comparable task](https://firefox-ci-tc.services.mozilla.com/tasks/Vge3ZOazSFu13IufBE2hMQ) on the old image:
image

Ram usage was identical between the two. The task here took almost twice as long as the comparable one (which I presume would be expected if it's not utilizing the GPU as much?)

`backwards` and `teacher` seem to be slightly slower here as well (5min vs 3min and 15min vs 13min respectively).

I also compared the task logs and timings. A [translate task in this PR](https://firefox-ci-tc.services.mozilla.com/tasks/MPdhSVpqRaeS6sqCSpFfhA) took ~10 seconds of wall time to translate. It had message in its log that indicated it saw all 4 GPUs:
```
[task 2025-02-26T14:13:32.003Z] [2025-02-26 14:13:32] [memory] Reserving 85 MB, device gpu2
[task 2025-02-26T14:13:32.006Z] [2025-02-26 14:13:32] [memory] Reserving 85 MB, device gpu1
[task 2025-02-26T14:13:32.009Z] [2025-02-26 14:13:32] [memory] Reserving 85 MB, device gpu0
[task 2025-02-26T14:13:32.025Z] [2025-02-26 14:13:32] [memory] Reserving 85 MB, device gpu3
```

A [comparable translate task](https://firefox-ci-tc.services.mozilla.com/tasks/fflnKc-OSje9QoooGeAW-w) that used the old image without docker had virtually identical wall clock time and log messages.

@eu9ene, do you have any pointers on known issues around GPU utilization in Docker? (This is the first I've heard about it.)

eu9ene: As soon as we see that the GPUs are being used I wouldn't worry. The issues are usually in using CPUs silently instead of GPUs because the Docker container is not set up correctly. It seems it's not the case if you see GPUs working on the dashboard.

bhearsum: OK, that's good to know. I do have a couple of questions for you:
* Are you concerned at all about what appears to be a regression in performance of the training tasks?
* Could you help me come up with a non-trivial training run to try out here? Nothing we would ship, of course, just something that would have some of the training tasks run for longer / with more data, such that we might suss out any other performance or scaling issues.

taskcluster/kinds/translate-corpus/kind.yml
bhearsum: I'll look into this; thanks for noticing it.

taskcluster/kinds/evaluate-quantized/kind.yml
eu9ene: I'm not concerned about things we see in CI, it's a very limited run. 

I can generate a config for some language pair and remove some datasets for it to be smaller, then look at utilization. I think @gregtatum has been looking at GPU utilization too lately. We can compare it to the previous training runs using W&B dashboards, but they track only training and not translation. Translation now reports GPU utilization itself in the logs. Let me look into it.

taskcluster/kinds/translate-corpus/kind.yml
bhearsum: It turns out we got lucky and this worked because the `RUNPATH` of `marian-decoder` lines up with where it is on disk:
```
❯ readelf -d marian-decoder | grep 'R.*PATH'
 0x000000000000001d (RUNPATH)            Library runpath: [/builds/worker/fetches/cuda-toolkit/lib64]
```

(I think the same is true in the training tasks.)

We should set it explicitly to make sure we get exactly the toolchain we want, though.

taskcluster/kinds/evaluate-quantized/kind.yml
bhearsum: @eu9ene provided me with a config elsewhere; I've fired off that training in https://firefox-ci-tc.services.mozilla.com/tasks/groups/GWtYq8JPSMqBa-hhcihAAw.

bhearsum: One thing has uncovered already is that bicleaner tasks were not using GPUs :(. I'll be pushing a fix for that and redoing.

bhearsum:
OK, bicleaner is trickier than I thought to deal with - most likely it's an issue with the cuda and/or cudnn version. Moving back to draft while I sort that out.

taskcluster/kinds/evaluate-quantized/kind.yml
bhearsum: I believe I've sorted out the issue with bicleaner; I've fired off a larger training run again in https://firefox-ci-tc.services.mozilla.com/tasks/groups/D1LGtquFR8C3mKhFkiAVQA

bhearsum:
The large test run I've been running uncovered an issue with spot termination handling. I've filed https://github.com/taskcluster/taskcluster/issues/7594 for this; it will require an updated worker image once that bug is fixed.

bhearsum:
We've now got another new worker image that should fix the aforementioned spot terminations. I'll verify that with a small run, and then re-do the en-ko one again.

bhearsum:
The latest full training run (which should handle spot terminations properly) is at https://firefox-ci-tc.services.mozilla.com/tasks/groups/Gu0Nb8GcS0SXfqLHUg_GoA

taskcluster/kinds/evaluate-quantized/kind.yml
bhearsum: @eu9ene - I've gotten the full `en-ko` run over in https://firefox-ci-tc.services.mozilla.com/tasks/groups/Gu0Nb8GcS0SXfqLHUg_GoA. Can you help assess whether the output is reasonable?

(That group did not end up getting any spot terminations unfortunately...I'm going to fire off of a different run to verify that those will be handled correctly with the new worker image.)

eu9ene: It looks good

eu9ene (approved):

Issue #689, opened 2024-06-19
tests fail with numpy 2.0


bhearsum:
```
[task 2024-06-19T13:43:12.117Z] ==================================== ERRORS ====================================
[task 2024-06-19T13:43:12.117Z] _________________ ERROR collecting tests/test_tracking_cli.py __________________
[task 2024-06-19T13:43:12.117Z] ../../.cache/pypoetry/virtualenvs/firefox-translations-training-hlUNdH3O-py3.10/lib/python3.10/site-packages/_pytest/runner.py:341: in from_call
[task 2024-06-19T13:43:12.117Z]     result: Optional[TResult] = func()
[task 2024-06-19T13:43:12.117Z] ../../.cache/pypoetry/virtualenvs/firefox-translations-training-hlUNdH3O-py3.10/lib/python3.10/site-packages/_pytest/runner.py:372: in <lambda>
[task 2024-06-19T13:43:12.117Z]     call = CallInfo.from_call(lambda: list(collector.collect()), "collect")
[task 2024-06-19T13:43:12.117Z] ../../.cache/pypoetry/virtualenvs/firefox-translations-training-hlUNdH3O-py3.10/lib/python3.10/site-packages/_pytest/python.py:531: in collect
[task 2024-06-19T13:43:12.117Z]     self._inject_setup_module_fixture()
[task 2024-06-19T13:43:12.117Z] ../../.cache/pypoetry/virtualenvs/firefox-translations-training-hlUNdH3O-py3.10/lib/python3.10/site-packages/_pytest/python.py:545: in _inject_setup_module_fixture
[task 2024-06-19T13:43:12.117Z]     self.obj, ("setUpModule", "setup_module")
[task 2024-06-19T13:43:12.117Z] ../../.cache/pypoetry/virtualenvs/firefox-translations-training-hlUNdH3O-py3.10/lib/python3.10/site-packages/_pytest/python.py:310: in obj
[task 2024-06-19T13:43:12.117Z]     self._obj = obj = self._getobj()
[task 2024-06-19T13:43:12.117Z] ../../.cache/pypoetry/virtualenvs/firefox-translations-training-hlUNdH3O-py3.10/lib/python3.10/site-packages/_pytest/python.py:528: in _getobj
[task 2024-06-19T13:43:12.117Z]     return self._importtestmodule()
[...]

PR #703, opened 2024-06-25
Optimize alignments

### LINKED ISSUE #507: Improve implementation of alignments
Issues with the current implementation:

- We use naive tokenization because it's what OpusTrainer requires. This might produce alignments of lower quality because we don't take into account punctuation and also the vocabulary for eflomal is getting very large. Ideally, we should switch somehow to Moses tokenization that also separates punctuation. 

- It's likely also more efficient and faster to process the Moses tokenized text due to smaller vocabulary 

- Because we don't do str.split() explicitly there might be some double spaces that might lead to discrepancies in tokenization of alignments and what OpusTrainer does

- We can see some warnings while training `[Trainer] [WARNING] Skipping line because of exception: ValueError('Out-of-bound alignment pairs')` likely related to the different whitespace, but it requires further investigation. There are not too many of them.

### LINKED ISSUE #663: Examine strategies for more efficient alignments
I haven't looked into this too deeply, but we are failing with OOM when computing alignments with eflomal.

https://firefox-ci-tc.services.mozilla.com/tasks/WoiZo-oDQAuRuN_yTu2EKw

Perhaps there is a more efficient way to do this, or we need chunking. Right now we are just increasing machine memory size. There could also be a memory leak in the implementation. It might be worth looking into, especially when we go to train high resource languages.

```
[task 2024-06-02T17:19:20.158Z] /fetches/mono.en.zst : 25511 MB...     
[task 2024-06-02T17:19:20.158Z]                                                                                
[task 2024-06-02T17:19:20.158Z] /builds/worker/fetches/mono.en.zst: 26827774098 bytes 
[task 2024-06-02T17:19:21.064Z] [alignments] Using provided priors: /builds/worker/fetches/corpus.priors
[task 2024-06-02T17:19:21.064Z] [alignments] Calculating alignments...
[task 2024-06-02T18:15:25.545Z] [eflomal] Prepared 200000000 sentences for alignment
[task 2024-06-02T18:15:25.545Z] [eflomal] Reading lexical priors...
[task 2024-06-02T18:17:15.950Z] [eflomal] 15689941 (of 25390768) pairs of lexical priors used
[task 2024-06-02T18:18:28.093Z] /builds/worker/.local/lib/python3.10/site-packages/eflomal/bin/eflomal -m 3 -s /tmp/tmpoojka2as -t /tmp/tmpd_2c0qs7 -n 3 -N 0.2 -1 2 -2 1 -3 2 -f /builds/worker/artifacts/tmp/aln.fwd -r /builds/worker/artifacts/tmp/aln.rev -p /tmp/tmpgxzt5ktb
[task 2024-06-02T18:28:18.390Z] Read texts (200000000 sentences): 590.297 s
[task 2024-06-02T18:28:18.390Z] Vocabulary sizes are 21977072 (source), 14776735 (target)
[task 2024-06-02T18:29:23.088Z] Created alignment structures: 64.692 s
[task 2024-06-02T18:29:45.552Z] Created alignment structures: 87.154 s
[task 2024-06-02T18:30:12.480Z] Randomized alignment: 49.392 s
[task 2024-06-02T18:30:12.480Z] Aligning with model 1 (2 iterations)
[task 2024-06-02T18:30:32.341Z] Randomized alignment: 46.788 s
[task 2024-06-02T18:30:32.341Z] Aligning with model 1 (2 iterations)
[task 2024-06-02T18:38:40.182Z] Traceback (most recent call last):
[task 2024-06-02T18:38:40.254Z]   File "/builds/worker/checkouts/vcs/pipeline/alignments/align.py", line 227, in <module>
[task 2024-06-02T18:38:40.254Z]     main()
[task 2024-06-02T18:38:40.254Z]   File "/builds/worker/checkouts/vcs/pipeline/alignments/align.py", line 216, in main
[task 2024-06-02T18:38:40.254Z]     run(
[task 2024-06-02T18:38:40.254Z]   File "/builds/worker/checkouts/vcs/pipeline/alignments/align.py", line 53, in run
[task 2024-06-02T18:38:40.254Z]     fwd_path, rev_path = align(
[task 2024-06-02T18:38:40.254Z]   File "/builds/worker/checkouts/vcs/pipeline/alignments/align.py", line 97, in align
[task 2024-06-02T18:38:40.254Z]     aligner.align(
[task 2024-06-02T18:38:40.254Z]   File "/builds/worker/.local/lib/python3.10/site-packages/eflomal/__init__.py", line 72, in align
[task 2024-06-02T18:38:40.271Z]     align(srcf.name, trgf.name,
[task 2024-06-02T18:38:40.271Z]   File "python/eflomal/eflomal.pyx", line 161, in eflomal.cython.align
[task 2024-06-02T18:38:40.502Z]   File "/usr/lib/python3.10/subprocess.py", line 526, in run
[task 2024-06-02T18:38:40.575Z]     raise CalledProcessError(retcode, process.args,
[task 2024-06-02T18:38:40.576Z] subprocess.CalledProcessError: Command '['/builds/worker/.local/lib/python3.10/site-packages/eflomal/bin/eflomal', '-m', '3', '-s', '/tmp/tmpoojka2as', '-t', '/tmp/tmpd_2c0qs7', '-n', '3', '-N', '0.2', '-1', '2', '-2', '1', '-3', '2', '-f', '/builds/worker/artifacts/tmp/aln.fwd', '-r', '/builds/worker/artifacts/tmp/aln.rev', '-p', '/tmp/tmpgxzt5ktb']' died with <Signals.SIGKILL: 9>.
```

eu9ene:
- Add Moses tokenization of the corpus to reduce the vocabulary and increase accuracy
- Remap the calculated alignments back to whitespace-tokenized ones expected by OpusTrainer
- Use a fast C++ tokenizer
- Use parallelization with multiprocessing to speed things up

Known issues:
- I tested it on a student 400M corpus on 512GB machine without priors and it didn't OOM, but it took ~2 days to complete
- We can likely make it more scalable by using chunks, especially if we use the priors pre-calculated on the teacher step. However, it will not help with preemptions if we don't split it into multiple tasks
- To run it with existing teachers we'll need to separately run `target: alignments-original` to recalculate the priors with the new tokenization and then run it again with `target: all` and `existing_tasks: { "alignments-original-src-trg": "<task_id>" }`

fixes #507 
fixes #663


eu9ene:
The tests will fail due to https://github.com/mozilla/firefox-translations-training/issues/689

eu9ene:
It turned out to be hard to restart only the alignments-original, so I added an extra task to recalculate the priors for the student alignments (in a separate branch). I'm testing it here for en-uk https://firefox-ci-tc.services.mozilla.com/tasks/groups/eZKkxqHISTCDwrsylAZZvA. This way we won't need to rerun things multiple times.

bhearsum (approved):
The taskgraph parts looks fine to me.

pipeline/alignments/align.py
```diff
@@ -153,17 +192,105 @@ def write_priors(
...
+        lines = zip(
+            stack.enter_context(open(src_path)),
+            stack.enter_context(open(trg_path)),
+            stack.enter_context(open(tok_src_path)),
+            stack.enter_context(open(tok_trg_path)),
+            stack.enter_context(open(aln_path)),
+        )
+
+        # send lines to worker processes in chunks
+        for aln in tqdm(pool.imap(remap_line, lines, chunksize=10000), mininterval=10):
```
gregtatum, line 243: Question: Did you profile any of these changes to see the impact? Parallelism can speed up tasks, but will increase the memory pressure. I'd be curious of the memory/timing differences here.

pipeline/alignments/align.py
```diff
@@ -153,17 +192,105 @@ def write_priors(
...
+            output.write(aln)
+
```
gregtatum, line 245: Thought:

Something to measure process memory on tasks with many steps could be interesting.

Something like:

```
import psutil


def print_memory():
    processes = []

    # Collect process information
    for proc in psutil.process_iter(["pid", "name", "memory_info"]):
        try:
            pid = proc.info["pid"]
            name = proc.info["name"]
            memory_info = proc.info["memory_info"]
            memory_usage = memory_info.rss
            processes.append((name, pid, memory_usage))
        except Exception:
            print("Failed to get process information")

    # Sort the processes based on memory usage (descending)
    processes.sort(key=lambda x: x[2], reverse=True)

    # Calculate the maximum length of "name (pid)" for alignment
    display_names = [f"{name} ({pid})" for name, pid, _ in processes]
    max_length = max(len(display_name) for display_name in display_names)

    # Print the sorted process information with right-padding
    for proc_info, display_name in zip(processes, display_names):
        _, _, memory_usage = proc_info
        print(f"{display_name.ljust(max_length)} {memory_usage / (1024 * 1024):.2f} MB")


if __name__ == "__main__":
    print_memory()
```

```
plugin-container (6622)                              1081.06 MB
firefox (6609)                                       963.39 MB
plugin-container (6618)                              845.61 MB
plugin-container (6619)                              734.81 MB
iTerm2 (88484)                                       540.56 MB
Slack Helper (Renderer) (49150)                      515.27 MB
Code Helper (Renderer) (47144)                       450.91 MB
plugin-container (81278)                             438.48 MB
plugin-container (6621)                              434.69 MB
plugin-container (17813)                             336.47 MB
plugin-container (6620)                              333.00 MB
```

pipeline/alignments/align.py
```diff
@@ -153,17 +192,105 @@ def write_priors(
...
+    Remaps alignments for a single line in a corpus
+    """
+    src, trg, tok_src, tok_trg, aln = params
+    src_map = map_indices(tok_src, src)
+    trg_map = map_indices(tok_trg, trg)
+
+    remapped_aln = []
+    for pair in aln.split():
+        src_idx, trg_idx = map(int, pair.split("-"))
+        new_pair = (src_map[src_idx], trg_map[trg_idx])
```
gregtatum, line 258: Question: What happens if this fails to map cleanly? Should we have assertions here?

pipeline/alignments/align.py
```diff
@@ -153,17 +192,105 @@ def write_priors(
...
+    tok_to_orig_indices = {}
+    orig_idx = 0
+    tok_idx = 0
+
+    # For 'Hello, world!'
+    # Map ['Hello', ',', 'world', '!'] [0, 1, 2, 3]
+    # To ['Hello,', 'world!'] [0, 1]
+    # tok -> orig: {0: 0, 1: 0, 2: 1, 3: 1}
+
+    while orig_idx < len(orig_words):
```
gregtatum, line 283: Thought: If this logic were extended a bit, we could remap arbitrary tokenizations, even if the boundaries aren't as well defined as here.

tests/fixtures/__init__.py
```diff
@@ -198,13 +198,22 @@ def run_task(
...
-        python_bin_dir = get_python_bin_dir(requirements)
+        python_bin_dir, venv_dir = get_python_dirs(requirements)
         if python_bin_dir:
             final_env = {**final_env, "PATH": f'{python_bin_dir}:{os.environ.get("PATH", "")}'}
             if command_parts[0].endswith(".py"):
                 # This script is relying on a shebang, add the python3 from the executable instead.
                 command_parts.insert(0, os.path.join(python_bin_dir, "python3"))
 
+        # We have to set the path to the C++ lib before the process is started
+        # https://github.com/Helsinki-NLP/opus-fast-mosestokenizer/issues/6
```
gregtatum, line 209: Thought: Yay for having docker.

gregtatum (approved):
Thanks for taking the time to make really clean commits here. It made it really easy to review these changes. This is some nice work.

pipeline/alignments/tokenizer.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def tokenize(input_path: str, output_path: str, lang: str, chunk_size: int = 100000) -> None:
+    logger.info(f"Tokenizing {input_path} with Moses tokenizer")
+
+    with multiprocessing.Pool(processes=multiprocessing.cpu_count()) as pool:
+        with open(output_path, "w") as output_file:
+            chunks = _read_file_in_chunks(input_path, chunk_size=chunk_size)
+
+            pbar = tqdm(mininterval=10)
```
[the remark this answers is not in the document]
eu9ene: It works fine everywhere but you should be careful with how often it writes stats, that's why I'm using mininterval. This code is being tested only inside `run_task` so, there are no issues with package installation if it's what you mean.

pipeline/alignments/align.py
eu9ene: I did not profile but I watched the CPU/memory utilization while running both locally (on 10M) and on a server (400M sentences). The bottleneck for tokenization and remapping is CPU, so I'm parallelizing it. There were no issues with memory at these stages. Memory utilization is a lot higher during the aligning process which happens inside eflomal and we don't have any control on it.

pipeline/alignments/align.py
```diff
@@ -12,101 +19,132 @@
...
-    if corpus_trg.endswith(".zst"):
-        logger.info("Decompressing target corpus...")
-        subprocess.check_call([COMPRESSION_CMD, "-d", "-f", "--rm", corpus_trg])
-        corpus_trg = corpus_trg[:-4]
+    if tokenization == Tokenization.moses:
+        tokenized_src, tokenized_trg = corpus_src + ".moses", corpus_trg + ".moses"
+        output_aln = os.path.join(tmp_dir, "aln")
+        # C++ tokenizer can process 100k sentences per second on a single core,
+        # so the chunks to parallelize things should be large enough
+        tokenize(corpus_src, tokenized_src, src, chunk_size=500000)
```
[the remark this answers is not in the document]
eu9ene: Tokenization currently takes ~20 min on 400M sentences dataset and 64 cores machine and doesn't allocate a lot of memory. Even if it's 10GB, it's the eflomal aligner that allocates ~70% on a 512GB machine, so since we're optimizing this all together having large chunks to process it faster makes sense. Also, I experimented with different values to maximize the throughput, and 500K looks fine. I see it tokenizes 1M sentences/sec for English.

pipeline/alignments/align.py
eu9ene: Agreed, proper measuring would definitely be interesting but since the main issue is inside C++ code in eflomal, we can't do much about it anyway. We can run such a tool locally on a 10M dataset if you're interested.

pipeline/alignments/align.py
eu9ene: Yes, it's what OpusTrainer does with its Retokenize modifier. I considered resuing its code but it didn't fit what we need to do here.

pipeline/alignments/align.py
eu9ene: I'd prefer just having an exception here which is essentially the same thing. It will likely be something like "out of boundaries". It is supposed to be inside the boundaries, otherwise there's a bug.

pipeline/alignments/align.py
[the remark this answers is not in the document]
gregtatum: Sounds good, thanks for the explanation.

pipeline/alignments/align.py
gregtatum: Ah, I guess the attribute access will cause an exception which is the same thing. That makes sense.

PR #704, opened 2024-06-26
Expose Takscluster task owner as author for Weight & Biases publication

### LINKED ISSUE #578: Add an author field
We should be able to identify the author of an experiment. Currently, it's always `moz-translations-wandb-bot` because of the service token we use. Let's find a way to either change this one or use another field.

La0:
Closes #578 

This PR exposes the `config.params["owner"]` from taskgraph as environment variable `WANDB_AUTHOR` so it's usable by the tracking Python script to set a Weight & Biases tag under the form `author:<owner>`


vrigal:
@La0 I cannot push directly, but only this should be missing from there. We should ask for a test run to be sure author is published as a tag. 
```diff
--- a/tracking/translations_parser/wandb.py
+++ b/tracking/translations_parser/wandb.py
@@ -140,6 +140,10 @@ def get_wandb_publisher(
     if logs_file:
         config["logs_file"] = logs_file
 
+    # Automatically adds experiment owner to the tags
+    if author := os.environ.get("WANDB_AUTHOR"):
+        tags.append(f"author:{author}")
+
     return WandB(
         project=project_name,
         group=group_name,
```

eu9ene (approved):

PR #706, opened 2024-06-27
Fix poetry lock

eu9ene:
Ok, we shouldn't lock dependencies for the package. It leads to conflicts when we try to lock things in downstream requirements. This fix prevents it to install numpy 2.0 but we should lock all other requirements together with tracking to prevent this issue in production tasks.  I filed https://github.com/mozilla/firefox-translations-training/issues/707 as a follow up. It will work now because even if we install numpy 2.0 as a part of the task requirements, then we install the tracking package and it downgrades numpy to the accepted version, but we should fix this to prevent breakage in the future.


bhearsum (approved):
If we're treating `tracking` as a library, where it's installed as one of many dependencies, I agree it doesn't make sense to lock the dependencies in _that_ module/project. It's almost always advisable to lock at the top level, though, to avoid bustage when a package upgrade inevitably happens.

I'm approving this anyways because I don't consider myself an owner on this part, but I wanted to provide this best practice advice regardless.

PR #712, opened 2024-06-28
fix: shorten non-URL dataset names that are more than 50 characters (fixes #654)

### LINKED ISSUE #654: Some mtdata datasets fail because of long name
It's related to the Taskcluster limitation of the label size (256 characters)

Example:

```
- mtdata_ELRC-convention_against_torture_other_cruel_inhuman_or_degrading_treatment_or_punishment_united_nations-1-ell-eng
```

bhearsum:
This takes the same approach as we used for custom datasets in #611 (thanks @gregtatum for making this so easy to extend!)


gregtatum (approved):
Thanks! This looks good to me.

PR #713, opened 2024-06-28
Exclude start stage tasks from existing tasks

eu9ene:
This fixes the edge case where we have alignments-original -> alignments-backtranslated both marked as `stage: alignments-teacher` and want to restart both of them. We should probably split them later to two stages but not now because it breaks the ability to reuse previous graphs.

### DIFF
--- a/taskcluster/translations_taskgraph/actions/train.py
+++ b/taskcluster/translations_taskgraph/actions/train.py
@@ -400,7 +400,13 @@ def train_action(parameters, graph_config, input, task_group_id, task_id):
         # as `existing_tasks`. These map task labels (eg: train-backwards-ru-en) to
         # task ids, and will be used instead of scheduling new tasks for any tasks with
         # an identical name.
-        parameters["existing_tasks"] = get_ancestors(start_task_ids)
+        # Make sure not to include the tasks with the start_stage in existing_tasks
+        existing_tasks = {
+            label: task_id
+            for label, task_id in get_ancestors(start_task_ids).items()
+            if task_id not in start_task_ids
+        }
+        parameters["existing_tasks"] = existing_tasks
 
     # Override the `existing_tasks` explicitly provided in the action's input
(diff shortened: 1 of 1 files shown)


bhearsum (commented):
Is this a generally good thing, or could there be groups in `previous_group_ids` where we'd want to overwrite one task from it but not another? The safer way to do this may be to allow a list of labels, substrings or prefixes to be specified in the training config, and filter anything matching them out of `existing_tasks`.

(We do [something similar for optimizations](https://searchfox.org/mozilla-central/rev/af09c55c0e6253ba8fcb69cf9534046405f4024d/taskcluster/docs/parameters.rst#177-180). We can't directly use that here, but we could use the same technique to keep things flexible.

eu9ene:
> Is this a generally good thing, or could there be groups in `previous_group_ids` where we'd want to overwrite one task from it but not another? The safer way to do this may be to allow a list of labels, substrings or prefixes to be specified in the training config, and filter anything matching them out of `existing_tasks`.
> 
> (We do [something similar for optimizations](https://searchfox.org/mozilla-central/rev/af09c55c0e6253ba8fcb69cf9534046405f4024d/taskcluster/docs/parameters.rst#177-180). We can't directly use that here, but we could use the same technique to keep things flexible.

@bhearsum It probably depends on the situation but usually, I want to restart all the tasks with `start_stage`. This is basically a hotfix so that I can at least have a way to schedule what I need. So I suggest we merge it for now and do a more comprehensive analysis in #719

gregtatum (approved):
Approving for the release branch. Can you rewrite the merge commit to have a message like `"RELEASE ONLY: Exclude start stage tasks from existing tasks"` so that we don't accidentally cherry pick it back?

eu9ene:
I think I'm doing fine with just cherry-picking this into a separate branch so we don't have to merge it at all if you have doubts it's the right solution here

PR #717, opened 2024-07-01
Switch aligning student corpus to fast_align

eu9ene:


eu9ene:
I think being able to switch to fast_align can still be useful, so I'll have another look at this one soon.

PR #722, opened 2024-07-08
Add a check that there are visible GPUs

gregtatum:
An example is here (id-en), where the task timed out.

https://firefox-ci-tc.services.mozilla.com/tasks/NWpySVWkT2SVBY3AOkDaIQ

### DIFF
--- a/pipeline/bicleaner/bicleaner.sh
+++ b/pipeline/bicleaner/bicleaner.sh
@@ -78,7 +78,7 @@ else
                # The GPU devices have failed to be found, and bicleaner AI falls back
                # to operate on the CPU very slowly. To guard against this wasting expensive
                # GPU time, always check that it can find GPUs.
-               python -c "import tensorflow; exit(0) if tensorflow.config.list_physical_devices('GPU') else exit(9001)"
+               python3 -c "import tensorflow; exit(0) if tensorflow.config.list_physical_devices('GPU') else exit(9001)"
                bicleaner-ai-classify ${hardrules} --scol ${scol} --tcol ${tcol} - - $1
        }
        export -f biclean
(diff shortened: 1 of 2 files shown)


taskcluster/kinds/bicleaner/kind.yml
```diff
@@ -80,7 +80,10 @@ tasks:
                 COMPRESSION_CMD: zstdmt
                 ARTIFACT_EXT: zst
             # 128 happens when cloning this repository fails
-            retry-exit-status: [128]
+            # 9001 is the code for when tensorflow fails to find GPUs on the system,
+            # and biclenaer reverts to CPU time. Rather than waste time, we should
+            # restart the task.
+            retry-exit-status: [128,9001]
```
eu9ene, line 86: Nice, I like the idea of targeted Taskcluster restarts based on specific exit codes. We can utilize this pattern for other tasks too instead of using restarts inside the Python code. It makes it more transparent to see when a task fails and restarts.

pipeline/bicleaner/bicleaner.sh
```diff
@@ -75,6 +75,10 @@ else
        biclean() {
                export CUDA_VISIBLE_ARRAY=(${CUDA_VISIBLE_DEVICES//,/ })
                export CUDA_VISIBLE_DEVICES=${CUDA_VISIBLE_ARRAY[$(($2-1))]}
+               # The GPU devices have failed to be found, and bicleaner AI falls back
+               # to operate on the CPU very slowly. To guard against this wasting expensive
+               # GPU time, always check that it can find GPUs.
+               python -c "import tensorflow; exit(0) if tensorflow.config.list_physical_devices('GPU') else exit(9001)"
```
eu9ene, line 81: I think you might need to use `python3` here to choose the correct version. It's what we usually do for python scripts in Taskcluster tasks.

bhearsum (approved):
Looks fine to me.

I ended up taking a look through Cloud Audit and didn't see any evidence that suggested the GPUs were spot terminated. I'm going to ask our GCP support folks as well though to see if there's a bug or some other thing that needs to be addressed.

PR #755, opened 2024-07-19
Delete temporary files after successfully generating alignments

bhearsum:
I noticed we're uploading these as part of alignments tasks, oops. This patch will remove them only when it completes successfully. If there's no value in keeping them even if something fails, I'll update it to always delete them.

### DIFF
--- a/pipeline/alignments/align.py
+++ b/pipeline/alignments/align.py
@@ -13,6 +13,7 @@ Example:
 
 import argparse
 import os
+import shutil
 import subprocess
 import sys
 from contextlib import ExitStack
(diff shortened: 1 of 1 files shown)


eu9ene (approved):

bhearsum:
My bad. It looks like we actually use these in shortlist? https://github.com/mozilla/firefox-translations-training/blob/1e9799c9bd5d567c3ccdd41114cb0a42388606f6/pipeline/alignments/generate-shortlist.sh#L33

eu9ene:
> My bad. It looks like we actually use these in shortlist?
> 
> https://github.com/mozilla/firefox-translations-training/blob/1e9799c9bd5d567c3ccdd41114cb0a42388606f6/pipeline/alignments/generate-shortlist.sh#L33

hm, maybe there shouldn't be in a tmp directory then

PR #738, opened 2024-07-16
chore: bump taskgraph to 9.2.0

bhearsum:
This picks up some fixes that are expected to fix #680.

PR #761, opened 2024-07-24
fix: invalidate caches when fetch, docker, or toolchain tasks change

bhearsum:
This was an early hack I put in to avoid, eg: retraining entire models because a comment changed in a Dockerfile. We still want to avoid this sort of thing, but these days we do so explicitly with `previous_group_ids` and other techniques in the training config.

Getting rid of this will improve our ability to catch issues in PRs. Eg: https://github.com/mozilla/firefox-translations-training/pull/738 had to be tricked into running things on a second iteration that only changed a Dockerfile.

### DIFF
--- a/taskcluster/translations_taskgraph/transforms/cached_tasks.py
+++ b/taskcluster/translations_taskgraph/transforms/cached_tasks.py
@@ -6,12 +6,6 @@
 # It exists because there are two features that we need that are missing upstream:
 # - The ability to influence the cache digest from parameters.
 #   (https://github.com/taskcluster/taskgraph/issues/391)
-# - The ability to avoid adding some upstream tasks to the cache digest (which
-#   allows us to avoid rebuilding the world when, eg: we upgrade a Docker base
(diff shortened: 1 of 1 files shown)


bhearsum:
@eu9ene - or @gregtatum: one of you should sign off on this as well.

eu9ene (approved):

taskcluster/translations_taskgraph/transforms/cached_tasks.py
```diff
@@ -6,12 +6,6 @@
 # It exists because there are two features that we need that are missing upstream:
```
jcristau, line 6: One feature left :)

jcristau (approved):

taskcluster/translations_taskgraph/transforms/cached_tasks.py
bhearsum: Indeed! I intend to follow up with this at some point.

PR #775, opened 2024-07-29
temp: use cpu worker pool that uses gpu image to see if it works

bhearsum:


bhearsum:
It didn't work! It also may not be necessary or beneficial to try to unify these, as the generic worker images we'll be using for CPU workers are already using images shared with Gecko and other pools.

Issue #628, opened 2024-05-23
Evaluation fails on a pre-trained backward model


eu9ene:
https://firefox-ci-tc.services.mozilla.com/tasks/CEUR_rZ1Qty22JNz3JC-mw

We shouldn't run evals for the pre-trained models though. It wasn't the case before, so something got broken in training continuation.

This is not critical as it does not block other tasks.

gabrielBusta:
Hmm, maybe it's because continuation used to be done at graph generation time rather than at run-time? Perhaps we can prune these eval tasks from the graph somehow. Alternatively, we could have the eval tasks exit successfully without doing anything if they detect that the model was pretrained.

gabrielBusta:
Hmm, maybe it's because continuation used to be done at graph generation time rather than at run-time? Perhaps we can prune these eval tasks from the graph using its parameters. Alternatively, we could have the eval tasks exit successfully without doing anything if they detect that the model was pretrained.

eu9ene:
We should remove any redundant tasks from the graph. We can assume the pre-trained model has already been evaluated.

bhearsum:
> Hmm, maybe it's because continuation used to be done at graph generation time rather than at run-time? Perhaps we can prune these eval tasks from the graph using its parameters. Alternatively, we could have the eval tasks exit successfully without doing anything if they detect that the model was pretrained.

As far as I can tell, the run that was linked to [is not using runtime continuation](https://firefox-ci-tc.services.mozilla.com/tasks/LphXRhZoScGYSNAPKCiVfQ). I suspect this regressed with one of the recent-ish changes to train.py: https://github.com/mozilla/firefox-translations-training/commits/main/taskcluster/translations_taskgraph/actions/train.py

PR #776, opened 2024-07-30
feat: add scaffolding and basic tests for taskgraph generation

bhearsum:
This is prep work for https://github.com/mozilla/firefox-translations-training/issues/628, where I'd like to add some tests to avoid regressing that again in the future.

The fixtures here are based on similar tests from Gecko: https://searchfox.org/mozilla-central/source/taskcluster/test. There's a bit of a terrible hack to make optimized task graphs testable, described more in the comments.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,6 +53,7 @@ sh = "^2.0.6"
 zstandard = "^0.22.0"
 translations_parser = {path="./tracking/", develop=true}
 taskcluster-taskgraph = "^10.0.1"
+translations_taskgraph = {path="./taskcluster/", develop=true}
 
 [tool.black]
 extend-exclude= "/3rd_party"
--- a/taskcluster/test/conftest.py
+++ b/taskcluster/test/conftest.py
@@ -1,0 +1,86 @@
+from copy import deepcopy
+import pytest
+import requests_mock
+from typing import Any, Dict, Generator, List, Protocol
+
+from taskgraph.generator import TaskGraph, TaskGraphGenerator
+from taskgraph.parameters import Parameters, parameters_loader
+from translations_taskgraph.util.substitution import substitute
+
+
+class CreateTgg(Protocol):
+    def __call__(
+        self, parameters: Parameters | None = None, overrides: dict | None = None
+    ) -> TaskGraphGenerator:
+        ...
+
+
+# These fixtures are largely cribbed from Gecko:
+# https://searchfox.org/mozilla-central/source/taskcluster/test
+@pytest.fixture(scope="session")
+def create_tgg():
+    def inner(
+        parameters: Parameters | None = None, overrides: dict | None = None
+    ) -> TaskGraphGenerator:
+        params = parameters_loader(parameters, strict=False, overrides=overrides)
+        return TaskGraphGenerator(None, params)
+
+    return inner
+
+
+@pytest.fixture(scope="module")
+def mock_requests() -> Generator[requests_mock.Mocker, None, None]:
+    with requests_mock.Mocker() as m:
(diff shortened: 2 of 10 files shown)


taskcluster/test/conftest.py
```diff
@@ -0,0 +1,74 @@
...
+def target_task_set(tgg: TaskGraphGenerator):
+    return tgg.target_task_set
+
+
+@pytest.fixture(scope="module")
+def optimized_task_graph(request, mock_requests, tgg: TaskGraphGenerator):
+    for resp in getattr(request.module, "MOCK_REQUESTS", {}):
+        responses = deepcopy(resp["responses"])
+        digests = {}
+        # This is a bit of a terribly hack, but it allows for cached task digests
```
ahal, line 59: nit: terrible

ahal (approved):
Nice, looks good!

taskcluster/test/conftest.py
```diff
@@ -0,0 +1,74 @@
...
+
+from taskgraph.generator import TaskGraphGenerator
+from taskgraph.parameters import parameters_loader
+from translations_taskgraph.util.substitution import substitute
+
+
+# These fixtures are largely cribbed from Gecko:
+# https://searchfox.org/mozilla-central/source/taskcluster/test
+@pytest.fixture(scope="session")
+def create_tgg():
```
gregtatum, line 13: Thought: I guess doing fixtures this way means that type hinting won't work. I'm guessing this is a pythonic practice, but I'm wondering if it would evolve with adding more robust typing requirements.

taskcluster/test/test_default_params.py
```diff
@@ -0,0 +1,97 @@
...
+]
+
+
+def test_last_task_is_targeted(target_task_set):
+    """Ensure that the last task in the pipeline is targeted by default"""
+    assert any([task == "all-ru-en-1" for task in target_task_set.tasks])
+
+
+def test_cached_tasks_optimized_away(optimized_task_graph):
+    """Ensure that any tasks found in a cache route are _not_ present
```
gregtatum, line 94: Thought: It's hard to tell what the types of the tests are with the fixture magic. Again, I think this is standard practice with pytest, but it's not great for discoverability.

Suggestion (optional): Adding at a minimum type hints here would be helpful.

gregtatum (approved):
I left a few thoughts and questions, but overall it looks workable to me. Praise: Thanks for more tests!

taskcluster/test/conftest.py
bhearsum: Yeah, this is quite standard pytest stuff (which probably predates `typing`...). I don't know of a way for fixtures to find their types based on where they are created, but I can update this to at least specify an explicit type for `create_tgg` where it's used.

taskcluster/test/test_default_params.py
bhearsum: Will do.

pyproject.toml
```diff
@@ -53,6 +53,7 @@ sh = "^2.0.6"
 zstandard = "^0.22.0"
 translations_parser = {path="./tracking/", develop=true}
 taskcluster-taskgraph = "^10.0.1"
+translations_taskgraph = {path="./taskcluster/", develop=true}
```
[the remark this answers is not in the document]
bhearsum: Is https://python-poetry.org/docs/dependency-specification/#path-dependencies what you're looking for? It's just a way to install a local dependency. It's a bit weird that the `path` is the path to the parent directory of the actual code, if that's what you're getting at. That's to allow for `pyproject.toml` (or other metadata files) to exist though, as I understand it.

PR #777, opened 2024-07-30
restrict github-push taskcluster events to `main`

bhearsum:
In https://bugzilla.mozilla.org/show_bug.cgi?id=1907217 we're becoming more explicit about scopes we grant to branches in Github, which means branches that do not show up in the explicit list in fxci-config (https://github.com/mozilla-releng/fxci-config/blob/main/projects.yml) will not be able to start tasks.

If we need to build on pushes to branches other than `main` and `release` we can amend the list here and in fxci-config. If needed, we can use a prefix to match multiple branches.


ahal (approved):

eu9ene:
@bhearsum my regular working pattern is to create experimental branches and then run experimental or even production training from their push tasks (see the amount of ad-hoc fixes we had to do for the big training). These experiments don't necessarily lead to pull requests. If we restrict it to main/release, does it mean I'll have to modify the config every time I need to run something from a branch? If so, it does not look convenient at all.

bhearsum:
> @bhearsum my regular working pattern is to create experimental branches and then run experimental or even production training from their push tasks (see the amount of ad-hoc fixes we had to do for the big training). These experiments don't necessarily lead to pull requests. If we restrict it to main/release, does it mean I'll have to modify the config every time I need to run something from a branch? If so, it does not look convenient at all.

As this patch stands - yes, but we can make these uses cases work, I think. Would it work if we allowed for `release*` and `dev*`? (Other prefixes are possible too.)

eu9ene:
> > @bhearsum my regular working pattern is to create experimental branches and then run experimental or even production training from their push tasks (see the amount of ad-hoc fixes we had to do for the big training). These experiments don't necessarily lead to pull requests. If we restrict it to main/release, does it mean I'll have to modify the config every time I need to run something from a branch? If so, it does not look convenient at all.
> 
> As this patch stands - yes, but we can make these uses cases work, I think. Would it work if we allowed for `release*` and `dev*`? (Other prefixes are possible too.)

yes, adding a prefix to every branch where I need to run things wouldn't be a problem. Let's just document those in the docs

bhearsum:
@eu9ene - How's this? We'll build for anything that matches `main` or starts with `release` or `dev`.

eu9ene (changes requested):
@bhearsum it looks good, could you also document this behavior in https://mozilla.github.io/firefox-translations-training/task-cluster.html as a part of this PR? This is something that might confuse developers initially. @gregtatum FYI we'll need to create branches with `release` or `dev` prefix to get Taskcluster decision tasks and run the Train action.

eu9ene (approved):

gregtatum:
I also filed #840 to add this new restriction to the train.py script I added recently.

bhearsum:
> I also filed #840 to add this new restriction to the train.py script I added recently.

Thanks for filing that; sorry I missed it.

PR #779, opened 2024-07-31
Expand out marian command arguments

gregtatum:
This should be a simple patch that expands out the 1 letter command args to the full word to make it easier to reason with these scripts. I added a few comments as well.

### DIFF
--- a/pipeline/cefilter/score.sh
+++ b/pipeline/cefilter/score.sh
@@ -33,17 +33,17 @@ dir=$(dirname "${output}")
 mkdir -p "${dir}"
 
 "${MARIAN}/marian-scorer" \
-  -m "${model}" \
-  -v "${vocab}" "${vocab}" \
-  -t "${corpus_prefix}.${TRG}${ARTIFACT_EXT}" "${corpus_prefix}.${SRC}${ARTIFACT_EXT}" \
+  --model "${model}" \
+  --vocabs "${vocab}" "${vocab}" \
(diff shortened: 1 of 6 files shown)


pipeline/cefilter/score.sh
```diff
@@ -33,17 +33,17 @@ dir=$(dirname "${output}")
...
   --mini-batch 32 \
   --mini-batch-words 1500 \
   --maxi-batch 1000 \
   --max-length 250 \
   --max-length-crop \
   --normalize \
-  -d ${GPUS} \
-  -w "${WORKSPACE}" \
+  --devices ${GPUS} \
+  --wworkspace "${WORKSPACE}" \
```
eu9ene, line 46: typo

eu9ene (changes requested):

pipeline/cefilter/score.sh
```diff
@@ -33,17 +33,17 @@ dir=$(dirname "${output}")
 mkdir -p "${dir}"
 
 "${MARIAN}/marian-scorer" \
-  -m "${model}" \
-  -v "${vocab}" "${vocab}" \
-  -t "${corpus_prefix}.${TRG}${ARTIFACT_EXT}" "${corpus_prefix}.${SRC}${ARTIFACT_EXT}" \
+  --model "${model}" \
```
eu9ene, line 36: @vrigal FYI this is another type of refactoring that can happen with Marian args, so experiment tracking should be robust to that

eu9ene (approved):

PR #781, opened 2024-07-31
fix: don't run evaluate tasks on pretrained models

### LINKED ISSUE #628: Evaluation fails on a pre-trained backward model

bhearsum:
This is accomplished with a new transform used by the `evaluate` tasks, which avoids yielding any tasks whose `stage` matches a `pretrained-models` stage.

Fixes #628


eu9ene (approved):
Nice addition of tests!

taskcluster/translations_taskgraph/transforms/training_continuation.py
```diff
@@ -92,3 +92,24 @@ def add_pretrained_model_mounts(config, jobs):
                 }
 
         yield job
+
+
+evaluate = TransformSequence()
```
ahal, line 97: nit: I feel like this is more of a "filter" than an evaluation.. but I'm very out of touch with nomenclature here, so it probably makes sense in context.

ahal (approved):

taskcluster/translations_taskgraph/transforms/training_continuation.py
bhearsum: Ah; in this case `evaluate` is a stage of the pipeline. What I intended by this is that this the transform(s) needed for training continuation support for the `evaluate` stage. (I can see that this is not obvious! I'll try to think of something better.)

PR #784, opened 2024-07-31
Fix restarting downloads

gregtatum:
Download restarts weren't working because the `chunk_iter` here was being closed, when only the response needed to be closed between restarts. I tested this manually locally.

### DIFF
--- a/pipeline/common/downloads.py
+++ b/pipeline/common/downloads.py
@@ -299,10 +299,15 @@ class DownloadChunkStreamer(io.IOBase):
                 # the retries are done.
                 logger.error(f"A download error occurred: {error}")
 
-            self.close()
+            # Close out the response on an error. It will be recreated when retrying.
+            if self.response:
+                self.response.close()
+                self.response = None
+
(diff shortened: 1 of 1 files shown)


pipeline/common/downloads.py
```diff
@@ -299,10 +299,15 @@ def download_chunks(self) -> Generator[bytes, None, None]:
                 # the retries are done.
                 logger.error(f"A download error occurred: {error}")
```
gregtatum, line 301: This is the codepath that runs when a response times out, or a network connection is lost.

eu9ene (approved):

pipeline/common/downloads.py
eu9ene: I'm not sure if "with" can be applied here to ensure the resources will be closed in any case

gregtatum: I have been playing around with `@contextmanager`(I think you pointed that out) and `ExitStack`. This could would probably be simpler with that.

Issue #691, opened 2024-06-20
ERROR - `task.dependencies` references tasks that expires before `task.deadline`


eu9ene:
It seems the issue with the deadline is even more serious. My train action with caching fails now: 

https://firefox-ci-tc.services.mozilla.com/tasks/DQkkrHmNQP6dSjg5p_YYfQ/runs/0

I'm running it from this PR push task: https://github.com/mozilla/firefox-translations-training/pull/690
With this config, which is a regular way to use the cached tasks:

```
# The initial configuration was generated using:
# task config-generator -- en uk --name spring-2024
#
# The documentation for this config can be found here:
# https://github.com/mozilla/firefox-translations-training/blob/001d15d5f2775e0e4c57717057a1906069e29dcc/taskcluster/configs/config.prod.yml
experiment:
  name: spring-2024
  src: en
  trg: uk
  best-model: chrf
  use-opuscleaner: 'true'
  opuscleaner-mode: defaults
  bicleaner:
    default-threshold: 0.5
    dataset-thresholds: {}
  mono-max-sentences-src: 500_000_000
  mono-max-sentences-trg: 200_000_000
  spm-sample-size: 10_000_000
  spm-vocab-size: 32000
  teacher-ensemble: 2
  teacher-mode: two-stage
  pretrained-models: {}
datasets:

  # Skipped test/devtest datasets:
  devtest:
  - flores_aug-mix_dev
  - mtdata_aug-mix_Neulab-tedtalks_dev-1-eng-ukr
  test:
  - flores_devtest
  - flores_aug-mix_devtest
  - flores_aug-title_devtest
  - flores_aug-upper_devtest
  - flores_aug-typos_devtest
  - flores_aug-noise_devtest
  - flores_aug-inline-noise_devtest
  - mtdata_Neulab-tedtalks_test-1-eng-ukr

  # The training data contains:
  #   58,968,083 sentences
  #
  # Skipped datasets:
  #  - opus_CCMatrix/v1 - ignored datasets (20,240,171 sentences)
  #  - opus_MultiMaCoCu/v2 - ignored datasets (6,406,288 sentences)
  #  - opus_GNOME/v1 - not enough data  (150 sentences)
  #  - opus_Ubuntu/v14.10 - not enough data  (0 sentences)
  #  - mtdata_ELRC-wikipedia_health-1-eng-ukr - duplicate with opus
  #  - mtdata_Facebook-wikimatrix-1-eng-ukr - duplicate with opus
[...]

PR #798, opened 2024-08-09
Add an action to rebuild pipeline toolchains and docker images

gabrielBusta:
We can add an action to rebuild the toolchains. Tangentially related to #691


taskcluster/translations_taskgraph/actions/rebuild_toolchains.py
```diff
@@ -0,0 +1,30 @@
...
+    symbol="rebuild-toolchains",
+    description="Create toolchain tasks to rebuild their artifacts.",
+    order=1000,
+    context=[],
+)
+def rebuild_toolchains_action(parameters, graph_config, input, task_group_id, task_id):
+    decision_task_id, full_task_graph, label_to_taskid = fetch_graph_and_labels(
+        parameters, graph_config, task_group_id=task_group_id
+    )
+    toolchains = [label for label, task in full_task_graph.tasks.items() if task.kind == "toolchain"]
```
bhearsum, line 21: I think we should rename this action and include docker images and fetches here as well, as they'll cause all the same problems as toolchains do.

bhearsum (changes requested):
Great idea! We should include a couple of more things here though.

taskcluster/translations_taskgraph/actions/rebuild_toolchains.py
gabrielBusta: Mm, any ideas on names? We can create those kinds too.

bhearsum: How about "Rebuild docker images and toolchains". (I don't think "fetch" needs to be included in the names, seeing as they just feed toolchains.)

bhearsum (approved):
This looks fine; can you make sure you test it explicitly either here or in staging before this merges, please?

eu9ene (approved):
Looks good, but I suggest to add a comment why it's needed and when to use it. Ideally even document it in the docs.

bhearsum:
I'll address those comments and drive this home.

eu9ene (approved):
Thanks for the docs!

PR #812, opened 2024-08-27
Add a train task

gregtatum:
This somewhat automates the process of starting training. I'd like to get the Taskcluster piece kicking off from the command line but I'm struggling a bit with getting the scope permissions working. This at least opens up the proper webpages making it so you don't have to navigate the GitHub->Taskcluster labyrinth.


utils/train.py
```diff
@@ -0,0 +1,142 @@
...
+
+    if branch:
+        print(f"Using --branch: {branch}")
+    else:
+        branch = run(["git", "branch", "--show-current"])
+        print(f"Using current branch: {branch}")
+
+    if check_if_pushed(branch):
+        print(f"Branch '{branch}' is up to date with origin.")
+    elif args.force:
```
eu9ene, line 94: what's the use case of doing this without a push?

utils/train.py
```diff
@@ -0,0 +1,142 @@
...
+
+    task_group_id = get_task_group_id(decision_task.details_url)
+
+    config: Path = args.config
+
+    url = f"https://firefox-ci-tc.services.mozilla.com/tasks/groups/{task_group_id}"
+    print(f"Opening {url}")
+
+    with config.open("rt", encoding="utf-8") as file:
+        print('The config has been copied to your clipboard. Select "Train" from the actions.')
```
eu9ene, line 135: Nice but I agree, it would be even better to just get a link to the running group or the train action.

eu9ene (approved):

utils/train.py
gregtatum: I have a patch for triggering the training from the command line but I need some help from Ben or Gabriel for debugging the scopes on it.

utils/train.py
gregtatum: I've done it a few times now on experiments where I am working locally and have some changes, but I want to trigger on the remote.

PR #813, opened 2024-08-27
Release cherry picks 2

eu9ene:
Cherry picks of the latest commits from release + relock poetry file after conflicts.


gregtatum (approved):
Thanks for handling this chore! It'll be nice to have these available on main.

PR #817, opened 2024-08-29
Publish experiments to W&B from the CI

### LINKED ISSUE #792: Add publishing to CI
Since experiment tracking is now tightly integrated with training and evaluation scripts we should make sure refactorings don't break compatibility with the tracking module.

Let's enable publishing in CI. To not pollute the current default project in CI config (ru-en), let's somehow detect that it's CI from the tracking module and publish it to a separate project (named `ci`).

vrigal:
Closes #792

### DIFF
--- a/taskcluster/translations_taskgraph/parameters.py
+++ b/taskcluster/translations_taskgraph/parameters.py
@@ -4,6 +4,7 @@
 
 from taskgraph.parameters import extend_parameters_schema
 from voluptuous import Extra, Optional, Required
+import uuid
 
 
 # These defaults line up with the `config.ci.yml` pipeline as much as possible.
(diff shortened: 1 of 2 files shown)


vrigal:
@eu9ene I found no simple way to disable cache in Taskcluster (usually I was updating `training_config.marian-args.training-backward.disp-freq` to trigger new jobs). Maybe it could be disabled directly from taskcluster ?

Otherwise any new pushes will trigger a CI job without training nor eval step, making it impossible to visualize on W&B.

eu9ene:
> @eu9ene I found no simple way to disable cache in Taskcluster (usually I was updating `training_config.marian-args.training-backward.disp-freq` to trigger new jobs). Maybe it could be disabled directly from taskcluster ?
> 
> Otherwise any new pushes will trigger a CI job without training nor eval step, making it impossible to visualize on W&B.

`tracking` is not included in cache digest by design since we don't want to rerun the core pipeline when tracking code changes. I think it's fine for now. It will be triggered by all the PRs that touch the training scripts and we can trigger it manually when we change `tracking`. We can ask @bhearsum whether it's possible to hack it and handle the latter case as well.

eu9ene (approved):

PR #821, opened 2024-09-03
Add a memory logger

gregtatum:
I found this really useful for managing OOM issues in the merge mono task by measuring the memory while running things.

[Example live log](https://firefox-ci-tc.services.mozilla.com/tasks/RnqpdpZRQL6vIbo5mjmKwA/runs/0/logs/public/logs/live.log).

This PR is blocking submitting the HPLT importer, as it's using the memory logger as well.


pipeline/common/memory.py
```diff
@@ -0,0 +1,52 @@
...
+    Get the current string representation of the memory usage.
+    """
+    global _memory_process
+    global _memory_last_bytes
+
+    # Lazily initial everything.
+    if not _memory_process:
+        _memory_process = psutil.Process(os.getpid())
+
+    gc.collect()
```
eu9ene, line 27: Generally triggering garbage collection manually is not a good idea (it interferes with the default efficient behavior of the interpreter and it's not free, so the program will likely become slower).  `log_memory` is a helper function and it has a severe side effect because of this. It is not expected by the user and will slow things down in the client code. I suggest either removing it and calling it manually in places you need it or adding a flag with disabled collection by default, like`gc_collect=False`. If you feel like it's safe to do it from the place in code where you call `log_memory`, pass `gc_collect=True`.

eu9ene (changes requested):

eu9ene:
Nice idea! I agree it can be more convenient to look at memory allocation in the logs, but we also have the GCP dashboard where you can filter things by the machine ID. For example: `translations-1-b-linux-large-gcp-300gb-gd4dp9-ht0mwubpmall1yq` for your task.

```
[taskcluster 2024-08-30 00:14:09.606Z] Hostname: translations-1-b-linux-large-gcp-300gb-gd4dp9-ht0mwubpmall1yq
```

gregtatum:
> Nice idea! I agree it can be more convenient to look at memory allocation in the logs.

I also like it because I can see it locally when I'm developing the script to ensure I'm not screwing things up with my mental model of what's being loaded in memory. It was quite handy on getting some of my scripts working on production levels of data.

eu9ene (approved):

PR #824, opened 2024-09-03
Remove the compression command configuration, and only use zstandard

### LINKED ISSUE #430: Fully switch to zstd compression
We have decided to reduce complexity of the pipeline by supporting only one and the most efficient compression algorithm - [zstd](https://github.com/facebook/zstd). We should remove all the gzip/pigz related code from the core pipeline scripts and taskcluster kinds ( `COMPRESSION_CMD`, `COMPRESSION_EXT` env variables, --compression_cmd cli args etc.)

gregtatum:
Resolves #430


gregtatum:
Ah, I missed a commit that fixes CI here.

pipeline/common/downloads.py
```diff
@@ -471,3 +471,81 @@ def write_lines(path: Path | str):
 
     finally:
         stack.close()
+
+
+def compress_file(
```
eu9ene, line 476: Should we add unit tests for compress/decompress functions? They are quite important to work correctly now.

eu9ene (commented):
Nice refactoring that removes a lot of unnecessary complexity from the code! I suggest adding unit tests for the new compression/decompression functions before merging.

pipeline/common/downloads.py
gregtatum: Good idea. I'll do it. I think we're covered by integration tests, but it's always nice to be explicit on stuff like this.

eu9ene (approved):
lgtm, but I think @bhearsum should have a look too, since a lot of Taskcluster code is affected

bhearsum (approved):
This is great! Thanks for doing it. The taskcluster stuff looks very safe; the pipeline run here seems to confirm that there's no issues.

Issue #455, opened 2024-02-16
Adjust OpusTrainer setting in CI


eu9ene:
OpuTrainer buffers too much data to prepare for real training, but our dataset for CI is tiny. This leads to reading the dataset many times and a very long preprocessing. We should adjust settings like `--chunk-size` `--batch-size` and `--workers`. 

Example student task from CI:
https://firefox-ci-tc.services.mozilla.com/tasks/AfSTqkqEQ5-vtmxJjkS3ZA/runs/0/logs/public/logs/live.log

Full help:

 
```
usage: opustrainer-train [-h] --config CONFIG [--state STATE] [--sync] [--temporary-directory TEMPORARY_DIRECTORY] [--do-not-resume] [--no-shuffle]
                         [--batch-size BATCH_SIZE] [--chunk-size CHUNK_SIZE] [--workers WORKERS] [--log-level LOG_LEVEL] [--log-file LOG_FILE]
                         ...

Feeds marian tsv data for training.

positional arguments:
  trainer               Trainer program that gets fed the input. If empty it is read from config.

optional arguments:
  -h, --help            show this help message and exit
  --config CONFIG, -c CONFIG
                        YML configuration input.
  --state STATE, -s STATE
                        YML state file, defaults to ${CONFIG}.state.
  --sync                Do not shuffle async
  --temporary-directory TEMPORARY_DIRECTORY, -T TEMPORARY_DIRECTORY
                        Temporary dir, used for shuffling and tracking state
  --do-not-resume, -d   Do not resume from the previous training state
  --no-shuffle, -n      Do not shuffle, for debugging
  --batch-size BATCH_SIZE, -b BATCH_SIZE
                        Batch size
  --chunk-size CHUNK_SIZE, -B CHUNK_SIZE
                        Chunk size of batches fed to modifiers
  --workers WORKERS, -j WORKERS
                        Number of workers
  --log-level LOG_LEVEL
                        Set log level. Available levels: DEBUG, INFO, WARNING, ERROR, CRITICAL. Default is INFO
  --log-file LOG_FILE, -l LOG_FILE
                        Target location for logging. Always logs to stderr and optionally to a file.

```

PR #833, opened 2024-09-04
Make the CI model training even slimmer

gregtatum:
I'm looking a bit into CI speeds, and I wanted to try to slim down the model even more.

### DIFF
--- a/taskcluster/configs/config.ci.yml
+++ b/taskcluster/configs/config.ci.yml
@@ -48,7 +48,6 @@ marian-args:
     transformer-dim-aan: "256"
     transformer-heads: "2"
     beam-size: "2"
-    precision: float16
   training-teacher:
     disp-freq: "2"
     save-freq: "10"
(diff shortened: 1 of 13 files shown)


taskcluster/configs/config.ci.yml
```diff
@@ -29,36 +32,79 @@ experiment:
 
 marian-args:
   training-backward:
-    disp-freq: "1"
-    save-freq: "5"
+    # Run training for 10 updates, and display 5 updates. Only validate and save the
+    # model once.
+    disp-freq: "2"
```
eu9ene, line 37: @gregtatum I don't think it's a good idea to touch model architecture too much here since those parameters are also something we want to test in CI. Student and backward models are already small. I suggest we do only minor changes here. You can try using Marian `task: transformer-base` instead of `task: transformer-big` but the main delay appeared after integration of OpusTrainer and it's likely this issue: https://github.com/mozilla/firefox-translations-training/issues/455

PR #836, opened 2024-09-05
Splitter compression

gregtatum:
I split this out from my #824 PR, as that one was already quite large, and it was somewhat adjacent.

Here is a production run:

https://firefox-ci-tc.services.mozilla.com/tasks/Lc41w4glQYi7TKVki9c4_w/runs/0/logs/public/logs/live.log

This simplifies the implementation a bit, and adds a test that splitting is done correctly.


bhearsum (approved):

tests/test_split_corpus.py
```diff
@@ -0,0 +1,57 @@
...
+from fixtures import DataDir
+
+from pipeline.common.downloads import read_lines
+
+# With 47 lines, there will be 5 lines per file, except the last file which should have 2.
+corpus_line_count = 47
+
+
+@pytest.mark.parametrize("task", ["src-en", "trg-ru"])
+def test_split_mono(task: str):
```
eu9ene, line 13: It's probably still useful but FYI there's also more high level https://github.com/mozilla/firefox-translations-training/blob/main/tests/test_split_collect.py

eu9ene (approved):

PR #842, opened 2024-09-10
Rewrite train.sh to train.py

### LINKED ISSUE #599: Rewrite the training script in Python
We started adding more and more features to the training procedure which increases complexity and it's time to switch to Python. This will help us to use named arguments, better handle training config templating and composing and simplify providing extra training options like "one-stage" teacher training mode.

Related to #579

### LINKED ISSUE #617: Switch OpusTrainer log level to info
I switched it to ERROR in https://github.com/mozilla/firefox-translations-training/pull/571 but ideally, we want to filter empty alignments before feeding the TSV to OpusTrainer. We can do this after #599

gregtatum:
Resolves #599, although it doesn't touch the taskcluster bits yet.

Resolves #617 as well.


pipeline/train/train.py
```diff
@@ -0,0 +1,413 @@
...
+
+            aln_lines: Generator[str, Any, Any] = stack.enter_context(
+                read_lines(f"{alignments_file}")
+            )
+            empty_alignments = []
+
+            for src_line, trg_line, aln_line in zip(src_lines, trg_lines, aln_lines):
+                if not aln_line:
+                    empty_alignments.append((src_line, trg_line))
+                    continue
```
gregtatum, line 87: This is a behavior change from the original script.

bhearsum (approved):
This looks fine from a taskcluster standpoint.

> Resolves https://github.com/mozilla/firefox-translations-training/issues/599, although it doesn't touch the taskcluster bits yet.

Does this refer to integrating `train-taskcluster.sh` with `train.py` as well? (That seems like it would be a fantastic thing to do.)

pipeline/train/train.py
```diff
@@ -0,0 +1,413 @@
...
+        return ["cat"]
+
+    logger.info("Weight & Biases publication is available.")
+    return ["parse_tc_logs", "--from-stream", "--publish-group-logs", "-v"]
+
+
+class TrainCLI:
+    def __init__(self, args: Any, temp_dir: Path) -> None:
+        self.temp_dir = temp_dir
+        self.vocab: Path = args.vocab
```
eu9ene, line 130: It might be worth moving the args to a separate data class and do arg parsing there. It will be easier to work with, log and extend (optional suggestion)

pipeline/train/train.py
```diff
@@ -0,0 +1,413 @@
...
+        self,
+    ):
+        """
+        Generate an OpusTraininer config that points to the current datasets and language
+        options.
+        """
+        if self.model_type == ModelType.teacher:
+            teacher_mode = self.teacher_mode.value
+            if teacher_mode == TeacherMode.none.value:
+                logger.error("Teacher mode was not properly set")
```
eu9ene, line 235: it should be `raise ValueError(...` if it's expected, otherwise there should be a default value in CLI args.

pipeline/train/train.py
```diff
@@ -0,0 +1,413 @@
...
+                teacher_mode = TeacherMode.two_stage.value
+
+            config_input = (
+                train_dir / f"configs/opustrainer/{self.model_type.value}.{teacher_mode}.yml"
+            )
+        else:
+            config_input = train_dir / f"configs/opustrainer/{self.model_type.value}.yml"
+
+        with open(config_input, "rt", encoding="utf-8") as file:
```
eu9ene, line 244: FYI just `open(config_input)` does the same thing

pipeline/train/train.py
```diff
@@ -0,0 +1,413 @@
...
+            file.write(config_text)
+
+    def get_opustrainer_cmd(self):
+        return [
+            "opustrainer-train",
+            *apply_command_args(
+                {
+                    "config": str(self.opustrainer_config),
+                    "log-file": str(self.artifacts / "opustrainer.log"),
+                    "log-level": "ERROR",
```
eu9ene, line 262: Log level should be switched to info if we make sure that there are no empty alignments. See https://github.com/mozilla/firefox-translations-training/issues/617

pipeline/train/train.py
```diff
@@ -0,0 +1,413 @@
...
+        help="Type of teacher training",
+    )
+    parser.add_argument("--gpus", type=str, required=True, help="GPUs to use")
+    parser.add_argument(
+        "--marian_dir",
+        type=Path,
+        required=True,
+        help="Path to Marian binary directory. This allows for overriding to use the browser-mt fork.",
+    )
+    parser.add_argument("--workspace", type=str, required=True, help="Workspace directory")
```
eu9ene, line 373: This is the amount of preallocated memory in Marian and not a directory

pipeline/train/train.py
```diff
@@ -0,0 +1,413 @@
...
+        required=True,
+        help="The type of model to train",
+    )
+    parser.add_argument(
+        "--training_type",
+        type=TrainingType,
+        choices=TrainingType,
+        help="Type of teacher training",
+    )
+    parser.add_argument("--gpus", type=str, required=True, help="GPUs to use")
```
eu9ene, line 366: It's not clear what format it expects

pipeline/train/train.py
```diff
@@ -0,0 +1,413 @@
...
+    parser.add_argument("--trg", type=str, help="Target language")
+    parser.add_argument(
+        "--train_set_prefixes",
+        type=str,
+        help="Comma separated prefixes to datasets for curriculum learning",
+    )
+    parser.add_argument("--validation_set_prefix", type=str, help="Prefix to validation dataset")
+    parser.add_argument("--artifacts", type=Path, help="Where to save the model artifacts")
+    parser.add_argument("--vocab", type=Path, help="Path to vocab file")
+    parser.add_argument("--best_model_metric", type=str, help="Best model metric")
```
eu9ene, line 384: it's in fact one of expected values, so it should be an enum of supported metrics

tracking/translations_parser/cli/taskcluster.py
```diff
@@ -82,6 +82,24 @@ def get_args() -> argparse.Namespace:
     return parser.parse_args()
 
 
+def is_running_in_ci():
```
eu9ene, line 85: @bhearsum you probably discussed this with Greg already but I'm curious if it's the only way to determine that we're in ci

tracking/translations_parser/parser.py
```diff
@@ -322,8 +326,12 @@ def parse_marian_context(self, logs_iter: Iterator[tuple[list[tuple[str]], str]]
...
         # Consume first lines until we get the Marian header
         while ("marian",) not in headers:
-            headers, text = next(logs_iter)
-            logger.debug(f"Marian header not found in: headers={headers} text={text.strip()}")
+            try:
+                headers, text = next(logs_iter)
+                logger.debug(f"Marian header not found in: headers={headers} text={text.strip()}")
+            except StopIteration as exception:
+                logger.debug("Could not find a [marian] entry in the training log.")
+                raise exception
```
eu9ene, line 334: maybe worth raising something else instead of StopIteration that can be confusing. Like `raise ValueError("Could not find a [marian] entry in the training log.")`

eu9ene (changes requested):
Great work @gregtatum! We wanted to do this for a long time :) It looks good overall, but I have a bunch of minor suggestions to address.

pipeline/common/processes.py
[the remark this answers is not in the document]
gregtatum: I renamed it to `command_runner`.

pipeline/train/train.py
gregtatum: Ah, I was going to follow-up with that, but happy to do it here.

pipeline/train/train.py
gregtatum: I looked at doing this, but my original intent was for the TrainCLI to be the dataclass. Since this is optional, I'll pass on it for now.

gregtatum:
I rebased, but only [73d9122](https://github.com/mozilla/firefox-translations-training/pull/842/commits/73d912235f2f75964664e1108f41ebf11c149225) and [d5a2518](https://github.com/mozilla/firefox-translations-training/pull/842/commits/d5a2518f8629acb499190e9ca79b63f139b02785) are new, the rest are the old commits.

eu9ene (approved):

PR #847, opened 2024-09-13
Speed up CI by changing model params

gregtatum:


gregtatum:
* Baseline run: 2:47 - https://share.firefox.dev/4gBCi8U
* Smaller model: 2:38 - https://share.firefox.dev/3MUEsTj

With as long as the runs are, this doesn't show great improvements in terms of speed. The only real affect will be in storage costs.

PR #851, opened 2024-09-17
Merge corpus rewrite to python

gregtatum:
This is ready for review now.

Here is an example CI run of the task.

* [merge-corpus](https://firefox-ci-tc.services.mozilla.com/tasks/Mz4fdN-oQ92WHOQ9fz5FiQ)
* [merge-devset](https://firefox-ci-tc.services.mozilla.com/tasks/NwVQoRvcQMOtUH6ruipdCA)


taskcluster/configs/config.ci.yml
```diff
@@ -10,11 +10,11 @@ experiment:
   teacher-ensemble: 1
 
   mono-max-sentences-src:
-    total: 10000
```
gregtatum, line 13: FYI, This is a behavior change for CI, but I did it to match the truncation for the corpus size.

bhearsum (approved):
Thank you for the cleanup in the merge kinds.

tests/test_merge_corpus.py
```diff
@@ -0,0 +1,219 @@
...
+
+        assert corpus_lines_sorted == sorted_lines
+        assert corpus_lines != corpus_lines_sorted, "The results are shuffled."
+
+
+@pytest.mark.parametrize(
+    "name",
+    ["corpus", "devset"],
+)
+def test_merge_corpus(data_dir, name):
```
eu9ene, line 70: I see this is a comprehensive full test. If something breaks it will be hard to say what exactly. Usually, the best practice is to split tests to use cases. For example, it would make sense to test deduplication separately as it's an important part of functionality. I assume it might be written like this to make it faster due to the slowness of run_task. In this case, a direct test for the class can help. Up to you though if you want to split/add it (likely you don't).

eu9ene (changes requested):
Nice, I see you added a bunch of new functionality here: 
- deduplication statistics
- corpus sampling
- in memory deduplication instead of using the dedupe tool (I guess it's easier to use with shuffle_with_max_lines?)

All those are useful things for sure!

Great work! I understand the reasons and that this follows our strategy of rewriting in Python. However, I'd like to point out that replacing 40 lines of bash with 300+ lines of Python is not free and our maintenance burden has been growing rapidly lately due to the rapidly increasing code base including rewriting external tools. Hopefully having more tests can balance this as the code is a lot easier to debug now. Just something to be aware of. The best code is no code :)

I have a minor suggestion but I'd rather have it fixed as it's super simple.

tests/test_merge_corpus.py
gregtatum: > I see this is a comprehensive full test

My general preference in tests is to test from the top down, so `run_task` is the layer I try to test at. The only time I like to go down to the more unit test layer is if I have specific algorithms that I'm concerned about. For instance I agree with your previous feedback on the common pipeline tools having more unit tests, as those are core utilities that if they broke could create mysterious behavior.

> slowness of run_task

The `run_tasks` aren't really that slow on their own. They are slow for the initial taskgraph generation (which `task test-fast` works around) and they are slow for the virtual env creation. Which again, only needs to happen once per run. The first `run_task` is slow, but every subsequent one is fast to run.

> a direct test for the class can help

I'd really prefer to test at the run_task level, as that's really the API that's being exposed to be exercised. The internals are more implementation details. I feel like the test I wrote gets really good coverage over the desired behavior, and should and represents a minimal setup. It should be close to 100% code coverage.

Can you give some examples of tests directly using the classes that would be helpful to have?

eu9ene: The DeduplicateCorpus class looks slightly more complex so I would write separate tests for it to cover deduplication test cases. If you feel like your current high-level test is sufficient I'm fine with that.

eu9ene (approved):
Looks good but CI is failing now.

PR #852, opened 2024-09-19
Add environment variables to all artifacts and fetches

gregtatum:
I keep on adding these when adding a new `run_task` test, so it would be nice to just have this corrected everywhere.


bhearsum (approved):
Thank you!

gregtatum:
I'm investigating the CI error.

bhearsum:
> I'm investigating the CI error.

I just reran the task to see if it's intermittent, FYI. (404s shouldn't be, but I've definitely seen similar things in the past.)

gregtatum:
I missed an artifacts folder in the path.

eu9ene (approved):

PR #856, opened 2024-09-20
Use the config.ci.yml for the training defaults

gregtatum:
This refactor makes it so that the `get_defaults` function pulls from the `config.ci.yml` file only. I modified the `config.ci.yml` file to have the exact same values as were present in the python. There was some variance that had built up over time.


bhearsum (approved):
This is excellent! Thank you!

PR #859, opened 2024-09-23
fix: pass 'permission' to train action

bhearsum:
This is required for non-generic actions as of 11.0. I mucked this up when I did the upgrade. (See https://taskcluster-taskgraph.readthedocs.io/en/latest/reference/migrations.html.)

### DIFF
--- a/taskcluster/translations_taskgraph/actions/train.py
+++ b/taskcluster/translations_taskgraph/actions/train.py
@@ -64,6 +64,7 @@ def validate_pretrained_models(params):
     symbol="train",
     description="Initiate part or all of the training pipeline",
     cb_name="train",
+    permission="train",
     order=500,
     context=[],
     available=can_train,


bhearsum:
@ahal - can you advise here? The migration guide says:
> If the argument isn’t present, or you are passing generic=False, then add a new argument called permission=<cb_name>, where <cb_name> is the value of whatever you are passing to the cb_name argument.

But we're getting:
```
[task 2024-09-23T17:54:31.899Z] E           Exception: Unknown callback: train. Known callbacks: rebuild-cached-tasks, retrigger-decision, retrigger-disabled, retrigger, rerun, retrigger-multiple, cancel, cancel-all, add-new-jobs
```

(Everything seems to work fine without this patch, eg: in https://github.com/mozilla/firefox-translations-training/pull/858 I managed to use the `train` action.)

bhearsum:
This ended up getting done in #953

PR #865, opened 2024-09-25
Output all of test spew

gregtatum:
This will let me get a profile of it.


gregtatum:
https://share.firefox.dev/3BcZZE8

PR #871, opened 2024-10-02
Add Inference Tasks to CI

nordzilla:
This pull request makes a few modifications to the taskcluster configurations for Docker. Each commit is separated into a logical, self-contained modification.

* Ensures that the `IS_DOCKER` environment variable is exported to `1` in all CI docker images.
* Modifies taskcluster naming conventions previously associated with `all-pr` to now be `all-pr-pipeline`.
* Introduces a new CI docker image, kind, and tasks for `inference`. 

New tasks added to the CI Pipeline:

* `@firefoxci-taskcluster` **inference-build** Successful in 8m — FirefoxCI (pull_request)
* `@firefoxci-taskcluster` **inference-build-wasm** Successful in 13m — FirefoxCI (pull_request)
* `@firefoxci-taskcluster` **inference-test Successful** in 10m — FirefoxCI (pull_request)

---

Note: 

This is my first time working with taskcluster, taskgraph, etc. I tried to follow the existing code as an example, but I'm happy to make any further modifications based on best-practice recommendations. 

The full pipeline training run has to run on this PR because I've modified the base Dockerfile, but ideally in the future only the inference-related tasks will run on inference-related PRs.


taskcluster/kinds/all-pr-inference/kind.yml
```diff
@@ -0,0 +1,36 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
bhearsum, line 1: You can add this if you like, but it's probably not necessary seeing as you only have one kind. Instead, I would suggest you add `run-on-tasks-for` to the `task-defaults` for the `inference` kind, which will have the same effect of causing these tasks to run in PRs without the dummy task.

You may also want to consider putting `github-push` in the `run-on-tasks-for` list so that the tests run on pushes as well.

The reason we have `all` and `all-pr` for the pipeline tasks is mainly to give us a single task to target that means "run the whole pipeline" (otherwise we'd need to specify `export` and a bunch of `evaluate` steps as `target-phase` whenever we kick things off).

taskcluster/kinds/all-pr-pipeline/kind.yml
```diff
@@ -23,13 +23,13 @@ kind-dependencies:
     - analyze-mono
 
 tasks:
-    all-pr:
+    all-pr-pipeline:
```
bhearsum, line 26: I think it's fine to keep this change either way, seeing as it does not in fact cause _all_ tasks to run even now (just pipeline ones, as this is making clear!). It's probably worth amending the `all` kind with the same suffix if you keep this though.

taskcluster/kinds/inference/kind.yml
```diff
@@ -0,0 +1,66 @@
...
+      docker-image: {in-tree: inference}
+    run:
+      command:
+        - bash
+        - -c
+        - >-
+            task inference-build
+
+  build-wasm:
+    worker-type: b-cpu
```
bhearsum, line 45: If you want, you could move `worker-type` and `worker` up to `task-defaults` seeing as they're the same in all `tasks` entries.

bhearsum (approved):
Nice work on this! I recommend that you drop `all-pr-inference`, but there's no harm in keeping it if you prefer.

taskcluster/kinds/all-pr-inference/kind.yml
nordzilla: I've removed this file, leaving only the `inference` kind that has `run-on-tasks-for` defined.

taskcluster/kinds/inference/kind.yml
nordzilla: I've refactored these configurations into the `task-defaults` section.

taskcluster/kinds/all-pr-pipeline/kind.yml
nordzilla: I renamed the `all*` task names to `all*-pipeline`.

nordzilla:
@bhearsum 

Just giving you another quick tag now that I've responded to your feedback.

taskcluster/kinds/inference/kind.yml
```diff
@@ -0,0 +1,76 @@
...
+    run:
+      command:
+        - bash
+        - -c
+        - >-
+            task inference-build
+
+  test-local:
+    description: "Run local-build tests for the inference engine"
+    worker-type: b-cpu
```
bhearsum, line 48: It looks like you dropped `worker-type` and some other things from `build-local` but not `test-local` nor `build-wasm`. (I assume this was an oversight!)

bhearsum (approved):

taskcluster/kinds/inference/kind.yml
nordzilla: Yes, just an oversight, thanks for catching that! I fixed it now.

gregtatum (approved):
It all seems reasonable to me!

Issue #174, opened 2023-08-30
Investigate using larger student models


marco-c:
Current models are around ~20 MB, this is very little for Desktop. We could try to go with larger models and see what kind of quality improvements we manage to get.

eu9ene:
I will try training the student Large or Base configurations from https://aclanthology.org/D19-5632.pdf. 


<img width="1049" alt="Screenshot 2024-10-10 at 11 58 31 AM" src="https://github.com/user-attachments/assets/c0da09e9-e1a8-4714-a932-5d911fb65c0a">

eu9ene:
I looked at the recommended configurations in https://github.com/browsermt/students/tree/master/train-student/models and what HPLT folks are training here https://github.com/hplt-project/bitextor-mt-models/tree/main and it seems the recommended approach to train a larger model is to go with the Base configuration. The difference I see in code is basically these two model parameters:

base
```yaml
    dim-emb: 512
    transformer-dim-ffn: 2048
```
tiny
```yaml
    dim-emb: 256
    transformer-dim-ffn: 1536
```

eu9ene:
It can give ~2 BLEU points bump and is twice slower, for example here https://github.com/hplt-project/bitextor-mt-models/tree/main/swa-eng.

I launched training for en-ru student in Base configuration: 
https://firefox-ci-tc.services.mozilla.com/tasks/P8eqvTKvRp640O8mcMZtWw/runs/0
https://wandb.ai/moz-translations/en-ru/runs/mjnbhst5?nw=nwuserepavlov

eu9ene:
It looks pretty good so far! The gap is significantly smaller on the validation set now.

<img width="1352" alt="Screenshot 2024-10-11 at 9 34 44 AM" src="https://github.com/user-attachments/assets/a7a5a2c0-95bc-4135-a34e-6292db2e305e">

gregtatum:
For clarity here, since I was confused for a bit, the `tiny` and `base` don't quite match up with the paper, as the paper's `base` model also increases the decoder depth. So here the terms have slightly different meanings.

gregtatum:
(I updated the [summary comment](https://github.com/mozilla/firefox-translations-training/issues/174#issuecomment-2414864806) for some clarity on the result)

PR #881, opened 2024-10-15
Add student base configuration option

eu9ene:
I'm adding student Base configuration as the core pipeline feature based on a successful experiment in #174.

I guess we'll be training more of such models soon. 

We should wait for fine-tuning and quantization to finish to make sure there are no issues there. [Task group](https://firefox-ci-tc.services.mozilla.com/tasks/groups/AYqN3ysXRp2EGkEqeaA5Rg)


taskcluster/kinds/finetune-student/kind.yml
```diff
@@ -47,7 +48,8 @@ tasks:
             cache:
                 type: finetune-student
                 resources:
-                    - pipeline/train/configs/model/student.yml
+                    - pipeline/train/configs/model/student.tiny.yml
+                    - pipeline/train/configs/model/student.base.yml
```
bhearsum, line 52: If you'd like to avoid rebuilding `tiny` tasks when only `base` changes you can avoid hardcoding both of these and using something like:
```
resources:
    - pipeline/train/configs/model/student.{student_model}.yml
```

And adding this to `task-context`:
```
task-context:
    from-parameters:
        student_model: training_config.experiment.student-model
```

...which should end up with just the actively used `student-model` yml file in the cache digest.

I know we're largely relying on `previous_group_ids` and `existing_tasks` rather than cached tasks, but this might still be useful to do.

bhearsum (approved):
I encourage you to correct the cache resources per below, but it's ultimately up to you whether or not that's important.

docs/training-guide.md
```diff
@@ -146,11 +146,19 @@ for example [`teacher.train.yml`] and in the [`train.py`] script.
 
 ### Model training
 
+#### Student architecture
+
+"base" or "tiny" based on [Bergamot configurations](https://github.com/browsermt/students/tree/master/train-student/models).
```
gregtatum, line 151: Suggestion: It would be nice to provide additional context: https://aclanthology.org/D19-5632.pdf

And note that the names differ from the paper due to different decoder depths.

pipeline/train/configs/model/student.base.yml
```diff
@@ -0,0 +1,27 @@
+# https://github.com/browsermt/students/tree/master/train-student/models/student.base
```
gregtatum, line 1: Suggestion: Either duplicate some of the notes about the "Ludicrously Fast" paper here, or link to the `.md` docs, e.g. `"For more details see: [link]"`

pipeline/train/configs/model/student.base.yml
```diff
@@ -0,0 +1,27 @@
...
+# the difference with the "tiny" configuration is the dimensionality of transformer-dim-ffn and dim-emb
```
gregtatum, line 2: Suggestion: Is there a way to link here in the code (or in the docs) to your results? This could be a github PR or issue or w&b link. This will help future contributors to have context as to why things exist the way they do.

gregtatum (approved):
Changes look good to me with the CI passing. If the CI requires significant changes please re-flag me for review.

docs/training-guide.md
eu9ene: I'm actually not sure from which paper those configs were taken as they don't match the ones in this paper. So I'm referring to the student repo everywhere.

pipeline/train/configs/model/student.base.yml
eu9ene: It's a good idea to add a link to the experiment I did but there are also other ones that show the difference between base and tiny, for example, hplt docs. The experiment I did in fact doesn't matter here.  We're just reproducing the Bergamot recipe which provided two student configurations. We supported only one of them before. I guess I'll add a link to the issue if it helps to understand the context better.

PR #892, opened 2024-10-21
Disable bilceaner hard rules completely

### LINKED ISSUE #888: Disable use of `bicleaner-hardrules`
I think the most safe option, specially if you already have a previous rule-based cleaning step, is to disable `bicleaner-hardrules` with `--disable_hardrules` for all language pairs.
https://github.com/mozilla/firefox-translations-training/blob/c588cdfe3b397a02e411a0a5168ff720b7df4770/pipeline/bicleaner/bicleaner.sh#L53-L58
Even though hardrules are for detecting the most obious noise, there are some rules that may be too much for cleaning corpora coming from cleaner sources than web-crawls.

Right now, if not disabled, it is discarding all the sentences with less than 3 words. So I think #215 is totally hardrules's fault.

If you still want to keep hardrules, the most harmful options for non-Paracrawl corpora, can be disabled with: `--disable_minimal_length --disable_lm_filter`.

eu9ene:
closes #888

### DIFF
--- a/pipeline/bicleaner/bicleaner.sh
+++ b/pipeline/bicleaner/bicleaner.sh
@@ -49,12 +49,6 @@ else
     export scol=2
     export tcol=1
   fi
-  # disable hard rules for multilingual model
-  if [ ${model_source_lang} == "xx" ] || [ ${model_target_lang} == "xx" ]; then
-    export hardrules="--disable_hardrules"
-  else
-    export hardrules=""
-  fi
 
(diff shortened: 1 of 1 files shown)


gregtatum (approved):
Looks good to me.

Can you also make sure this is filed as an issue?

> I'm disabling it but we can compare the translation of short sentences for the languages where we used a custom model with enabled hard rules versus the languages where we used the multilingual model with disabled ones.

PR #900, opened 2024-10-23
Adjust data cleaning for CJK

### LINKED ISSUE #742: Support CJK in OpusCleaner
Nikolay:
Chinese alphabet should be added. In general we can use a unicode ranges to do so, but they are somewhat complicated: https://stackoverflow.com/questions/43418812/check-whether-a-string-contains-japanese-chinese-characters In the past i have used something like u'[\u4e00-\u9fff]', but this may be improved.

Length filtering. As Chinese sentences come normally as one continuous string of characters, traditional length filtering doesn't work. Furthermore, as one word can be made of 1-4 Chinese characters, we can't have some hard-and-fast conversion rule. What people normally do is they use a Chinese tokenizer (like jieba https://github.com/fxsjy/jieba#jieba-1 ) to split the Chinese text to words. We can then safely apply the filtering here:

https://github.com/mozilla/firefox-translations-training/blob/3b3f33bf2581238d325f05015123fc0a026c394e/pipeline/clean/tools/clean_parallel.py#L93

Most papers recommend to discard lines where the ratio of English to Chinese or Chinese to English words is more than 1.3

Afterwards the text should be de-segmented again and prepared for training

Japanese tokenizer should be used in place of jieba for japanese

eu9ene:
Use custom OpusCleaner configs with disabled word-based filters.

The filters are copied from https://github.com/hplt-project/HPLT-MT-Models/blob/main/v1.0/data/en-zh_hant/raw/v2/HPLT-v1.1.en-zh_hant.filters.json. 

I don't think it's feasible to do the src-trg-ratio that requires tokenization now. We would have to move tokenization to a separate step for that and somehow adjust the cleaning step to work with that instead of the original text. I filed https://github.com/mozilla/firefox-translations-training/issues/899

closes #742

### DIFF
--- a/pipeline/clean/opuscleaner/configs/en-ja/default.filters.json
+++ b/pipeline/clean/opuscleaner/configs/en-ja/default.filters.json
@@ -1,0 +1,77 @@
+{
+  "version": 1,
+  "files": [
+  ],
+  "filters": [
+    {
+      "filter": "remove_empty_lines",
+      "parameters": {},
+      "language": null
+    },
+    {
+      "filter": "normalize_whitespace",
+      "parameters": {
+        "COLLAPSE": true
+      },
+      "language": "<src>"
+    },
+    {
+      "filter": "deescape-special-chars",
+      "parameters": {
+        "LANG1": "other"
+      },
+      "language": "<src>"
+    },
+    {
+      "filter": "max_length",
+      "parameters": {
+        "MAXLENGTH": 150,
+        "MINLENGTH": 1
+      },
+      "language": null
+    },
+    {
+      "filter": "fix_wiki",
+      "parameters": {
+        "ALWAYS": false,
+        "FOOTNOTES": true,
+        "URLS": true,
+        "WIKILINKS": true,
+        "CODE": true,
+        "HEADINGS": true,
+        "REMOVEEMPTYLINES": true
+      },
+      "language": null
+    },
+    {
+      "filter": "alpha_ratio",
+      "parameters": {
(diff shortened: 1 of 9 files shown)


pipeline/clean/opuscleaner/configs/en-ja/default.filters.json
```diff
@@ -0,0 +1,77 @@
+{
```
gregtatum, line 1: Suggestion (optional): This is duplicating a lot of configs, and doesn't explain their provenance. Is there a way to share a single config between multiple languages?

gregtatum: Suggestion: If not, can you alternately add some documentation somewhere nearby, like `pipeline/clean/opuscleaner/README.md` that explains that they are duplicated and link to `https://github.com/hplt-project/HPLT-MT-Models/blob/main/v1.0/data/en-zh_hant/raw/v2/HPLT-v1.1.en-zh_hant.filters.json`

gregtatum (approved):

pipeline/clean/opuscleaner/configs/en-ja/default.filters.json
eu9ene: They are duplicated for now but potentially can be different if, say, we need a special rule for Japanese.  The en-xx and xx-en are already different. So I would leave it as is not to complicate other parts.

ZJaume (approved):

PR #901, opened 2024-10-23
Remove max_words filtering from data importers

### LINKED ISSUE #424: Monolingual data has a word splitter that won't work for CJK
Right now it splits on word boundaries, and limits the size of the monolingual data to be less than 100 "words". This needs to be changed to support another segmentation strategy for CJK languages, maybe just a byte limit.

eu9ene:
Any filtering should happen only in the cleaning stage (eventually in OpusCleaner). The max_words filtering on importing was originally a copy-paste from some random Bergamot bash script and was not needed at all. Even if we have some number of longer sentences that will later be cleaned, we can always compensate for that by adjusting max sentences in the config. We definitely don't want to deal with tokenization at this stage.

It's required for CJK.

closes #424


eu9ene:
@gregtatum I'm trying to figure out how to deal with the new HPLT importer. As we discussed before dataset modification should not be a default behavior. Also now we don't know the number of words at this stage. So, my options are:
1. Completely remove line accumulation and just write line by line as I would expect from a data importer. If we want to experiment with merging some document lines together, we can do it later. We'll need to think about this though and figure out what specific issues we want to address with that and what should be the output length distribution of the corpus.
2. Make the condition work based on the number of characters instead of the number of words and disable the accumulation by default.
Also, we'll need to migrate to HPLT 2.0 and I haven't looked at it yet, maybe the structure is different there. It should be done separately.
Thoughts?

gregtatum:
> As we discussed before dataset modification should not be a default behavior.

I don't know that this statement is correct. I remember in the review you bringing it up, but I don't think we have an agreement here.

The importer is more of a pre-processor, turning a document into one side of a "training pair". Training pairs can include single words, short phrases, sentences, or multiple sentences. Datasets are generally preprocessed by OPUS to be ready to be trained, and the HPLT documents are not preprocessed to be used as one side of a training pair. They are documents that are sentence split.

The statement "modification should not be a default behavior" is strange to me because CJK is transliterating at this same point in the pipeline, and the data augmentations are mutating the sentence pairs here.

> Completely remove line accumulation and just write line by line as I would expect from a data importer. If we want to experiment with merging some document lines together, we can do it later.

Merging requires access to the full document, as the merged sentences need to be colocated in the same document, and be concurrent in the sentence stream. This step couldn't be pushed to a later part of the pipeline unless a mapping for the document index, and line index was also saved to disk.

> Make the condition work based on the number of characters instead of the number of words and disable the accumulation by default.

My preference would be to do this, and table this discussion to happen outside of a code review. I think we should consider the overall mix of training pair text length in our pipeline, and how the inference engine segments and actually does the training. The current implementation of the HPLT generates a mix of short and long training data.

(marking as comment as the PR is still draft)

gregtatum (commented):
Update to "comment" status.

eu9ene:
> > As we discussed before dataset modification should not be a default behavior.
> 
> I don't know that this statement is correct. I remember in the review you bringing it up, but I don't think we have an agreement here.
> 
> The importer is more of a pre-processor, turning a document into one side of a "training pair". Training pairs can include single words, short phrases, sentences, or multiple sentences. Datasets are generally preprocessed by OPUS to be ready to be trained, and the HPLT documents are not preprocessed to be used as one side of a training pair. They are documents that are sentence split.
> 
> The statement "modification should not be a default behavior" is strange to me because CJK is transliterating at this same point in the pipeline, and the data augmentations are mutating the sentence pairs here.
> 
> > Completely remove line accumulation and just write line by line as I would expect from a data importer. If we want to experiment with merging some document lines together, we can do it later.
> 
> Merging requires access to the full document, as the merged sentences need to be colocated in the same document, and be concurrent in the sentence stream. This step couldn't be pushed to a later part of the pipeline unless a mapping for the document index, and line index was also saved to disk.
> 
> > Make the condition work based on the number of characters instead of the number of words and disable the accumulation by default.
> 
> My preference would be to do this, and table this discussion to happen outside of a code review. I think we should consider the overall mix of training pair text length in our pipeline, and how the inference engine segments and actually does the training. The current implementation of the HPLT generates a mix of short and long training data.
> 
> (marking as comment as the PR is still draft)

Ok, that sounds good. I'll leave it based on characters and make it optional if it's easy to do. Then we can figure out what length distribution we need and whether it should be a single or multi-sentence dataset. It might also depend on HPLT 2.0. He can ask folks from HPLT and Prompsit how they suggest using this dataset.

tests/test_data_importer.py
```diff
@@ -11,33 +11,6 @@
 TRG = "en"
 CURRENT_FOLDER = os.path.dirname(os.path.abspath(__file__))
 
-hplt_translations = {
```
eu9ene, line 14: This is fragile and hard to update. Also checking the first 10 lines is limited, so I replaced it with checking max line length for the whole dataset.

gregtatum: I would prefer to keep it, as you can copy and paste the new values in there, otherwise you don't know if it's doing the correct thing. This shows the behavior of the code, where otherwise mistakes can happen. For instance in code review I don't know what the new behavior is now, since this would demonstrate a behavior change.

gregtatum (approved):

tests/test_data_importer.py
eu9ene: ok, I can return it, but I need to change the structure so that it's easier to copy-paste it from a file

Issue #751, opened 2024-07-16
Check alignments for CJK


eu9ene:
Do our alignment procedures work correctly for CJK?

Check both: `align.py` and `dataset_importer.py` where we use `sim_align` lib.

eu9ene:
The current idea is to tokenize everything with Moses and then feed tokenized text and alignments for it to OpusTrainer, then detokenize using this config:

```
- Tags: 0.005
  custom_detok_src: <src>
  custom_detok_trg: <trg>
  augment: 1
```

This way it works for both CJK and other languages and we don't need to maintain a complex procedure to remap alignments from Moses to whitespace tokenized to feed to OpusTrainer.

One downside is that this method has a normalizing effect on spaces which is probably fine.

Issue #750, opened 2024-07-16
Investigate OpusTrainer compatibility for CJK


eu9ene:
As far as I understand some modifiers are not needed (UpperCase, TitleCase) but some can still be used:
- Noise
- Inline noise
- Typos? (will the current typos library work with CJK)

gregtatum:
I'm pretty concerned with this one, especially since OpusTrainer relies so heavily on Python's whitespace splitting. We might have to rely on a fork here if we want to use these robustness features. I could see replacing all of the instances with a more robust splitter, for instance based on the ICU segmenter, which is fully internationalized for our current language list.

eu9ene:
Yeah, it basically expects the text to be already "tokenized". There's support for detokenization/retokenization, so we could tokenize (segment) text in Chinese before feeding it to OpusTrainer and then detokenize before forwarding it to Marian. Another issue is that we also tokenize the text with Moses now when producing the alignments, so we might want to reuse this tokenization.

Overall I'm quite confident that it's all possible because HPLT have trained Chinese models and [unit tests](https://github.com/hplt-project/OpusTrainer/tree/main/contrib) for OpusTrainer are even written for Chinese.

eu9ene:
@ZJaume here's a context of our discussion with Jelmer: https://github.com/hplt-project/OpusTrainer/issues/38.

So basically so far the decision was that OpusTrainer supports either whitespace tokenized alignments or Moses tokenized ones with detok directives in the config. See this code: https://github.com/hplt-project/OpusTrainer/blob/7be3b4dcc711d740b4b1a48d91a9b0a13d4ea276/src/opustrainer/modifiers/placeholders.py#L260

We used to do Moses tokenization, then training alignments, and then remapping them back to whitespace-based tokenization to be compatible with OpusTrainer. We feed those alignments and original untokenized corpus to OpusTrainer. We also use guided alignments for the student model based on the bergamot recipe. So for this one we specify [sentencepiece vocab in the training config](https://github.com/mozilla/firefox-translations-training/blob/9956ef28e27d051489d7fceaaf78542be2b9d55d/pipeline/train/configs/opustrainer/student.yml#L29) and OpusTrainer then remaps the alignments to spm based tokenization and feeds to Marian. Kind off crazy but it works :)

Now, with CJK where whitespace-based word splitting doesn't make sense, I changed all that to the following:
1. Same as before, first we run Moses tokenization 
2. Train alignments on the tokenized corpus
[...]

Issue #860, opened 2024-09-23
Investigate switching to ICU segmenter


eu9ene:
We currently use Moses tokenizer for alignments because it seems like it's a standard in the MT world and it's what OpusTrainer supports for detokenization (we will likely feed tokenized text to it to support CJK).

We should investigate switching to the ICU segmenter because we use it in Firefox during inference.

Greg:

> Two things that could be helpful here are the ICU segmenter (which is equivalent to the powerful Intl.Segmenter in javascript. I believe there are ICU bindings available in Python.
> 
> Tutorial: https://www.bedrick.org/notes/python-icu-bindings/
> 
> The other package is the unicodedata one, which gives you more information about codepoints.
> 
> https://docs.python.org/3/library/unicodedata.html
> 
> You can also write regexes using character properties.
> 
> It's also important to run normalization on the text (so the codepoints are all represented consistently), which I don't think the opus tools have done a good job using consistently.
> 
> I think we should lean in to our OpusCleaner fork for fixes here, and once we are done see if they want to accept any of our fixes.

eu9ene:
A reminder od our goals here:

Input: 
- source and target text for each line in a parallel corpus

OpusTrainer's input:
- tokenized text (tokens are the whole words or punctuation). Tokens are split by a whitespace
- alignments information that matches the positions of the tokens

OpusTrainer's output (Marian's input): 
- source and target text with random noise inserted in the correct positions that match the translation
- alignments information that matches sentence piece tokenization performed by marian


I've done some experiments with tokenization/detokenization using [ICU segmenter](https://pypi.org/project/PyICU/), [opus-fast-mosestokenizer](https://github.com/Helsinki-NLP/opus-fast-mosestokenizer) that we use in the pipeline and [sacremoses](https://github.com/hplt-project/sacremoses/tree/master) detokenizer used by OpusTrainer.

[code](https://gist.github.com/eu9ene/b92e15653bd1728c2a7c82a3eaa19122)
[results](https://gist.github.com/eu9ene/71df37376fb6b284f7f90955e33d08e2)

Takeaways:
- ICU is good at tokenization for the languages I tried (ru, en, zh)
- Detokenization is a non-trivial problem and requires implementing language specific rules for punctuation (see [sacremoses detokenizer](https://github.com/hplt-project/sacremoses/blob/65543c34baf589f30260488d882d0060abaa4087/sacremoses/tokenize.py#L543))
- Moses tokenizer doesn't work properly with Chinese
[...]

PR #904, opened 2024-10-26
Update training to support CJK

### LINKED ISSUE #747: Check training for CJK
Does it require any adjustment? Should we change any hyperparameters etc.?

eu9ene:
- Output Moses-tokenized text from the alignments step (we used to remap alignments to whitespace-based tokenization to match the text)
- Use detok OpusTrainer modifiers to detokenize the text back after inline noise is added
- Add specific OpusTrainer configs
- Test Chinese in training tests
- Reduce chunk size for alignments as I had problems with it for CJK

closes #747 #751 #750


eu9ene:
I realized that when running the [experiment](https://github.com/mozilla/firefox-translations-training/pull/839) I forgot to change `finetune-student` Takscluster kind. As a result raw text was fed to OpusTrainer instead of the tokenized one. So it likely messed things up, that's why I have a bad model. Maybe not only because of that :)

gregtatum (changes requested):
I started this review, basically from the last commits to the newer commits. Getting into the business logic of it, I'd prefer for the CI to be clean before I review further. It looks like the `train-teacher` task is failing. Please re-flag me for review when it's ready.

tests/fixtures/__init__.py
```diff
@@ -125,9 +135,10 @@ def run_task(
         env: dict[str, str] = {},
         extra_args: List[str] = None,
         replace_args: List[str] = None,
+        config: Optional[str] = None,
     ):
         """
-        Runs a task from the taskgraph. See artifacts/full-task-graph.json after running a
+        Runs a task from the taskgraph. See data_dir/full-task-graph.json after running a
```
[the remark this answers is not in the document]
eu9ene: yes, I tried to move it there but it didn't work, rolling back

taskcluster/kinds/train-student/kind.yml
```diff
@@ -129,15 +128,14 @@ tasks:
...
-                - artifact: corpus.{trg_locale}.zst
-                  extract: false
             merge-devset:
                 - artifact: devset.{src_locale}.zst
                   extract: false
                 - artifact: devset.{trg_locale}.zst
                   extract: false
             alignments:
                 - artifact: corpus.aln.zst
+                - artifact: corpus.moses.{src_locale}.zst
```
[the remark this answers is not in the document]
eu9ene: I would prefer keeping the names short as it's used in a bunch of places. Moses tokenization is mentioned in the docs so I think when someone gets more familiar with the pipeline it won't be that mysterious.

ZJaume (changes requested):
This is not changing the tokenization for OpusTrainer, right? [Moses has poor support](https://github.com/hplt-project/sacremoses/issues/120) for CJK. I think specific tokenizer for those languages should be used, like jieba or MeCab. I have no experience working with CJK tokenizers, so I'm not really sure which of them should be used.

eu9ene:
> This is not changing the tokenization for OpusTrainer, right? [Moses has poor support](https://github.com/hplt-project/sacremoses/issues/120) for CJK. I think specific tokenizer for those languages should be used, like jieba or MeCab. I have no experience working with CJK tokenizers, so I'm not really sure which of them should be used.

@ZJaume great feedback! I didn't see this issue.

Yes, this is using the default OpusTrainer tokenization method which is Moses. I suggest leaving tokenizer replacement out of the scope of this already big PR. We have a separate issue about this: #860. So the idea is to use ICU segmented for everything including inference in Firefox. The problem is OpusTrainer doesn't support it. So the plan is to investigate if we can use it, then add support in OpusTrainer and only then we'll be able to switch to it in our pipeline. 

I think the current tokenization method might be sufficient for the first experiments with CJK.

eu9ene:
Maybe what I can do here is abstracting a bit from the tokenization method, for example calling artifacts ".tokenized" instead of ".moses". I can also add some TODOs.

pipeline/train/train.py
```diff
@@ -20,6 +20,9 @@
 train_dir = Path(__file__).parent
 
 
+CJK_LANGS = ["zh", "ja", "ko"]
```
gregtatum, line 23: Thought: I guess we could want to put other languages in here that would use this same type of strategy, so CJK_LANGS could be other things. This is probably fine.

taskcluster/kinds/alignments-backtranslated/kind.yml
```diff
@@ -79,6 +79,7 @@ tasks:
                     --corpus_src=$MOZ_FETCHES_DIR/mono.{src_locale}.zst
                     --corpus_trg=$MOZ_FETCHES_DIR/mono.{trg_locale}.zst
                     --output_path=$TASK_WORKDIR/artifacts/mono.aln.zst
+                    --output_tokenized
                     --tokenization=moses
```
gregtatum, line 83: Thought: So thinking through the state of the pipeline, every language will go through tokenized, and it will use Moses. Moses is not great at CJK, but it works well for many other languages. So eventually we'll need better tokenization (one option being #860). But the implementation should be correct for _SOME_ kind of tokenization happening that we can swap in different tokenizers as we decide what's best for everything, or even per-lang tokenizers if we decide to go that route.

taskcluster/kinds/train-student/kind.yml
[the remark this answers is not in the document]
gregtatum: What about just `tok-moses` which is short. We could then do that for things like `tok-icu` or `tok-space`.

taskcluster/kinds/finetune-student/kind.yml
```diff
@@ -104,7 +104,7 @@ tasks:
                     finetune
                     {src_locale}
                     {trg_locale}
-                    $MOZ_FETCHES_DIR/corpus
+                    $MOZ_FETCHES_DIR/corpus.moses
```
gregtatum, line 107: From my idea in another comment, `tok-*` would look like this here, which to me reads a bit more intuitively.

```suggestion
                    $MOZ_FETCHES_DIR/corpus.tok-moses
```

tests/fixtures/config.pytest.enzh.yml
```diff
@@ -0,0 +1,62 @@
+experiment:
```
gregtatum, line 1: Question: We're getting more and more configs. Do you think we should file an issue to make more of the values in them optional so our configs can be more minimal for tests like this? I'm a bit concerned about maintaining a large list of configs when we want to change things.

tests/test_training.py
```diff
@@ -45,39 +45,72 @@ def data_dir():
...
+    zh_config_path = os.path.abspath(os.path.join(FIXTURES_PATH, "config.pytest.enzh.yml"))
+    return zh_config_path if trg_lang == "zh" else None
+
+
 @pytest.fixture()
-def vocab(data_dir):
-    shutil.copyfile("tests/data/vocab.spm", data_dir.join("vocab.spm"))
+def vocab(data_dir, trg_lang):
+    output_path = data_dir.join("vocab.spm")
+    if trg_lang == "ru":
```
gregtatum, line 62: Suggestion: It will help in debugging failures if there is a print statement on which vocab was used.

tests/test_training.py
```diff
@@ -45,39 +45,72 @@ def data_dir():
...
         "MARIAN": marian_dir,
         "SRC": "en",
-        "TRG": "ru",
+        "TRG": trg_lang,
     }
-    data_dir.run_task("alignments-original-en-ru", env=env)
-    shutil.copyfile(
-        os.path.join(data_dir.path, "artifacts", "corpus.aln.zst"), data_dir.join("corpus.aln.zst")
-    )
+    for task, corpus in [("original", "corpus"), ("backtranslated", "mono")]:
```
gregtatum, line 90: Question: Can these be pytest parameters instead? Or do they need to run sequential. If they can be parameters it would make it easier to re-run the tests and isolate failures.

tests/test_training.py
```diff
@@ -45,39 +45,72 @@ def data_dir():
...
+        data_dir.run_task(f"alignments-{task}-en-{trg_lang}", env=env, config=config)
+        shutil.copyfile(
+            os.path.join(data_dir.path, "artifacts", f"{corpus}.aln.zst"),
```
gregtatum, line 93: Polish: Here and elsewhere in the file there is `data_dir.join` which is a bit cleaner.

```py
data_dir.join(f"artifacts/{corpus}.aln.zst")
```

tests/test_training.py
```diff
@@ -3,7 +3,7 @@
 
 import pytest
```
gregtatum, line 4: Praise: Thanks for all the tests!

taskcluster/kinds/train-student/kind.yml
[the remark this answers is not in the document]
eu9ene: ok, I can do that. Another option would be to hide the tokenization methods and call this just ".tok" but it's probably better to be explicit, especially if we for some reason decide to use multiple methods. Also, Moses tokenized part of the pipeline should not be compatible with ICU tokenized, so it's safer to do "tok-icu" etc.

pipeline/train/train.py
eu9ene: yeah, probably we can expand that in the future and rename

taskcluster/kinds/alignments-backtranslated/kind.yml
eu9ene: yes, the tokenization method is an abstraction from alignments and pipeline perspective. The only limitation now is compatibility with OpusTrainer for detokenization

tests/fixtures/config.pytest.enzh.yml
eu9ene: I don't like this either. See https://github.com/mozilla/translations/issues/597. The goal of this PR is to test CJK, we can refactor and consolidate the configs later. I imagine we could just copy the ci config and replace the target language to "zh" in tests.

tests/test_training.py
eu9ene: It's not a test, it's a fixture that produces alignments for the original and back-translated corpus.

PR #906, opened 2024-10-30
Configure vocab for CJK

### LINKED ISSUE #745: Investigate issues with SentencePiece vocabulary for CJK
See comments from Jaume:

https://github.com/mozilla/firefox-translations-training/issues/45#issuecomment-1036191497
https://github.com/mozilla/firefox-translations-training/issues/45#issuecomment-1036198055

eu9ene:
- character coverage
- size

closes #745


pipeline/train/spm-vocab.sh
```diff
@@ -70,6 +80,17 @@ zstdmt -dc "${merged_corpus_trg}" >"${vocab_dir}/data.trg.txt"
...
+      ${TRG} == 'zh' ||
+      ${SRC} == 'ja' ||
+      ${TRG} == 'ja' ||
+      ${SRC} == 'ko' ||
+      ${TRG} == 'ko' ]]
+then
+  character_coverage=0.9995
+else
+  character_coverage=1.0
+fi
```
ZJaume, line 93: As I said in #745, I think default coverage would be enough for all languages if byte fallckback is enabled.

gregtatum (approved):
Seems reasonable to me.

pipeline/train/spm-vocab.sh
eu9ene: ok, replaced with sp default (removed the argument)

ZJaume (approved):

ZJaume:
I'm thinking now that the 64K vocab as a workaround of split 32K vocabs may end up slowing down the decoding process because the softmax will be more expensive. Also I remember in the past the s2s models struggling to learn with large vocabulary outputs like 80k. So I wonder if we will see something of this in the Transformer. We'll see what happens in this preliminary experiments, but I'm starting to be more inclined to split vocabs.

eu9ene:
> I'm thinking now that the 64K vocab as a workaround of split 32K vocabs may end up slowing down the decoding process because the softmax will be more expensive. Also I remember in the past the s2s models struggling to learn with large vocabulary outputs like 80k. So I wonder if we will see something of this in the Transformer. We'll see what happens in this preliminary experiments, but I'm starting to be more inclined to split vocabs.

yeah, I guess I'll rerun the experiment with the latest fixes and we can test it. I'll also work on https://github.com/mozilla/translations/issues/913 so that we can experiment with the split vocabs

PR #907, opened 2024-10-30
Use GCP standard instances for alignment tasks

eu9ene:
[skip ci]


bhearsum (approved):
AIUI, these are long running but can't be continued the way that training tasks can be. Assuming that's true, can you add a comment to this effect, please?

eu9ene:
> AIUI, these are long running but can't be continued the way that training tasks can be. Assuming that's true, can you add a comment to this effect, please?

correct, added a comment

PR #908, opened 2024-10-30
Switch bestbleu to chrF

### LINKED ISSUE #748: Check decoding for CJK
Does decoding, extract-best and other procedures for translation work the same way for CJK?

### LINKED ISSUE #911: Do not use WMTNews as training!
The WMTNews corpus at OPUS is just a compilation of the WMT test sets, so it must not be included as training

https://github.com/mozilla/firefox-translations-training/blob/1f7ab70cd4dbb64e16bb6b38840490c2f2259cb0/configs/autogenerated/en-tr-spring-2024.yml#L79-L79

https://github.com/mozilla/firefox-translations-training/blob/1f7ab70cd4dbb64e16bb6b38840490c2f2259cb0/configs/autogenerated/en-ro-spring-2024.yml#L123

eu9ene:
chrF is now considered more reliable than BLEU, and should work better for CJK
Based on advice from #748

+unify sacrebleu and mtdata versions everywhere
+update config generator and skip some datasets there (it's not final though)

closes #748 
closes #911


ZJaume (approved):

utils/config_generator.py
```diff
@@ -45,6 +45,13 @@
     "lithuanian_legislation_seimas_lithuania",
     # Fails to load from OPUS.
     "SPC",
+    # MTdata duplicates Flores that we pull directly
+    "flores101_dev",
+    "flores101_devtest",
+    "flores200_dev",
+    "flores200_devtest",
+    # Skip OPUS WMT news test sets
```
gregtatum, line 53: Suggestion: This comment says the "what", but not the "why". Perhaps it could be rephrased to something like: "OPUS WMT is used in our evaluations and shouldn't be used for training".

taskcluster/kinds/extract-best/kind.yml
```diff
@@ -81,11 +82,13 @@ tasks:
                 - >-
                     zstd -d --rm $MOZ_FETCHES_DIR/*.zst &&
                     export PYTHONPATH=$PYTHONPATH:$VCS_PATH &&
+                    pip install --upgrade pip &&
+                    pip install -r $VCS_PATH/pipeline/translate/requirements/bestbleu.txt &&
                     python3 $VCS_PATH/pipeline/translate/bestbleu.py
                     -i "$MOZ_FETCHES_DIR/file.{this_chunk}.nbest"
```
gregtatum, line 88: Suggestion: While you are touching this code, can you expand the arguments?

```suggestion
                    --nbest "$MOZ_FETCHES_DIR/file.{this_chunk}.nbest"
```

taskcluster/kinds/extract-best/kind.yml
```diff
@@ -81,11 +82,13 @@ tasks:
                 - >-
                     zstd -d --rm $MOZ_FETCHES_DIR/*.zst &&
                     export PYTHONPATH=$PYTHONPATH:$VCS_PATH &&
+                    pip install --upgrade pip &&
+                    pip install -r $VCS_PATH/pipeline/translate/requirements/bestbleu.txt &&
                     python3 $VCS_PATH/pipeline/translate/bestbleu.py
```
gregtatum, line 87: Suggestion: This file name doesn't really make sense anymore, since it's not using bleu, but chrf. Can we rename it to `extract_best.py` to have it match the kind?

pipeline/translate/bestbleu.py
```diff
@@ -96,7 +96,7 @@ def marian_best_bleu(args, score_function):
 
 def compute_chrf(references, translation):
     hypo = " ".join(translation)
-    refs = [" ".join(r) for r in references][0]
+    refs = [" ".join(r) for r in references]
```
gregtatum, line 99: Suggestion (optional): It would be nice to have a `run_task` test for this file if we're changing code. I'm not sure how we know this code is correct otherwise. I'm marking as optional since it's not changing much here.

gregtatum (approved):
I'm marking as approved, but I'm a bit concerned on knowing if this code is correct without tests. I looked at the `extract-best` task, and there is no logging and at the end there is an outfile artifact which content. Clearly it's writing out something, but it's hard to know what's going on here. I'm fine if you want to go ahead and merge, since it's existing code, but we are using new code paths that haven't been proven yet.

https://github.com/mozilla/firefox-translations-training/runs/32308849855

PR #909, opened 2024-10-30
Fix shortlist pruning for CJK

### LINKED ISSUE #753: Check shortlist for CJK
Does it require and modifications?

eu9ene:
I don't have a good understanding of why some lines are suddenly empty as a result of running "extract_lex".

There are just a few of them and the model trained ok so I assume it's fine. 

@ZJaume did you run into similar issues? Or you don't use the shortlist in your pipeline?

closes #753


gregtatum (approved):

ZJaume (approved):
We still use shortlists for generic models (not for domain models) but we do not train much of those and I don't remember the error happening recently. Maybe in the past, but I'm not sure.

PR #914, opened 2024-10-31
Rename repo

gregtatum:
This updates all of the internal references to the repo for a https://github.com/mozilla/translations rename.


bhearsum (approved):

bhearsum (changes requested):
While poking around I did end up noticing that https://github.com/mozilla/firefox-translations-training/blob/50ecec102f3e9b290ca0d6651c13e3e21c431996/taskcluster/config.yml#L15-L16 ought to be updated too. Both of those can simply change to `translations`.

bhearsum (approved):

PR #916, opened 2024-11-01
Rename docker tags following repository rename

nordzilla:
Our docker tags are labled "ftt" for the previous
repository name "firefox-translations-training".

This patch renames instances of "ftt" to "translations" since the initialism no longer has meaning to new contributors.

### DIFF
--- a/utils/tasks/docker-build.sh
+++ b/utils/tasks/docker-build.sh
@@ -7,14 +7,14 @@ source utils/tasks/docker-setup.sh
 
 docker build \
   --file taskcluster/docker/base/Dockerfile \
-  --tag ftt-base .
+  --tag translations-base .
 
 docker build \
-  --build-arg DOCKER_IMAGE_PARENT=ftt-base \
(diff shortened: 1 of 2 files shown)


nordzilla:
@bhearsum I'm not sure if these tags are used anywhere else that needs changing.

bhearsum (commented):
These don't affect taskcluster tasks at all AFAIK. These are additional consumers of those images. (Taskcluster builds happen through [the docker-image kind](https://github.com/mozilla/translations/blob/main/taskcluster/kinds/docker-image/kind.yml).) So - all good as far as I'm concerned, but I'm probably not the ideal reviewer!

gregtatum (approved):
Looks good to me. To avoid rebuilds you can add these tags as well.

```
docker tag ftt-local:latest translations-local:latest
docker tag ftt-test:latest translations-test:latest
docker tag ftt-base:latest translations-base:latest
```

PR #902, opened 2024-10-24
Rework wasm build scripts for gecko

nordzilla:
@gregtatum 

The goal of this patch is to move much of the functionality from the [build-bergamot.py](https://searchfox.org/mozilla-central/rev/dca2603d55b5b39d3b8ab8e93c08b42563f5aad8/toolkit/components/translations/bergamot-translator/build-bergamot.py) script in Mozilla Central upstream into this repository to better streamline how WASM artifacts are built and exported for use in Gecko.

There are three major changes that occur:

1) The current version of [build-bergamot.py](https://searchfox.org/mozilla-central/rev/dca2603d55b5b39d3b8ab8e93c08b42563f5aad8/toolkit/components/translations/bergamot-translator/build-bergamot.py) applies a patch to the upstream Bergamot source code after cloning the repository. This is no longer necessary because we can make the change permanently upstream here. 

2) [build-bergamot.py](https://searchfox.org/mozilla-central/rev/dca2603d55b5b39d3b8ab8e93c08b42563f5aad8/toolkit/components/translations/bergamot-translator/build-bergamot.py) is moved in its entirety into this repository as `build-wasm.py`. It is then modified to work within this ecosystem. I moved the whole file first so that you can look commit by commit to clearly see what changed from the original source. 

3) The `docker-run` task now utilizes Python instead of bash, and it now has the ability for the user to specify additional volumes to be shared between the host file system and the docker container. 

```
task docker-run -- task inference-build-wasm --volume "HOST_PATH/build-wasm:inference/build-wasm"
``` 

This allows us to build the WASM artifacts within docker and export them directly into tree when making updates in Mozilla Central. 

---

I will also be submitting a sister patch on the Mozilla Central side that updates the fetch/build scripts to utilize this new functionality.

PR #918, opened 2024-11-04
Add emsdk as a git submodule

nordzilla:
#902 reworked the WASM build script to use Python. It pulled some functionality from the pre-existing script within the Gecko source tree.

The Gecko script had to manually clone the `emsdk` repository, because we cannot add a proper git submodule in the Gecko source tree.

In this repository, however, we can and should use git submodules.

This patch removes the manual cloning of `emsdk` and adds it to the repository as a submodule.

### DIFF
--- a/inference/scripts/build-wasm.py
+++ b/inference/scripts/build-wasm.py
@@ -6,6 +6,10 @@ import shutil
 import subprocess
 from typing import Optional
 
+# The emsdk git submodule is set to revision 2346baa7bb44a4a0571cc75f1986ab9aaa35aa03 which
+# corresponds to version 3.1.8. The latest version of emsdk had errors building sentencepiece.
+EMSDK_VERSION = "3.1.8"
+
 SCRIPTS_PATH = os.path.realpath(os.path.dirname(__file__))
 INFERENCE_PATH = os.path.dirname(SCRIPTS_PATH)
 PROJECT_ROOT_PATH = os.path.dirname(INFERENCE_PATH)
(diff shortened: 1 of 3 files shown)


inference/scripts/build-wasm.py
```diff
@@ -131,7 +128,7 @@ def revert_git_patch(repo_path, patch_path):
     subprocess.check_call(["git", "apply", "-R", "--reject", patch_path], cwd=PROJECT_ROOT_PATH)
 
 
-def build_bergamot(args: ArgNamespace):
+def build_bergamot(args):
```
gregtatum, line 97: Question: Did you mean to remove the type here? Does it not pull in correctly? Generally I try to type things at the function boundaries.

inference/scripts/build-wasm.py
```diff
@@ -57,41 +57,7 @@ def ensure_docker():
...
-
-def install_and_activate_emscripten(args: ArgNamespace):
-    git_clone_update(
-        name="emsdk",
-        repo_path=EMSDK_PATH,
-        repo_url="https://github.com/emscripten-core/emsdk.git",
-        revision=EMSDK_REVISION,
-    )
-
+def install_and_activate_emscripten():
```
gregtatum, line 60: Question: Do we have any check here that makes sure we've done a recursive initiation of git submodules? Seems like a point of failure if we don't.

inference/scripts/build-wasm.py
nordzilla: I did remove it intentionally. The explicit import of `namedtuple` struck me as odd because I grepped for it within this repo and didn't see any instances of it. 

Given that some of this script was ported over from mozilla-unified, and how `argparse` itself seems to be a well-understood staple within the Python ecosystem, I decided to take it out. 

It looks like most of the scripts here use `Optional[list[str]] = None`, which I neglected to add. 

I'm happy to put the `namedtuple` back if you want it there. 

I think I would take the automatically provided `--help` menu as a source of truth for the script args.

```
❯ ./inference/scripts/build-wasm.py --help
usage: build-wasm.py [-h] [--clobber] [--debug] [-j J]

options:
  -h, --help  show this help message and exit
  --clobber   Clobber the build artifacts
  --debug     Build with debug symbols, useful for profiling
  -j J        Number of cores to use for building (default: all available cores)
```

gregtatum: Ah yeah this is a bit weird. I don't see it used anywhere else, so this is fine.

PR #920, opened 2024-11-04
Setup WASM test infrastructure for CI

nordzilla:
This patch introduces infrastructure to test our WASM JS using https://vitest.dev/. 

For now, it just configures everything and introduces a stubbed test file with a simple assertion. 

Vitest looked simple, widely recommended, and easy to use to me, but the only JS testing framework that I've personally used is `mochitest` in the Mozilla Unified code.

If people have preferences of what we should use here, I'm happy to switch, and I'd rather do so sooner than later.


inference/wasm/tests/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "version": "1.0.0",
+  "directories": {
+    "test": "tests"
+  },
+  "scripts": {
+    "test": "vitest --run",
+    "test:watch": "vitest"
+  },
+  "keywords": [],
+  "author": "Erik Nordin <enordin@mozilla.com>",
```
gregtatum, line 12: niptick: This author field should probably be removed.

inference/wasm/tests/stub.test.js
```diff
@@ -0,0 +1,7 @@
+import { describe, it, expect } from 'vitest';
+
+describe('Basic Test Suite', () => {
```
gregtatum, line 3: Thought: Ah neat, looks like it's a pre-configured jest test.

gregtatum (approved):

nordzilla:
@jcristau I'm tagging you since you reviewed my previous patch related to this. I hadn't realized that Ben wasn't available to review this week.

taskcluster/kinds/inference/kind.yml
```diff
@@ -65,4 +65,16 @@ tasks:
...
         - -c
         - >-
-            task inference-build-wasm
\ No newline at end of file
+            task inference-build-wasm
+
+  test-wasm:
+    description: "Run build-wasm JS tests for the inference engine"
+    dependencies:
+      build-local: inference-build-wasm
```
jcristau, line 73: Maybe use a different label for the dependency, since AIUI `build-local` refers to the C++ build?

jcristau (approved):
r+ with nit for the taskcluster bits.

taskcluster/kinds/inference/kind.yml
nordzilla: Ah yes, thank you. This was just an oversight from a copy-and-paste of config. A careless mistake. I appreciate you pointing it out.

PR #921, opened 2024-11-04
Make `npm` available to `local` and `inference` docker images

nordzilla:
This patch contains Docker changes that I will need for adding WASM JS tests to CI.

### DIFF
--- a/taskcluster/docker/inference/Dockerfile
+++ b/taskcluster/docker/inference/Dockerfile
@@ -27,6 +27,28 @@ RUN apt-get update -qq \
                           libtool \
     && apt-get clean
 
+ARG NVM_VERSION=0.40.1
+ARG NVM_CHECKSUM=abdb525ee9f5b48b34d8ed9fc67c6013fb0f659712e401ecd88ab989b3af8f53
+
+ARG NODE_VERSION=23.1.0
+ARG NODE_CHECKSUM=c438df636858200cffdfc3b579a1d784047da5c0e70dd647616c215726e254e2
+
+ARG NPM_VERSION=10.9.0
+ARG NPM_CHECKSUM=8e5f6f3429f8cdbe693cdc29904e9d5a7b127a494bd15c804bd54c7403bfcbe7
+
+RUN curl -o install.sh -fsSL https://raw.githubusercontent.com/nvm-sh/nvm/v${NVM_VERSION}/install.sh && \
+    echo "${NVM_CHECKSUM} install.sh" | sha256sum -c - && \
+    bash install.sh && \
+    export NVM_DIR="$HOME/.nvm" && \
+    [ -s "$NVM_DIR/nvm.sh" ] && . "$NVM_DIR/nvm.sh" && \
+    nvm install ${NODE_VERSION} && \
+    nvm use ${NODE_VERSION} && \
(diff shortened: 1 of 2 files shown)


taskcluster/kinds/inference/kind.yml
```diff
@@ -53,12 +53,12 @@ tasks:
...
         - -c
         - >-
-            task inference-test
+            task inference-test-local
 
   build-wasm:
     description: "Build the wasm bindings for the inference engine"
     dependencies:
-      build-local: inference-build-local
+      build-local: inference-test-local
```
jcristau, line 61: What is this change for?

taskcluster/docker/inference/Dockerfile
```diff
@@ -27,6 +27,14 @@ RUN apt-get update -qq \
                           libtool \
     && apt-get clean
 
+RUN curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.40.1/install.sh | bash && \
```
jcristau, line 30: Best practice would be to verify the downloaded file against a known checksum (maybe meaningless in this case as the script itself, and then nvm, go off and download more stuff...)

taskcluster/docker/inference/Dockerfile
```diff
@@ -27,6 +27,14 @@ RUN apt-get update -qq \
...
+    export NVM_DIR="$HOME/.nvm" && \
+    [ -s "$NVM_DIR/nvm.sh" ] && . "$NVM_DIR/nvm.sh" && \
+    nvm install node && \
```
jcristau, line 33: Best specify a version instead of "the current stable version at the time the docker image is built", to avoid breakage at random times when the image is rebuilt (and to ensure the taskcluster cache key is different when the node version changes).

taskcluster/kinds/inference/kind.yml
nordzilla: I'm preemptively renaming inference-test to inference-test-local to make way for inference-test-wasm in a subsequent PR.

taskcluster/docker/inference/Dockerfile
nordzilla: No, that's fine. I can compute a checksum and compare it.

taskcluster/docker/inference/Dockerfile
nordzilla: Sure. I'll lock it in to a version.

taskcluster/kinds/inference/kind.yml
jcristau: Right, but here you're changing the `build-wasm` task's dependency away from `inference-build-local`, `inference-test` is not involved?

nordzilla: Oh sorry, I actually misread which line you were asking about (still having my morning coffee). 

To me it made sense to not even build the WASM unless the C++ _tests_ are passing, rather than just the C++ build. 

Somewhat of a drive-by change, but I feel that it should have always been this from the start.

nordzilla: I'm just going to move this change entirely to the other PR. Yesterday, I was thinking "I'll just make all the config changes in one patch." But this change is more related to the new tests than it is to the Dockerfiles. 

I understand your confusion on why it's here.

gregtatum (approved):

Issue #933, opened 2024-11-18
decoding-teacher config property is not being used in translate.sh or translate-nbest.sh


gregtatum:
The values aren't being forwarded to Marian.

Issue #931, opened 2024-11-15
Experiment with distillation data inference


gregtatum:
I'm going to do several experiments around distillation data decoding, and it will be easier to write up the results here as they are all related to the same part of the pipeline, `translate-mono-src` and `translate-corpus`.

 * Investigate using CTranslate2 for decoding #165
 * Remove the teacher ensemble #778
 * Improve student GPU utilization #785

My plan is to test on `da-en` since it was a good model result and should be indicative of quality drops.

The data for this experiment is [available in this spreadsheet](https://docs.google.com/spreadsheets/d/19n2Pn_gXpgz8d5w8d5HfUfSR9e7WDYcLkgkat4EStqg/edit?gid=1587084528#gid=1587084528). I measured both the GPU utilization and how much data was being written into the target file in terms of bytes/sec. Each run operated on the same subset of the data. I measured 15 minutes of translations across 5 batches, and summarized the results to get a sample of how fast the translations were happening.

| decoder       | precision | teachers | maxi-batch-words | maxi-batch | gpu utilization | bytes/sec      | vs 500 | vs Marian Best |
|---------------|-----------|----------|------------------|------------|-----------------|----------------|--------|----------------|
| marian        | float32   | 2        | 500              | 1,000      | 64.7            | 156,118        |        |                |
| marian        | float16   | 2        | 500              | 1,000      | 57.5            | 183,627        | 118%   |                |
| marian        | float16   | 2        | 4,000            | 1,000      | 61.5            | 329,703        | 211%   |                |
| marian        | float16   | 2        | 5,000            | 1,000      | 57.4            | 338,798        | 217%   |                |
| marian        | float16   | 2        | 5,000            | 10,000     | 57.2            | 348,805        | 223%   |                |
| marian        | float16   | 1        | 5,000            | 10,000     | 55.9            | 601,635        | 385%   |                |
| marian        | float16   | 2        | 8,000            | 1,000      | -               | Out of Memory  | -      | -              |
| ctranslate2   | float16   | 1        | 5,000            | -          | 97.2            | 1,187,192      | 760%   | 197%           |

Then run the experiments on the decoder/ensemble configurations.

| inference   | teacher ensemble | student comet | vs baseline | gpu hours | vs baseline | wall time  | vs baseline |
| ----------- | ---------------- | ------------- | ----------- | --------- | ----------- | ---------- | ----------- |
[...]

PR #934, opened 2024-11-18
Adjust default values for batching

gregtatum:
This results in a 2.2x speed up for train-mono tasks according to my testing in #931. There's also a bug with loading in Marian args #933, so I'm putting the values in the decoder.yml file as well.


ZJaume:
The `fp16: true` option may be safer than jus `precision: float16`, as it is an alias for float16 precision that also adjusts other precision scaling parameters.

gregtatum:
> The fp16: true option may be safer than just precision: float16, as it is an alias for float16 precision that also adjusts other precision scaling parameters.

I checked the Marian code and it matters for training, but it's the same for just translating. I'm happy to switch to it though.

eu9ene (approved):

PR #938, opened 2024-11-22
fix: get rid of `hunspell` python package

bhearsum:
As far as I can tell, `cyhunspell` is a full on replacement for `hunspell`, to the point where the module name that is imported is the same. It looks to me that we're installing `hunspell`, and then immediately overwriting those files with `cyhunspell`, leaving the former unused.


ZJaume:
yes, `hunspell` package is not needed for bicleaner.

gregtatum (approved):

gregtatum:
It would be nice to get a CI run before we merge this.

PR #945, opened 2024-11-26
Use `Intl.Segmenter` instead of `ssplit` for segmentation in WASM builds

nordzilla:
## Description

> [!NOTE]
> I have two follow-up issues filed for after this patch lands. Please take them into consideration when reviewing.
> * #946 
> * #947

This pull request fully removes the dependency on [ssplit](https://github.com/browsermt/ssplit-cpp) from WASM builds of the inference engine, instead utilizing the [Intl.Segmenter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Segmenter) implementation within the JavaScript context in which the WASM is being run. This offers a number of benefits.

---

### Correctness Improvements
`Intl.Segmenter` is created from a specific locale and knows how to segment text according to that language.

`ssplit` has limited support for CJK segmentation and is a one-size-fits-all segmentation solution that overall works well, but is not robust enough in its current state for all of the languages we want to support.

As an experiment, I cherry-picked the new test cases from this pull request onto the main branch and ran them.

| Category     | Passed | Failed | Total |
|--------------:|-------:|-------:|------:|
| Test Files    |      5 |      7 |    12 |
| Test Cases    |     47 |     21 |    68 |

> [!NOTE]  
> While the `Intl.Segmenter` improves correctness for languages that do not utilize whitespace between sentences, I have no reason to believe that it has regressed the preexisting behavior of translating between two languages that utilize whitespace between sentences. As a check, I did a diff of the HTML output of the [benchmark page](https://gregtatum.github.io/taskcluster-tools/src/benchmark/) after doing a pivot translation from Spanish to French, and the output was identical both before and after the changes in this PR.

---

### Upstream Updates
By moving to `Intl.Segmenter` we also get free, standardized updates whenever its algorithms are improved as part of the JavaScript API.

---

### Compiled WASM Binary Size Reduction

By fully removing `ssplit` as a dependency from the WASM build, we reduce the size of our compiled WASM binary:

| File          | Before     | After     |
|-----------------------|------------|-----------|
| WASM (Uncompressed)    | 5.01 MB    | 4.73 MB   |
| WASM (Compressed)       | 1.78 MB  | 1.71 MB   |

---

### Similar Performance

The new implementation actually showed improved performance on the [benchmark page](https://gregtatum.github.io/taskcluster-tools/src/benchmark/). However, this is negligible with a sample size of 1, and is almost certainly within the range of statistical noise.

Nonetheless, I feel encouraged that there is no major or obvious regression coupled with these changes.

> [!NOTE]  
> * Only `esfr` translation benchmarks were profiled (links in table).
> * The benchmark metrics in the table related to rates are from distinct runs without the profiler running.

| Lang Pair       | Sentence Segmentation             | Benchmark Time (s) | Words Per Second (wps) | Tokens Per Second (tps) | Max Memory Usage (MB)                |
|:---------------:|:-----------------:|--------------------:|------------------------:|------------------------:|-------------------------------------:|
| `esen`          | `ssplit`          |                4.53 |                1531.22 |                2415.62 |                  N/A                 |
| `esen`          | `Intl.Segmenter`  |                4.54 |                1529.19 |                2412.43 |                  N/A                 |
| `esfr`          | `ssplit`          |                9.46 |                 734.11 |                1158.12 | [359](https://share.firefox.dev/3Zn31zg) |
| `esfr`          | `Intl.Segmenter`  |                9.38 |                 739.58 |                1166.76 | [350](https://share.firefox.dev/3VewPf1) |

---


inference/wasm/tests/engine/translations-engine.worker.mjs
```diff
@@ -423,7 +423,30 @@ function cleanText(sourceText) {
   const whitespaceAfter = result[3];
   let cleanedSourceText = result[2];
 
-  cleanedSourceText = cleanedSourceText.replace(/\u00AD/g, "");
+  // Remove all soft hyphens from the text.
```
gregtatum, line 426: Suggestion: I'm a little hesitant to micro-optimize non-production code, but since this is supposed to be close to what we ship in Gecko, we should only do this for locales where it matters. So here we would check for a locale before applying these fixes. I may leave these same comments on the Gecko side when I review it.

inference/wasm/tests/engine/translations-engine.worker.mjs
```diff
@@ -423,7 +423,30 @@ function cleanText(sourceText) {
...
+  // Actual Segmentation:
+  //   - Object { index: 0, segment: 这是第一句话。“ }
+  //   - Object { index: 8, segment: 这是第二句话。” }
+  //
+  // By inserting a space between the full-width punctuation and the Left Double Quotation Mark,
+  // we can trick the segmenter into breaking the sentence at the correct location.
+  //
+  // This code may be able to be removed with further upstream improvements to Intl.Segmenter.
+  // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Segmenter
+  cleanedSourceText = cleanedSourceText.replaceAll(/([。！？])“/g, "$1 “");
```
gregtatum, line 449: Suggestion: Again, micro-optimizing non-production code, but this regex could be cached to not be compiled on every function run, since it will be run frequently.

inference/src/translator/annotation.cpp
```diff
@@ -34,8 +34,83 @@ void AnnotatedText::appendSentence(string_view prefix, std::vector<string_view>:
...
+  }
+
+  return !shouldOmitSpaceBetweenSentences();
+}
+
+void AnnotatedText::maybeAppendHTMLTagsFromWhitespace(string_view whitespace) {
+  size_t currentIndex = 0;
+  while (currentIndex < whitespace.size()) {
+    // Find the next open bracket '<' for an HTML tag.
+    size_t tagStart = whitespace.find('<', currentIndex);
```
gregtatum, line 68: Thought: So this can be more complicated, and can break with how the web actually works. `<pre>x < 5</pre>` will render perfectly fine on the web. However, I think this may be matching existing behavior with the HTML parsing.

Suggestion: It may be worth doing a little bit of investigating here.

inference/wasm/tests/test-cases/translate-html-multi-paragraph-no-pivot.test.mjs
```diff
@@ -0,0 +1,62 @@
...
+import { describe } from "vitest";
+import { runTranslationTest } from "./test-cases/shared.mjs";
+
+/**
+ * This file tests the WASM bindings for non-pivot translation requests
+ * that contain HTML tags within the source text, extended to multiple paragraphs.
+ */
+const testCases = [
+  {
+    sourceLanguage: "es",
```
gregtatum, line 14: Thought: Just as an FYI, this style of test won't work with Jest's `matchInlineSnapshot`, since it's iterating over data, rather than including the assertions inline inside of a `matchInlineSnapshot` call. No action is needed, I just wanted to point it out.

inference/wasm/tests/models/enzh/lex.50.50.enzh.s2t.bin.gz
gregtatum, line 1: I'm fine with it since it's Git LFS.

inference/wasm/tests/test-cases/translate-html-multi-paragraph-no-pivot.test.mjs
```diff
@@ -55,6 +55,15 @@ const testCases = [
       "<b>The blue dog. </b><u><i>It runs in the park.\n\nIt looks very happy. </i></u>It's playing with other dogs.",
     isHTML: true,
   },
+  {
+    sourceLanguage: "en",
+    targetLanguage: "zh",
```
gregtatum, line 60: Suggestion: This may be surprising for future code wanderers without a comment explaining it. Squash merges remove these nice commit messages, yay GitHub. Maybe just do a quick search/replace for `targetLanguage: "zh",` to add a comment.

gregtatum (commented):
I'm leaving my first round of comments, I still need to review 5f6ca11439346519ffcd44927a7a895de0dd08a9, but have looked at all the other commits.

inference/src/translator/annotation.cpp
```diff
@@ -34,8 +34,83 @@ void AnnotatedText::appendSentence(string_view prefix, std::vector<string_view>:
...
+    return false;
+  }
+
+  return (
+    // Japanese does not use space between sentences.
+    targetLanguage_ == "ja" ||
+    // Korean does not use space between sentences.
+    targetLanguage_ == "ko" ||
+    // Chinese does not use space between sentences.
+    targetLanguage_ == "zh"
```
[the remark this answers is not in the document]
nordzilla: Agreed, I will file a follow-up on this!

inference/BERGAMOT_VERSION
```diff
@@ -1 +1 @@
-v0.4.5
+v0.5.0
```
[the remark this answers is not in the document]
nordzilla: I would like to eventually rename everything to `inference`, but we have so many references to the word "bergamot" both in this repo and in Firefox, and even in Remote Settings. It feels like something that may get tangled up and could maybe be done all at once.

inference/wasm/tests/engine/translations-engine.worker.mjs
nordzilla: I was actually thinking this would be fine, because I thought that JavaScript was "smart" about only compiling regex literals once and caching them globally. 

However, if I'm understanding correctly according to this post (https://stackoverflow.com/a/32524171), it turns out that even though the regex literal is only "compiled" once globally, it still creates a regex object each time.

In that case, I think any regex's in this `cleanText()` function should be cached on the class. 

I'll make this change in both locations.

inference/wasm/tests/test-cases/translate-html-multi-paragraph-no-pivot.test.mjs
nordzilla: Agreed, I'll add comments to all of them.

inference/src/translator/annotation.cpp
nordzilla: You're definitely right about an edge case like `<pre>x < 5</pre>` being an issue in HTML parsing in general. 

However, at this stage in the algorithm, we are only dealing with pre-parsed HTML tags that were are adjacent to sentence-separating whitespace and need to be preserved. Consider the following example:

I've added the following to this function:
```c++
  std::cout << "!!! Whitespace Value: " << whitespace << std::endl;
```

Then I've run this test case:
```javascript
"<b>The closing for this bold tag and <em>emphasis tag will show up because they are adjacent to a whitespace sentence boundary.</em></b> However, this <u>underline</u> tag will not be present in the middle of a sentence.",
```

And the resulting logs:
```
!!! Whitespace Value: 
!!! Whitespace Value:  
!!! Whitespace Value: 
!!! Whitespace Value: 
!!! Whitespace Value: </em></b> 
!!! Whitespace Value: 
!!! Whitespace Value: 
!!! Whitespace Value: </em></b>
!!! Whitespace Value: 
```

---

I will try to make the comments and documentation more clear here that we are only extracting pre-parsed tags that are adjacent to sentence-separating whitespace.

inference/wasm/tests/engine/translations-engine.worker.mjs
nordzilla: (See reply on other comment)

inference/BERGAMOT_VERSION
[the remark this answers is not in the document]
gregtatum: Yeah agreed!

inference/wasm/tests/engine/translations-engine.worker.mjs
gregtatum: Yeah I feel like regexes show up in tight for loops when I've profiled.

inference/src/translator/annotation.cpp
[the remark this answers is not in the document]
nordzilla: * #950

nordzilla:
@gregtatum 

Thanks for the review!

I've made changes according to most of your feedback so far ([changeset](https://github.com/mozilla/translations/compare/bbf255178c076e8a978491e619f5b8c2d555bd5f..86f4bab5634502aeeedc9b321aeff4d21b224c40)).

Any comments that are not yet resolved have replies.

Happy to keep iterating on this as needed!

inference/src/translator/text_processor.cpp
```diff
@@ -106,6 +106,117 @@ void TextProcessor::process(std::string &&input, AnnotatedText &source, Segments
...
+   * The address of the internal pointer is passed to JavaScript where it is allocated
+   * via `malloc`, however we are responsible for deallocating the memory via `free` on
+   * the C++ side.
+   *
+   * Wrapping the deallocation in the checked destructor ensures that once the memory has been 
+   * allocated on the JS side, it is guaranteed to be deallocated when this stack frame is popped,
+   * whether from returning or from an exception being thrown.
+   */
+  struct ScopedPtr {
+    int32_t* ptr = nullptr;
```
gregtatum, line 123: Question: Would this be more portable as a `uintptr_t`?

inference/src/translator/text_processor.cpp
```diff
@@ -106,6 +106,117 @@ void TextProcessor::process(std::string &&input, AnnotatedText &source, Segments
...
+    const inputUTF16 = UTF8ToString($0);
+    const lang = UTF8ToString($1);
+
+    // Segment the UTF-16 input with the Intl.Segmenter.
+    const segmenter = Module.getOrCreateSentenceSegmenter(lang);
+    const sentencesUTF16 = Array.from(segmenter.segment(inputUTF16));
+    const sentenceCount = sentencesUTF16.length;
+
+    // Allocate enough space to mark the start and end UTF-8 byte indices for each sentence.
+    const bytesPerInt = 4;
```
gregtatum, line 172: Suggestion: If `uintptr_t` works, then this could be `sizeof(uintptr_t)`.

inference/src/translator/text_processor.cpp
```diff
@@ -106,6 +106,117 @@ void TextProcessor::process(std::string &&input, AnnotatedText &source, Segments
...
+            segmenters.set(lang, new Intl.Segmenter(lang, { granularity: "sentence" }));
+          }
+
+          return segmenters.get(lang);
+        };
+      })();
+    }
+
+    // Convert the UTF-8 C++ strings into UTF-16 for JavaScript.
+    const inputUTF16 = UTF8ToString($0);
```
gregtatum, line 163: Question: Was this change performance profiled? I remember your initial concerns with the approach.

inference/src/translator/text_processor.cpp
```diff
@@ -106,6 +106,117 @@ void TextProcessor::process(std::string &&input, AnnotatedText &source, Segments
...
+  EM_ASM({
+    // Attach a lazily initialized cache of the Intl.Segmenter to the WASM Module.
+    // This is critical to prevent creating one instance of Intl.Segmenter per
+    // translation request.
+    if (!Module.getOrCreateSentenceSegmenter) {
+      Module.getOrCreateSentenceSegmenter = (function() {
+        let segmenters = new Map();
+
+        return function(lang) {
+          if (!segmenters.has(lang)) {
```
gregtatum, line 153: Nitpick: This may not matter, but since this is performance critical code, this could remove a lookup, and is the general pattern I use for this kind of code.

```js
let segmenter = segmenters.get(lang);
if (!segmenter) {
  segmenter = new Intl.Segmenter(lang, { granularity: "sentence" });
  segmenters.set(lang, segmenter);
}
return segmenter;
```

gregtatum (approved):
Looks good to me! It would be nice to do some performance profiling of the new allocation behavior around the utf8-16 conversions if you haven't already, and also benchmark this on our internal benchmark and maybe a few text heavy wikipedia pages.

gregtatum:
Ah I missed the perf numbers at the end. The only potential follow-up would be to run it on a text heavy wikipedia page, but I'm happy with a profile on the benchmark.

inference/src/translator/text_processor.cpp
nordzilla: I don't think that it would. The C++ reference [section on integer types](https://en.cppreference.com/w/cpp/types/integer) describes `uintptr_t` as an "unsigned integer type capable of holding a pointer to void."

My understanding is that `uintptr_t` is to be used when you need to treat the actual numerical value of an address as an integer to perform arithmetic operations on it. 

In this case, I truly am just creating a pointer to an array of signed 32-bit integers, and I don't particularly care about what the size of my addresses are (though, in Emscripten WASM32, they are also 32 bits). 

I need to assign these values from the JavaScript space, using `setValue()` with one of the listed [LLVM IR types](https://emscripten.org/docs/porting/connecting_cpp_and_javascript/Interacting-with-code.html#access-memory-from-javascript), `i32` in this case. 

I think that this code is already portable in the sense that even if Emscripten were to hypothetically find a way to update all existing projects from WASM32 address space to a WASM64 address space, then I would still have an array of 32-bit integers here, even if my pointers to those integers have doubled in size. 

Since I'm not actually operating on the pointer's address value itself, I don't think that `uintptr_t` is the right type here. I truly want a pointer to `int32_t` regardless of the architecture.

inference/src/translator/text_processor.cpp
nordzilla: See detailed response [here](https://github.com/mozilla/translations/pull/945#discussion_r1868004804).

inference/src/translator/text_processor.cpp
nordzilla: There are profiles in the **Similar Performance** section of the [PR description](https://github.com/mozilla/translations/pull/945#issue-2696456246).

ZJaume:
I was able to run our sentence splitting benchmark for ICU4X, see "2022 (mixed) avg" sheet for a summary and "2022 (mixed)" sheet for details on each tested language: https://docs.google.com/spreadsheets/d/1mGJ9MSyMlsK0EUDRC2J50uxApiti3ggnlrzAWn8rkMg/edit?usp=sharing

Not the best, but also no significant flaws.

nordzilla:
> I was able to run our sentence splitting benchmark for ICU4X, see "2022 (mixed) avg" sheet for a summary and "2022 (mixed)" sheet for details on each tested language: https://docs.google.com/spreadsheets/d/1mGJ9MSyMlsK0EUDRC2J50uxApiti3ggnlrzAWn8rkMg/edit?usp=sharing
> 
> Not the best, but also no significant flaws.

This is really great @ZJaume, thank you for putting this together.

Issue #375, opened 2024-01-19
Switch on-push away from level 1 Taskcluster workers


bhearsum:
We had a meeting with our security team to talk about the language training pipeline. We agreed that because we generate artifacts that eventually ship to users, that we would move production model training off of level 1 workers. Additionally, because the needs of the training pipeline prevent us from running on level 3 instances (most notably - we need access to live logs and interactive tasks, which is not allowed on level 3), that we would create something new for this. The obvious thing to do is call it "level 2", but we still haven't determined exactly what the name will be.

In any case, at least the following will be needed:
- A new GCP project to run instances in
- Changes to [config.yml](https://github.com/mozilla/firefox-translations-training/blob/main/taskcluster/config.yml) and [.taskcluster.yml](https://github.com/mozilla/firefox-translations-training/blob/main/.taskcluster.yml) to add support for this new level (or whatever we call it)
- New images built for this new level, through [monopacker](https://github.com/mozilla-platform-ops/monopacker)
- [Worker image](https://hg.mozilla.org/ci/ci-configuration/file/tip/worker-images.yml) and [worker pool](https://hg.mozilla.org/ci/ci-configuration/file/tip/worker-pools.yml) definitions in ci-configuration

When we make this change, the production training will no longer be able to pull artifacts from level 1 caches. If we care about any of the artifacts currently cached there we have a couple of options:
- Migrate them to the new GCP bucket we're setting up for long term storage (#191)
- Add level 2 index routes that point at the level 1 tasks (this would be a one time temporary measure - not something we can do on an ongoing basis)

bhearsum:
I discussed this with @JohanLorenzo. We agreed that we'll forego this because:
- We've been training and shipping models trained on Level 1 instances for a long time already
- The artifacts that we ship don't execute arbitrary code in the way that other things we require Level 3 for do
- There's likely to be disruption to training if we do this
- Implementing this is a non-trivial effort

It's possible we may revive this at some point in the future, but at the moment we're considering it not worthwhile for the above reasons.

PR #949, opened 2024-11-27
feat: upgrade cpu workers to ubuntu 24.04 generic-worker image

bhearsum:
This will remove docker-worker from translations, and use the most recent generic-worker version instead. This halfway unblocks #375 and #466 (the other half is upgrading the GPU workers to this image, which is still in progress).

The other notable part of this is that it moves these workers to a common shared image that will be updated regularly. This means that we'll regularly be picking up things like Taskcluster version upgrades. There are risks to this of course, in that any change to the image can theoretically cause issues with tasks that use it. Over in https://github.com/mozilla-releng/fxci-config/pull/192, I'm working on adding the ability for us to run the training pipeline to help validate new images before they land to help minimize this.

We may also want to consider pinning to a specific image the next time we cut a `release` branch to minimize risk/change during long training sessions.

(If someone wants to make the case that we should always pin to a specific image rather than a rolling one, we can consider that.)

### DIFF
--- a/taskcluster/config.yml
+++ b/taskcluster/config.yml
@@ -64,32 +64,32 @@ workers:
             provisioner: '{trust-domain}-{level}'
             implementation: docker-worker
             os: linux
-            worker-type: '{alias}'
+            worker-type: 'b-linux-large-gcp-d2g'
         b-linux-large-gcp-300gb:
             provisioner: '{trust-domain}-{level}'
             implementation: docker-worker
             os: linux
-            worker-type: '{alias}'
+            worker-type: 'b-linux-large-gcp-d2g-300gb'
         b-linux-large-gcp-1tb-32-256:
             provisioner: '{trust-domain}-{level}'
             implementation: docker-worker
             os: linux
-            worker-type: b-linux-large-gcp-1tb-32-256-d2g
+            worker-type: 'b-linux-large-gcp-1tb-32-256-d2g'
         b-linux-large-gcp-1tb-32-256-standard:
             provisioner: '{trust-domain}-{level}'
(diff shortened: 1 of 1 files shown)


bhearsum:
This change originally included a forced cache bust to make sure the entire pipeline ran, which I've since removed to avoid taking it on main. That graph is https://firefox-ci-tc.services.mozilla.com/tasks/EyBbWlbmTpCQOT0Lwtp7pw, and confirms that the pipeline runs fine on the new image.

gregtatum (approved):
Looks reasonable to me, but I'd like @eu9ene's review as well.

taskcluster/config.yml
```diff
@@ -64,32 +64,32 @@ workers:
...
             implementation: docker-worker
             os: linux
-            worker-type: '{alias}'
+            worker-type: 'b-linux-large-gcp-d2g-300gb'
         b-linux-large-gcp-1tb-32-256:
             provisioner: '{trust-domain}-{level}'
             implementation: docker-worker
             os: linux
-            worker-type: b-linux-large-gcp-1tb-32-256-d2g
+            worker-type: 'b-linux-large-gcp-d2g-300gb'
```
eu9ene, line 77: @bhearsum I don't understand this change. Why everything is 300gb now? I guess it will break the pipeline steps that need 1TB of disk and other specific machine requirements. Or the `worker-type` setting doesn't control those?

bhearsum: Sorry...this part was an oversight! My testing was all done with 300gb workers. I'll fix this up.

eu9ene (approved):

PR #952, opened 2024-12-05
Add pyright to check python types

gregtatum:
This starts some incremental static type checking for the project in the utils/* directory.


taskcluster/kinds/tests/kind.yml
```diff
@@ -91,6 +91,20 @@ tasks:
             task lint-black
     run-on-tasks-for: ["github-push", "github-pull-request"]
 
+  lint-pylance:
+    # Run python's pylance type checker.
```
nordzilla, line 95: > Question

I think maybe this is supposed to be `pright` instead of `pylance`. Is that right?

nordzilla (approved):
Changes all look reasonable to me, though I'm not a Python expert. 

I am certainly in favor of having type checkers whenever possible.

I just have the one inline comment question about `pylance` vs. `pyright`

> Edit: Looks like the pylance thing failed CI as is, so I think it was just an oversight.

taskcluster/kinds/tests/kind.yml
gregtatum: Whoops, yeah pylance is the name of the VSCode extension. That's a mistake.

gregtatum:
Looks like it requires the world to be installed to find all of the modules, and is currently failing.

PR #953, opened 2024-12-05
fix: set permission for train action

bhearsum:
This used to default to the action name, but changed in taskgraph 11.0. Without this, we try to use the `generic` permission, which doesn't have the scopes needed.

See https://github.com/taskcluster/taskgraph/pull/553


bhearsum:
Removing this from review queues until I manually test training actions, and look into test failures.

taskcluster/translations_taskgraph/actions/train.py
```diff
@@ -64,6 +64,7 @@ def validate_pretrained_models(params):
     symbol="train",
     description="Initiate part or all of the training pipeline",
     cb_name="train",
+    permission="train",
```
ahal, line 67: Heh, that's a lot of "train"

ahal (approved):

taskcluster/translations_taskgraph/actions/train.py
bhearsum: :train: :train2: :bullettrain_side: :steam_locomotive: :bullettrain_front:

Issue #843, opened 2024-09-12
Made-up words in translations


eu9ene:
We recently got a [report](https://github.com/mozilla/firefox-translations-training/issues/816) about made up words in Turkish.

I also tested the new ru-en model and noticed a lot of non-existent words there as well. 

For example: https://habr.com/ru/articles/842924/

The title is translated as "Language environment and teacher-**sdocument**. Non-**evous** moments". 
It should be "Language environment and native-speaking teachers. Non-obvious points".

We should investigate why this happens. It might have something to do with the recent robustness fixes or using a shortlist.

eu9ene:
Translation by the teacher model:
"Language environment and native teachers. Non-obvious moments".

I don't see any made-up words in the full text as well even though the fluency is far from perfect. Only a couple of informal abbreviation were translated as is, for example "дз" (home work) as "DZ" and "выпускники инязов" (graduates of universities that specialize in foreign languages) as "graduates of inaz". This text is written in a very informal style, so I guess overall lower quality of translation is expected.

This means we have the issue with made-up words specifically for the student model.

eu9ene:
I don't see any made-up words in translation of news where more formal language is used.

gregtatum:
I wonder if this could be caused by the decoder being too shallow, or the decoder not being big enough. This could be good to experiment with, and also test the difference in the performance of the models.

marco-c:
Maybe lexical shortlisting could also be affecting this?

eu9ene:
I see the same thing in the new en-ru model in Base architecture. It has 0.88 COMET on Flores compared to Google 0.9 and Microsoft 0.89 so the issue is not in translation quality. The grammar is also pretty good and very readable. 

I see a lot of made-up words in a [random CNN article](https://www.cnn.com/2024/12/19/politics/government-shutdown-disaster-aid-congress/index.html). 

We should test older models, maybe we introduced these issues when implementing robustness fixes. It might have to do with the alpha parameter in the Sentencepiece model.

https://github.com/mozilla/translations/blob/cfbaf72c57f05021c8960f2f06baeba55bcb2798/pipeline/train/configs/training/student.train.yml#L24

eu9ene:
Surprisingly I don't see any made-up words in the evaluation outputs. Also, the translation quality is just better when I look at the evals output of the final quantized model and compare it to the results from about:translations.

gregtatum:
I think this is fixed by removing the lexical shortlists.

PR #955, opened 2024-12-11
Add a util for local remote settings

gregtatum:
I've been using this to test the new model architectures end to end.


Taskfile.yml
```diff
@@ -49,6 +49,17 @@ tasks:
           poetry run python -W ignore utils/taskcluster_downloader.py
           --mode=model {{.CLI_ARGS}}
 
+  local-remote-settings:
+    desc: Runs models for use within Firefox via local Remote Settings.
+    summary: |
+      The artifacts will be saved to: ./data/artifacts
+      Example: `task local-remote-settings -- --task-group-ids GU9ZyWFhRDe_nxlAHcen8g`
+    deps: [poetry-install-utils]
```
nordzilla, line 57: > [!CAUTION] 
> I had to add `zstandard = "^0.22.0"` to the `[tool.poetry.group.utils.dependencies]` list for this to work.
> ```diff
> diff --git a/pyproject.toml b/pyproject.toml
> index 787fa1ab..cfbb3236 100644
> --- a/pyproject.toml
> +++ b/pyproject.toml
> @@ -48,6 +48,7 @@ taskcluster-taskgraph = "^11.1.0"
> kinto-http="11.7.0"
> # Use an outdated version of pydantic due to dependency requirements conflict.
> pydantic="1.10.19"
> +zstandard = "^0.22.0"
>
> # This install group is for running tests. Note that any dependencies in the
>  # pipeline are installed separately through the run_task test abstraction. This
> ```

utils/local_remote_settings/mount/translations.ini
```diff
@@ -0,0 +1,155 @@
+#
```
nordzilla, line 1: > Praise

Thanks for adding all of these.

nordzilla (changes requested):
I tried this out on my Linux and macOS machine and it worked.

> [!CAUTION]
> Something about the local Remote Settings configuration when pulling from taskcluster does not work with the following `mozconfig` option:
> ```
>  ac_add_options --enable-debug
> ```
> 
> ---
> 
> **Suggestion**
> Add this info to the `--help` menu. 

> [!NOTE]
> 
> 
> **Thought**:
> I initially got confused why this was not working and did not notice the section of the `--help` menu that says to comment out the lines in the Remote Settings file. 
> 
> ---
> 
> **Suggestion**:
> I think it would be a better experience to have this task default to the help menu if no flags are passed. And then it can accept explicit flags `--mirrored` or `--taskgroup_ids`. 
>
> That way a user is forced to encounter the `--help` screen unless they already know what they're doing.

eu9ene:
This is great! 

Is it possible to download a model from the model registry to test it? The Taskcluster tasks tend to expire and we don't have them for older models. 

Also is it possible to use a regular downloaded Nightly with the remote-settings addon without building it?

Taskfile.yml
gregtatum: This is actually an issue on all of our utils at the moment. I'd like to follow-up and consolidate the test/utils groups I think.

gregtatum:
> Is it possible to download a model from the model registry to test it? The Taskcluster tasks tend to expire and we don't have them for older models.

Which model registry are you referring to? The GCP bucket? We can always file issues to add more sources, these were just the ones I was working with right now.

> Also is it possible to use a regular downloaded Nightly with the remote-settings addon without building it?

Yeah, I updated the docs with a new solution to bypass the validation checks using the browser console.

Taskfile.yml
gregtatum: https://github.com/mozilla/translations/issues/971

eu9ene:
> > Is it possible to download a model from the model registry to test it? The Taskcluster tasks tend to expire and we don't have them for older models.
> 
> Which model registry are you referring to? The GCP bucket? We can always file issues to add more sources, these were just the ones I was working with right now.
> 
> > Also is it possible to use a regular downloaded Nightly with the remote-settings addon without building it?
> 
> Yeah, I updated the docs with a new solution to bypass the validation checks using the browser console.

I was referring to the current Github models repo. Basically, I need to download super old model files to compare with the new model (see #843).

gregtatum:
> I was referring to the current Github models repo. Basically, I need to download super old model files to compare with the new model

Yeah, support for that could be added, but it's not currently there. You'd just need to add another function and CLI arg similar to `add_model_from_taskgroup_id`. Generically could this could be `add_model_from_url`.

utils/local_remote_settings/__main__.py
```diff
@@ -0,0 +1,661 @@
...
+ * Click "Sync"
+ * Once you are done, make sure to switch the Environment back to Prod.
+
+Usage:
+
+- Load a model from a task group
+  task local-remote-settings -- --taskgroup_ids I9uKJEPvQd-1zeItJK0cOQ aAVZJcsXQg-vfGIjHmcCTw
+
+- Run a local mirror of the production models
+  task local-remote-settings
```
nordzilla, line 29: > [!NOTE]  
> This example needs to be updated to say
> ```
> task local-remote-settings -- --sync_remote_settings
> ```

utils/local_remote_settings/__main__.py
```diff
@@ -0,0 +1,661 @@
...
+The artifacts will be saved to: ./data/artifacts
+
+To run the local models:
+ * Open Firefox (debug builds don't seem to work)
+ * Run the following from the web console.
+
+    ChromeUtils.importESModule("resource://services-settings/remote-settings.sys.mjs")
+        .RemoteSettings("translations-wasm").verifySignature = false;
+    ChromeUtils.importESModule("resource://services-settings/remote-settings.sys.mjs")
+        .RemoteSettings("translations-models").verifySignature = false;
```
nordzilla, line 13: > praise

This is a lot nicer than having to modify the source code!

Thanks for updating the `--help` menu.

utils/local_remote_settings/__main__.py
```diff
@@ -0,0 +1,661 @@
...
+        )
+        logger.info(f"Attachment added {record.name} {record.version}")
+
+
+# I9uKJEPvQd-1zeItJK0cOQ decoder-base
+# aAVZJcsXQg-vfGIjHmcCTw decoder-depth-3
+# e1DMdEzNSGyGhdjaWFYpxQ decoder-depth-6
+# Nu0YuyBLRuiimYaBKE0RcQ decoder-emb-biger
+# OABpAkBMQvapHE1lNozs0A decoder-ffn-bigger
+# TaeCdUs5Rqq7w1Tbf1PShQ decoder-tiny
```
nordzilla, line 503: > question

Are these leftover debugging comments?

nordzilla (approved):
I tested it out again and it works great!

The experience of the default `--help` menu it is much nicer, compared to the first-pass of review. Thank you! 

There are just a few small things to fix before landing, but non-blocking.

utils/local_remote_settings/__main__.py
gregtatum: Yeah, they were my IDs I was using for testing, thanks!

PR #963, opened 2024-12-17
Do not output empty alignments

eu9ene:
It pollutes the log and I think it also causes issues with the W&B parser


pipeline/train/train.py
```diff
@@ -104,13 +104,16 @@ def build_dataset_tsv(
             empty_alignments = []
 
             for src_line, trg_line, aln_line in zip(src_lines, trg_lines, aln_lines):
-                if not aln_line:
+                if aln_line.strip():
```
gregtatum, line 107: I'm not understanding this patch from the description of the changes compared to the code change. The only real behavior change I'm seeing is that the alignment line is stripped first to see if it's blank. The rest is just code refactoring.

It's also still outputting the a sample of the alignments.

eu9ene: It did the trick. Marian used to complain without outputting the sample, so I assume there are some newlines or something else that made it not empty: https://firefox-ci-tc.services.mozilla.com/tasks/JQuHsJS2R6-RuPR-O9XMXQ/runs/0/logs/public/logs/live.log

PR #962, opened 2024-12-17
Rename "all" to "all-pipeline"

gregtatum:
This fixes the "all" to "all-pipeline" change.

I verified a full training run via: https://firefox-ci-tc.services.mozilla.com/tasks/groups/VKohAC8NTWazaqYK2OnhpA

```
task config-generator -- da en --name all-pipeline-test
task train -- --config configs/autogenerated/da-en-all-pipeline-test.yml --force
```

Two things were likely causing an issue, the prod config needed to be updated, and the valid stages need to have `all-pipeline` and `all-pr-pipeline` added to it.

PR #966, opened 2024-12-18
feat: add cron task that runs the minimal training pipeline nightly

bhearsum:
Relates to https://bugzilla.mozilla.org/show_bug.cgi?id=1937882, where we're adding integration tests for the translations pipeline in https://github.com/mozilla-releng/fxci-config/pull/210. These work by looking a decision task and rerunning the tasks it creates (and those upstream of it). Running the pipeline on a cron schedule allows us to have an easily findable, up to date, decision task to point the tests at.

### DIFF
--- a/.cron.yml
+++ b/.cron.yml
@@ -1,0 +1,13 @@
+jobs:
+  # Run the minimal pipeline once a day to support integration testing before
+  # worker image changes are made (see https://bugzilla.mozilla.org/show_bug.cgi?id=1937882).
+  # These runs will pick up cached tasks, so most of the time this will simply
+  # end up running `all-pipeline`.
+  - name: run-pipeline
+    job:
+      type: decision-task
+      # we don't use treeherder...but this is a required field
+      treeherder-symbol: pipeline
+      target-tasks-method: train-target-tasks
+    when:
+      - {hour: 0, minute: 0}
--- a/taskcluster/translations_taskgraph/parameters.py
+++ b/taskcluster/translations_taskgraph/parameters.py
@@ -2,10 +2,13 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+import logging
 from pathlib import Path
 from taskgraph.parameters import extend_parameters_schema
 from voluptuous import Extra, Optional, Required
 import yaml
+
+logger = logging.getLogger(__name__)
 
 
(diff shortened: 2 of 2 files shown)


bhearsum:
It looks like the defaults updating is causing some issues (either with some test assumptions, or genuine issues); moving back into draft while I work that out.

bhearsum:
> It looks like the defaults updating is causing some issues (either with some test assumptions, or genuine issues); moving back into draft while I work that out.

This turns out not to be necessary, so I removed it! (The defaults end up getting set in the parameters: https://github.com/mozilla/translations/blob/7d45b3a169c275ab051b11d1064bcb5ab1df6622/taskcluster/translations_taskgraph/parameters.py#L105.)

This is ready for review now, but it does have #962 cherry picked to it to ensure the correct kind name is used. This isn't landable until that merges and I rebase on top of it.

bhearsum:
https://github.com/mozilla/translations/pull/966/commits/cbc9e4c9b8d7cdefe2f1a9d80b6d8d1788b9d8d2 is the only part that needs review here.

jcristau (commented):
Kind of wondering if the cron job shouldn't be using an action instead of decision-task.  That'd be a bit more code but wouldn't require the special-casing in get_decision_parameters.

Also, does this really need to run daily (vs weekly for gecko's os-integration cron)?

bhearsum:
> Kind of wondering if the cron job shouldn't be using an action instead of decision-task. That'd be a bit more code but wouldn't require the special-casing in get_decision_parameters.

Hmm...I do recall looking at that, but I'm not sure why that didn't work out. Or I might be confusing that with the abandoned https://github.com/mozilla-releng/fxci-config/pull/192. I'll have a second look at this though; I agree that's cleaner if it works out.

> Also, does this really need to run daily (vs weekly for gecko's os-integration cron)?

I think daily is actually more appropriate for this, because all meaningful tasks will be cached 99% of the time. Daily lets us virtually eliminate the possibility that an image change will run against outdated versions of these tasks.

jcristau (approved):
r+ to make it clear my comment isn't meant to be blocking.

bhearsum:
I gave `trigger-action` another try [over here](https://firefox-ci-tc.services.mozilla.com/tasks/anWHlArDS0STOYnr0rwMYw). It reminded me that, unless I'm missing something, it would require duplicating the configuration into `.cron.yml` or somewhere in fxci-config to have the schema check pass.

PR #968, opened 2024-12-18
Remove expired-after

eu9ene:
Remove 90 days expiration and leave only root level 360 days one. I think it was added by mistake. As a result all our alignments tasks from spring2024 and other experiments have already expired.


eu9ene:
@gregtatum FYI it will be a bit more complicated to retrain the student models in base configuration for spring2024 languages as they use the output of the alignments tasks which are now expired.

bhearsum (approved):

bhearsum:
In the future, perhaps we should avoid landing or cherry picking things like this to `main` (if they come again).

Issue #971, opened 2024-12-20
Combine poetry's utils and test groups


gregtatum:
Now that we have more dependencies across files, imports are easily missing. For instance a missing zstandard import now breaks utils unless you also install the `test` groups. Rather than duplicate the imports between the two groups, I would propose that we combine them together and just have a single general installation group.

Potential negatives are conflicting import requirements and bigger installs in CI.

If we don't want to do this, then at least we should fix it and write a CI test that runs the `--help` of every util to ensure that it's working.

gregtatum:
Looking into this more, it seems like we should have two groups, something like `dev` and `dev-docker`, where the latter would include things like the ICU package, which doesn't build on macOS.

PR #972, opened 2025-01-02
Add pyright coverage for tests/**/*

gregtatum:
I originally just added one file in the test directory that I was touching, but discarded those code changes. I figured I'd keep the typing changes as a PR, and it was quick to just type the rest of the files.


gregtatum:
Hmm... I think this needs to block on #971.

PR #980, opened 2025-01-09
use a hopefully more reliable mtdata dataset in ci config

### LINKED ISSUE #974: elrc-share.net has been down for awhile, breaking our ci pipeline configuration
http://elrc-share.eu/ has been returning 500s for at least the past few days. This leads to 500s whenever an `mtdata-ELRC` task is run, which happens frequently because [we have one such dataset in the CI config](https://github.com/mozilla/translations/blob/2fe2ada160751495fa8d60226f53607ea457f331/taskcluster/configs/config.ci.yml#L80).

I suggest we disable or replace this with another dataset until this server comes back up, as it blocks proper testing of pipeline changes at the moment.

bhearsum:
The current mtdata dataset here has been unavailable for many days in a row, making it difficult to verify pipeline changes.

Should fix #974

### DIFF
--- a/taskcluster/configs/config.ci.yml
+++ b/taskcluster/configs/config.ci.yml
@@ -77,7 +77,7 @@ datasets:
     - opus_ada83/v1
     - opus_ELRC-3075-wikipedia_health/v1
     - url_https://storage.googleapis.com/releng-translations-dev/data/en-ru/pytest-dataset.[LANG].zst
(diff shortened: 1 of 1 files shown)


bhearsum:
Training run with this dataset is going in https://firefox-ci-tc.services.mozilla.com/tasks/groups/WTJ89EYJSEuE9_Dt45HI3Q; so far so good.

eu9ene (approved):

Issue #803, opened 2024-08-19
[meta] Train low resource languages


gregtatum:
After #524 and #425 we can start focusing on lower resource languages. When data is small we'll probably need to focus on different techniques. For instance data synthesizing will be important, with both back translations and enough monolingual data to distill.

We'll probably need to experiment with multilingual models. See #684. We will have to probably re-evaluate our release criteria. Most likely other translation systems will also have lower evaluation scores, and we can compare with our results.

PR #984, opened 2025-01-11
Update model training guide

### LINKED ISSUE #978: Update model training guide

eu9ene:
We have a lot of stuff and it's getting outdated quickly. At the same time, I tried to explain the current practices we use. @ZJaume let me know if you feel like something is lacking here.

closes #978


ZJaume (approved):
I don't feel anything too obvious lacking here. Although I might found something when I start training models. Unless you want me to go through full training until merge this, I think it is sufficient.

docs/training-guide.md
```diff
@@ -8,90 +8,101 @@ nav_order: 2
...
+Explore available supported datasets:
+
+1. Go to [OPUS](https://opus.nlpl.eu/) and see how much parallel data is available for the language pair
+2. Use [find-corpus](https://github.com/mozilla/translations/tree/main/utils/find-corpus.py) tool. 
+Install [poetry](https://python-poetry.org/) and [Taskfile](https://taskfile.dev/installation/) first, then run:
+```bash
+task find-corpus -- ru en
+```
 
+It's hard to say how much parallel data is required to train something useful.
```
gregtatum, line 35: > It's hard to say how much parallel data

This statement is ambiguous. I'd prefer to list our definitions of corpus sizes for small, medium, and large resources. I think this is a more helpful way to frame it. We can list issues as well for supporting low and high resource training. #803. Although I don't see one for high resource. We can also mention that corpus size doesn't guarantee a quality model, which I think what this statement is referring to.

docs/training-guide.md
```diff
@@ -191,6 +265,18 @@ More details:
 
 ### Decoding (translation)
 
+It's used for producing back-translations and for knowledge distillation.
+
+#### Decoder type
+
+Except Marian the pipeline supports CTranslate2 for faster decoding. CTranslate2 does not support ensembles.
```
gregtatum, line 272:
```suggestion
Besides Marian, the pipeline supports CTranslate2 for faster decoding. CTranslate2 does not support ensembles.
```

gregtatum (approved):
Thanks for the updates, this is much clearer for where we are now.

ZJaume:
Should this PR add documentation about the usage of Ctranslate2?

docs/training-guide.md
eu9ene: I replaced it to a table with clarification

PR #988, opened 2025-01-13
Update development docs

### LINKED ISSUE #979: Update development docs
Specifically how to use Docker and what's our regular pipeline to run tests etc.

eu9ene:
I updated the things that I use. Let me know if you think something else should be added.

closes #979


docs/development.md
```diff
@@ -6,57 +6,105 @@ nav_order: 8
...
 We do not run the pipeline on branches without a corresponding pull request.
 
+## Architecture
+
+All the pipeline steps are independent and contain scripts that accept arguments, read input files from disk and output the results to disk.
+It allows writing the steps in any language (currently it's historically mostly bash and Python) and
+represent the pipeline as a directed acyclic graph (DAG).
+
+The DAG of tasks can be launched using any workflow manager
+(currently we support [Snakemake](snakemake.md) and [Taskcluster](task-cluster.md)).
```
gregtatum, line 92: I'm not sure we should claim to actually support Snakemake. Maybe mention that is unmaintained and we will accept contributions to fix the support.

gregtatum (approved):

docs/development.md
eu9ene: Agreed, it's misleading

ZJaume (approved):

PR #991, opened 2025-01-15
Add monocleaner

### LINKED ISSUE #247: Consider using monocleaner for cleaning monolingual corpuses
https://github.com/bitextor/monocleaner/

### LINKED ISSUE #476: Investigate monolingual cleaning
Based on @marco-c's feedback we should investigate how the HPLT project cleans monolingual data and whether we should adjust our cleaning procedure.

https://hplt-project.org/HPLT_D3_1___Software_for_cleaning_data_sets.pdf

### LINKED ISSUE #789: Filter monolingual data based on fluency scores
The HPLT dataset includes a fluency score. We should look at filtering our own data by this fluency metric, and see if it improves.

https://hplt-project.org/datasets/v1.2

I assume this would be useful for synthesizing back translations, and less useful for synthesizing distillation data.

https://aclanthology.org/2024.lrec-main.100.pdf

> fluency score, computed with a 7-gram modified Knesser-Ney character language model

### LINKED ISSUE #985: Add fluency scoring + filtering
So, I've been trying to add myself the necessary code to do this, but I haven't been able to set up a local taskcluster and write configuration for one or two new taskcluster tasks. Maybe it is quicker if I just write here what is needed and @eu9ene can write the kind.yml?

The simplest solution might be to create a new step to download the monocleaner model for each language and add the filtering to `clean-mono` in a similar way to the `langid` filter. To do this, I guess the download step should be something like `bicleaner-model` kind but called `monocleaner-model` and do one task for source and other for target, like `clean-mono`. The installation could be handled like bicleaner because monocleaner also needs kenlm and cyhunspell. So, the `monocleaner-model` kind could just contain the same commands but calling
```
monocleaner-download $TASK_WORKDIR/artifacts/monocleaner-{locale}
```
instead of `download_pack.py`. But this will [download and decompress](https://github.com/bitextor/monocleaner/blob/main/scripts/monocleaner-download) the tarball, so if you want to handle it in a different way, the download urls follow this pattern `https://github.com/bitextor/monocleaner/blob/main/scripts/monocleaner-download/$lang.tgz`.

Then to add an additional step to `clean-mono.sh`, could be like this after the langid or rule filtering:
```bash
echo "### Filter by fluency score"
test -s "${output_prefix}.${lang}.fluency.zst" ||
  zstd -dc "${output_prefix}.${lang}.langid.zst" |
  # memory intensive
  #TODO add monocleaner model path
  parallel --no-notice --pipe -k -j "$(echo "${threads}"/4 | bc)" --block 50M "monocleaner --disable_hardrules $monocleaner_models/$lang" |
  #TODO needs parameter fluency score threshold, this hardcodes to 0.7
  awk -F'\t' '$2>0.7' | cut -f1 |
  zstdmt >"${output_prefix}.${lang}.fluency.zst"

######################################################################
echo "### Rule-based filtering"

zstdmt -dc "${output_prefix}.${lang}.fluency.zst" |
...
```

eu9ene:
- Add [Monocleaner](https://github.com/bitextor/monocleaner/tree/main) to the cleaning pipeline
- Fix issues with dataset names in cleaning configs

Let's start testing it first, then we can update docs with recommendations based on results.

closes #247 
closes #476 
closes #985 
closes #789


pipeline/clean/clean-mono.sh
```diff
@@ -74,14 +75,36 @@ zstdmt -dc "${output_prefix}.${lang}.langid.zst" |
...
+test -s "${output_prefix}.${lang}.rule-based.zst" || exit 1
+
+######################################################################
+echo "### Filter by fluency score"
+
+if [ "${fluency_threshold}" == "0" ] || [ "${fluency_threshold}" == "0.0" ]; then
+  echo "Threshold is 0, skipping filtering"
+  cp "${output_prefix}.${lang}.rule-based.zst" "${output_prefix}.${lang}.zst"
+else
+  # the model is 125MB, similar in size to the fastText one, so it's ok to download it here
```
eu9ene, line 89: @bhearsum do you think it's still worth extracting this download to a separate task? I see the models are 100-200Mb, similar to the fastText ones. Extracting would save on some ingress traffic, but would complicate the pipeline and model updates, similar to the bilceaner ones. The bicleaner models are a lot larger, so it's justified to cache them.

bhearsum: I have no objections. IMO the more important downside is that not caching these means that you'll be more susceptible to failures when wherever is hosting them is down.

(And I could be wrong, but it looks like [ingress is free](https://cloud.google.com/network-tiers/pricing))

eu9ene: Ok, I'll skip it for now then. Let's if we have issues with that. Github is hosting them, so I guess it should be reliable.

https://github.com/bitextor/monocleaner/blob/5a743407179468c83259999987138d6ae590c687/scripts/monocleaner-download#L32

ZJaume (approved):

taskcluster/kinds/bicleaner/kind.yml
```diff
@@ -33,7 +33,7 @@ tasks:
                     - pipeline/bicleaner/requirements/bicleaner-ai.txt
                 from-parameters:
                     bicleaner_threshold:
-                        - training_config.experiment.bicleaner.dataset-thresholds.{dataset}
+                        - training_config.experiment.bicleaner.dataset-thresholds.{provider}_{dataset_sanitized}
```
bhearsum, line 36: Were we ignoring bicleaner thresholds (other than the default) because of this previously? (Based on the commit that added this it looks like maybe not? I just wanted to double check...)

bhearsum (approved):
This is fine from a taskgraph standpoint. Nice work!

taskcluster/kinds/bicleaner/kind.yml
eu9ene: Only for the datasets with "." in the names, but we didn't use the thresholds for such datasets, so no.

PR #999, opened 2025-01-20
WIP: try generic worker with podman instead of docker for container engine

bhearsum:
This is a variant of #700 to see if podman works without issues with caches after some recent fixes.


bhearsum:
Podman has had many issues (both here and in other places we use it); we're definitely not going this route.

PR #1001, opened 2025-01-21
OpusCleaner fixes for Japanese and Korean

ZJaume:
[skip ci]


eu9ene:
@ZJaume I'm using this in my training. Is it ready to be merged?

pipeline/clean/opuscleaner/configs/en-ja/default.filters.json
```diff
@@ -30,6 +30,69 @@
       },
       "language": null
     },
+    {
```
eu9ene, line 33: I didn't notice before, but why don't we filter those for the opposite direction?

ZJaume: My guess is we want the model to be able to translate regardless of the input punctuation. e.g. If we train only with fixed size punctuation, the model might struggle with website that might have ascii punctuation.

eu9ene (approved):

eu9ene:
@ZJaume feel free to merge this one, but I think you'll need to update the commit history first.

PR #1018, opened 2025-01-28
Add CJK one stage config

eu9ene:
I'm using it now to retrain teacher models. The two-stage approach stops training too early, similar to what I had when training en-ru. Probably there are issues in parallel data.


pipeline/train/configs/opustrainer/teacher.one-stage.cjk.yml
```diff
@@ -0,0 +1,37 @@
+datasets:
```
gregtatum, line 1: This could use a few sentences of documentation at the top to explain why this config exists, and why you would use it compared to other configs.

gregtatum (approved):

PR #1020, opened 2025-01-29
Default configuration updates

eu9ene:
Various small fixes that I've been doing in configs for my training.

The default config should be as close as possible to what we want but it still depends on language and requires manual inspection.

### DIFF
--- a/pipeline/translate/decoder.yml
+++ b/pipeline/translate/decoder.yml
@@ -9,6 +9,14 @@ mini-batch-words: 5000
 maxi-batch: 10000
 maxi-batch-sort: src
 
+# Make sure the GPU model supports half-precision decoding
+# This flag works for both Marian and CTranslate2
+precision: float16
+# This Marian flag appears to be slightly different than "precision: float16":
+#       "Shortcut for mixed precision training with float16 and cost-scaling, "
+#       "corresponds to: --precision float16 float32 --cost-scaling 8.f 10000 1.f 8.f"
+fp16: true
+
 max-length: 200
 max-length-crop: true
 beam-size: 8
--- a/pipeline/translate/decoder.yml
+++ b/pipeline/translate/decoder.yml
@@ -9,12 +9,8 @@ mini-batch-words: 5000
 maxi-batch: 10000
 maxi-batch-sort: src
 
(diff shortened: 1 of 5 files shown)


pipeline/translate/decoder.yml
```diff
@@ -8,6 +8,13 @@ word-penalty: 0
 mini-batch-words: 5000
 maxi-batch: 10000
 maxi-batch-sort: src
+
+# Make sure the GPU model supports half-precision decoding
+# This flag works for both Marian and CTranslate2
+precision: float16
```
eu9ene, line 14: Do we have a consensus that half-precision decoding is safe to use for both student and s2s backward-models using both Marinan and CTranslate2?

eu9ene: Based on the current config it was enabled for everything except CTranslate2 backward

ZJaume: I would just set `fp16: true`.

ZJaume (approved):

ZJaume (changes requested):
Remove `precision` and use only `fp16: true`

eu9ene:
While looking at CTranslate2 precision settings, I wondered whether quantizing to fp16 is the same thing as happens in Marian and it seems it's not. There is a separate option in [Translator](https://opennmt.net/CTranslate2/python/ctranslate2.Translator.html?highlight=float16) `compute_type: float16` that enables half-precision decoding on a GPU level. I don't know which option is better and whether there will be any difference in performance or quality. @gregtatum Did you consider this option?

pipeline/translate/decoder.yml
eu9ene: Done, but it seems it's the same thing for decoding based on Marian's docs. fp16 enables several options only in training mode.

ZJaume (approved):

gregtatum (approved):
I don't really remember looking into the precision settings more than carrying over the precision setting. I'm a bit nervous on the CTranslate2 step missing the `fp16: true` setting and getting a surprise at some point in the future.

It might be worth looking into the settings a bit more to make sure we are doing the right thing.

PR #1021, opened 2025-01-30
[Experiments] Arabic

ZJaume:
[skip ci]

### DIFF
--- a/configs/autogenerated/ar-en-experiments-ar-v2.yml
+++ b/configs/autogenerated/ar-en-experiments-ar-v2.yml
@@ -1,0 +1,172 @@
+experiment:
+  name: experiments-ar-v2
+  src: ar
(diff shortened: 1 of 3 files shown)


eu9ene:
@ZJaume I'm curious why you didn't use Marian's `--right-left` option. The quality is still good, so it's fine. And also it means we don't need to do anything on the inference side.

ZJaume:
I think that option was meant to be for other purposes (such as translation candidates re-ranking with a r2l that processes tokens backwards). For Arabic, even if text needs to be displayed r2l, I guess the model without r2l option is still reading the Arabic characters in the same order as they are read because that is how they are appear in the txt files.

evgenyrp:
@ZJaume you can close it now if you don't need this branch

PR #1032, opened 2025-02-11
Add a static site directory ./site and switch docs to mkdocs

gregtatum:
This will enable migrating over my Taskcluster Tools. The `./site` will serve as a static site that will be deployed to GitHub pages via a GitHub action. This replaces the existing (built-in) Jekyll GitHub action. Adding @bhearsum to the review since it's not using a Taskcluster task but a GitHub action and to double check that this approach is OK.

I styled the landing page and docs after https://www.mozilla.org/en-US/firefox/new/

The docs I applied some light style and JavaScript corrections to the theme to make it match. I also reorganized things a bit to fit the expectations for nested menus rather than explicitly setting the order of the nav bar.

For a preview see: https://gregtatum.github.io/translations/

Screenshot of the GitHub page


.github/workflows/github-pages.yml
```diff
@@ -0,0 +1,47 @@
+name: Deploy Github Pages
+
+on:
+  push:
+    branches:
+      - mkdocs
```
bhearsum, line 6: Is this intentional or, eg: a holdover from testing? `main` and `mkdocs` seem like they may eventually get out of sync if you're deploying from this branch. Totally up to you though!

bhearsum (approved):
This seems sane and fine to me, modulo the one comment below.

PR #1043, opened 2025-02-24
force translate tasks to run to see if they encoder gpustat installation issue

bhearsum:
I'm seeing this in #700; it's unclear to me whether it's related to the new worker images or not. Let's run those tasks on the existing workers with code that's otherwise the same to see if we hit it here.


bhearsum:
They do! This is because of https://github.com/pypa/setuptools-scm/issues/1112. I'll open a cleaner PR with a fix when I find one.

PR #1045, opened 2025-02-24
fix: gpustat installation in translate-mono tasks

bhearsum:
Due to the issue described in pypa/setuptools-scm#1112, gpustat is no longer installable on ubuntu 22.04 (which we use on both cpu and gpu workers). If setuptools-scm lowers their setuptools requirement to ~59, or gpustat pins its build requirements to an older version of setuptools-scm, this will fix the issue, and we can revert this patch.


bhearsum:
see also https://github.com/mozilla/translations/issues/1044, where I'm tracking reverting this when we can.

gregtatum:
Thanks for looking into this and providing a quick fix.

PR #1046, opened 2025-02-25
Stub out an initial model registry page

gregtatum:
Edit: I pushed an update that makes it so that all the generated artifacts are uploaded to GCS rather than output into the repo. You can still generate and inspect everything locally in the `./data` folder for debugging scripts.

---
This is kind of a big PR, but it stubs out some initial model registry infrastructure. The goal here is to get something together so that I can easily kick off the next big training run. I initially just built a script that would generate the necessary JSON for the registry to correlate all of the available information.

However, after I did that, it wasn't too much work to build a nice UI on top of it with our new static site infrastructure. Initially I'd like to check in all of the JSON information into our repo, and afterwards come to a consensus on how we want to store and work with this information.

There are 3 big commits, with 1 of them being the auto-generated JSON, so shouldn't be a big review burden. There are a few minor commits as well which should be quick to review.

The first big commit is the model_registry.py which builds the JSON. I ran into some issues installing dependencies with the google storage's protobuf version not being compatible with our existing dependencies. I decided not to solve that here, and include some short-term manual steps for install dependencies. I'll file a follow-up issue after review and before merging.

The next big reviewable commit is the model registry site. It's kind of a big thing, and the risk for breakage is low since we'll be the ones to view it. I did notice that the TypeScript from the original patch for integrating the Firefox models wasn't actually being checked.

If there are no big disagreements on approach, my general preference on review here is to follow-up on functionality and aggressively file issues on how to improve things rather than rework it all in this bigger PR. I feel the risk is low as it's just an internal tool, and an attempt to understand how are models are stored across things.

screenshot of the model registry

screenshot of a model card

## TODO Before Merging

 - [ ] File google-cloud-storage install issue
 - [ ] File issue for checking TypeScript in CI
 - [ ] File issue for hooking up eslint

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -288,9 +288,8 @@ tasks:
 
   serve-site:
     desc: Serve the GitHub Pages website locally
-    deps: [build-docs]
     cmds:
-      - poetry run python3 -m http.server 8080 --directory site
+      - poetry run python utils/site_server.py
 
   preflight-check:
     desc: Perform pre-flight checks for a training run.
--- a/site/@types/globals.d.ts
+++ b/site/@types/globals.d.ts
@@ -1,1 +1,10 @@
-// Intentionally left blank for future migrated code.
+
+export interface CreateElementOptions<T> {
+  style: Partial<CSSStyleDeclaration>,
+  parent: Element,
+  children: Node | string | Array<string | Node> | [],
+  href: string,
+  className: string,
+  title: string,
+  onClick:(this: HTMLElement, event: MouseEvent) => unknown
+}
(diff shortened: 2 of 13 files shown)


gregtatum:
I'm moving to draft since I just tried moving all the generated artifacts to GCS and it works great, so I think that's a better path forward. I'll rework this a bit. The downside is that the artifacts aren't versioned, which is a bit risky, but it'll probably easier than checking in code all time.

gregtatum:
I have a live copy of the site here:

https://gregtatum.github.io/translations/model-registry/

eu9ene (approved):
I'm approving based on the doc review but here are some suggestions for the web site: 
- show absolute COMET numbers as well as the percentage or absolute differences with Google (similar to https://mozilla.github.io/translations/firefox-models/ and the training dashboard)
- use a different style for clickable elements like model scores, otherwise it's not clear that they are clickable (I assume underline)
- sort experiments by date by default to have the most recent ones on the top

PR #1049, opened 2025-03-03
Do not convert Chinese Traditional to Simplified for the target language

### LINKED ISSUE #1033: Improve translation quality of English to Chinese
The quality of the teacher model doesn't look great: COMET 85.67 vs Google Translate 89.45 (-4.41%) on flores-test. 

I suspect we have some issues in the training data:
1. I found that we used dataset `mtdata_Statmt-backtrans_enzh-wmt20-eng-zho` which is backtranslations for zh-en WMT20 task. So the zh part of it might be of poor quality. We already have our own back-translations.
2. We convert all the data from Traditional to Simplified with hanzi-convert. People on the internet say such conversion is imprecise due to the regional differences, vocabulary, word composition etc., so it is not equal to a proper translation. It might be the case that it pollutes data when Chinese is the target language. We can just filter all the data in Traditional and see if it makes a difference.
3. We use a 64k joint vocabulary for very different scripts, it's worth running an experiment with separate vocabularies 32k in size. Depends on https://github.com/mozilla/translations/issues/913
4. There are known issues for some datasets: [UN](https://github.com/mozilla/translations/issues/744), [HPLT](https://github.com/mozilla/translations/issues/1030), [desegmentation](https://github.com/mozilla/translations/issues/743). We can address them with ad-hoc fixes through OpusCleaner and the mono-cleaning scripts.

The question is which of those issues contribute most. It would require running separate experiments for most of them to see if it moves the needle.

Another indicator that something is wrong with the data is that one of my experiments completely diverged and I had to reduce the learning rate. Also when training in two stages there is a big bump in the cost when moving to training on the original parallel corpus only. This typically indicates that the fine-tuning data is noisier compared to the pre-training one (a mix of back-translated and original corpus).

Image

eu9ene:
closes #1033

### DIFF
--- a/pipeline/data/cjk.py
+++ b/pipeline/data/cjk.py
@@ -26,10 +26,13 @@ class ConversionStep(Statistics):
     When converting data, count how many sentences were converted, and how many were visited.
     """
 
-    def __init__(self, description: str, converted=0, dataset_path: Optional[Path] = None) -> None:
+    def __init__(
+        self, description: str, converted=0, filtered=0, dataset_path: Optional[Path] = None
+    ) -> None:
         super().__init__(dataset_path)
         self.description = description
         self.converted = converted
+        self.filtered = filtered
         self.visited = 0
 
 
@@ -50,6 +53,9 @@ class ChineseConverter:
     def convert_file(
         self, input_path: Path, output_path: Path, to: ChineseType
     ) -> DatasetStatistics:
+        """
+        Convert all lines to one variant of Chinese
+        """
         stats = DatasetStatistics(output_path, to)
         with write_lines(output_path) as out_file, read_lines(input_path) as lines:
             for line in lines:
@@ -61,6 +67,22 @@ class ChineseConverter:
                     new_line = self._convert_line(line, to)
                     stats.script_conversion.converted += 1
                 out_file.write(new_line)
+        return stats
+
+    def filter_file(self, input_path: Path, output_path: Path, variant: ChineseType):
+        """
+        Filter everything except the specified variant of Chinese
+        """
+        stats = DatasetStatistics(output_path, variant)
+        with write_lines(output_path) as out_file, read_lines(input_path) as lines:
+            for line in lines:
+                stats.script_conversion.visited += 1
+                ch_type = self._detect(line)
+                if ch_type == variant:
+                    out_file.write(line)
+                else:
+                    stats.script_conversion.filtered += 1
+
         return stats
 
     @staticmethod
--- a/pipeline/data/cjk.py
+++ b/pipeline/data/cjk.py
@@ -1,7 +1,6 @@
 """
 Chinese, Japanese, Korean (CJK) specific data importing code
 """
-import shutil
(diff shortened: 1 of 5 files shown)


pipeline/data/download-mono.py
```diff
@@ -133,20 +133,37 @@ def main(args_list: Optional[list[str]] = None) -> None:
...
-        converted_path = file_destination.with_suffix(".converted.zst")
-        stats = chinese_converter.convert_file(
-            file_destination, converted_path, ChineseType.simplified
-        )
-        shutil.move(converted_path, file_destination)
-        print(
-            f"Converted {stats.script_conversion.converted} lines from {stats.script_conversion.visited} to Chinese Simplified"
-        )
-        stats.save_json()
+        if os.environ["SRC"] == "zh":
```
gregtatum: Can we pass this in as an argument in argparse rather than rely on the environment variables? I'd like to not rely on two API surfaces for our python scripts, as it's confusing on what the inputs are.

pipeline/data/download-mono.py
```diff
@@ -133,20 +133,37 @@ def main(args_list: Optional[list[str]] = None) -> None:
...
+            logger.info("Converting the output file to Chinese Simplified")
+            chinese_converter = ChineseConverter()
+            converted_path = file_destination.with_suffix(".converted.zst")
+            stats = chinese_converter.convert_file(
+                file_destination, converted_path, ChineseType.simplified
+            )
+            shutil.move(converted_path, file_destination)
```
gregtatum: I guess this move overwrites the file_destination file, so we don't retain extra copies of the datasets. It might be more explicit to remove the old one and then move so it's clearer. I had to look up this API.

pipeline/data/cjk.py
```diff
@@ -102,3 +134,50 @@ def _convert_line(self, text: str, to: ChineseType) -> str:
...
+def handle_chinese_mono(file_destination: Path, is_src: bool, variant: ChineseType):
+    converted_path = file_destination.with_suffix(".converted.zst")
+    chinese_converter = ChineseConverter()
+    if is_src:
+        logger.info(f"Converting the output file to {variant}")
+        stats = chinese_converter.convert_file(file_destination, converted_path, variant)
+    else:
+        logger.info(f"Filtering out everything except {variant} in the output file")
+        stats = chinese_converter.filter_file(file_destination, converted_path, variant)
+    shutil.move(converted_path, file_destination)
```
gregtatum, line 148: Stylistically, since we're using `pathlib`, I'd prefer to use its methods like `.rename` and `.unlink` rather than using multiple libraries like `shutil`, unless there is a good reason to do so. I generally prefer `pathlib` over other file and file path manipulation tools, as it's pretty ergonomic to work with.

gregtatum (approved):

pipeline/data/cjk.py
eu9ene: `.replace` similar to `.move` does it in one line

bhearsum (approved):

Issue #891, opened 2024-10-21
[meta] Retrain older high resource models


evgenyrp:
A meta-issue to track retraining of the older models like Italian, Portuguese, French, German, Spanish etc.

We'll need to retrain them to incorporate the latest robustness fixes. Also open-source datasets have improved since we trained them. 

Those are high-resource languages with massive amounts of available training data, so training them is a bit more expensive than a typical language from our last big training. We'll need to do the optimizations first and also fix as many quality issues as we can before retraining not to do this again in the near future.

### Must have
- [x] https://github.com/mozilla/translations/issues/1040
- [x] https://github.com/mozilla/translations/issues/942
- [x] https://github.com/mozilla/translations/issues/76
- [x] https://github.com/mozilla/translations/issues/651
- [x] https://github.com/mozilla/translations/issues/631
- [x] https://github.com/mozilla/translations/issues/1025
- [x] https://github.com/mozilla/translations/issues/711
- [x] https://github.com/mozilla/translations/issues/774
- [x] https://github.com/mozilla/translations/issues/758
- [x] https://github.com/mozilla/translations/issues/177
- [x] https://github.com/mozilla/translations/issues/870
- [x] https://github.com/mozilla/translations/issues/928
- [x] https://github.com/mozilla/translations/issues/903
- [x] https://github.com/mozilla/firefox-translations-training/issues/215
- [ ] https://github.com/mozilla/translations/issues/248

### Nice to have
- [x] https://github.com/mozilla/translations/issues/728
- [x] https://github.com/mozilla/translations/issues/466
- [x] https://github.com/mozilla/translations/issues/736
- [x] https://github.com/mozilla/translations/issues/905
- [x] https://github.com/mozilla/translations/issues/910
- [x] https://github.com/mozilla/translations/issues/783
- [ ] https://github.com/mozilla/translations/issues/940
- [ ] https://github.com/mozilla/translations/issues/634
- [ ] https://github.com/mozilla/translations/issues/394
- [ ] https://github.com/mozilla/translations/issues/987
- [ ] https://github.com/mozilla/translations/issues/92
- [ ] https://github.com/mozilla/translations/issues/1038
- [ ] https://github.com/mozilla/translations/issues/89
- [ ] https://github.com/mozilla/translations/issues/414
- [ ] https://github.com/mozilla/translations/issues/274
- [ ] https://github.com/mozilla/translations/issues/577
- [x] https://github.com/mozilla/translations/issues/800
[...]

PR #1054, opened 2025-03-06
[Experiment] Retrain de-en [skip ci]

eu9ene:
Run the first high-resource language pair to test pipeline scalability before moving forward with #891.

[skip ci]

### DIFF
--- a/configs/autogenerated/de-en-retrain_hr.yml
+++ b/configs/autogenerated/de-en-retrain_hr.yml
@@ -1,0 +1,411 @@
+# The initial configuration was generated using:
+# task config-generator -- de en --name retrain_hr
+#
+# The documentation for this config can be found here:
+# https://github.com/mozilla/translations/blob/ab61ad90e0b3511ca1417cf448dc6e3faf80c732/taskcluster/configs/config.prod.yml
+experiment:
+  name: retrain_hr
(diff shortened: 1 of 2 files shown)


bhearsum (approved):
I'm not qualified to review the ce-filter changes, but the taskgraph parts are fine!

taskcluster/kinds/backtranslations-train-backwards-model/kind.yml
```diff
@@ -72,7 +72,7 @@ tasks:
                 # that doesn't requires as much disk as full fledged runs
                 github-pull-request: b-largegpu
                 cron: b-largegpu
-                default: b-largegpu-largedisk
+                default: b-largegpu-xlargedisk
```
gregtatum, line 75: So this will increase the cost for even low resource languages. Could we consider filing an issue for making some of these machine decisions dynamic based on resource size? We estimate the size of the corpus during config generation time, so I'm pretty sure we could write a transform to read the config to determine if the corpus is `high` or `low` or something, and adjust sizes up or down.

evgenyrp: https://github.com/mozilla/translations/issues/1158

PR #1056, opened 2025-03-12
Rework the configs to be type checked with dataclasses

gregtatum:
I'm working on corpus continuation, and want to have more assurances over manipulating the config. This PR adds type checking to the config via dataclasses. I ended up writing some custom logic around serialization/deserialization to handle kebab casing. I also wrote converters to both the json schema and the voluptuous type schema as both were previously used for validation steps.

I also added validation steps for all of the existing configs that we have in the repo so that we can trust that things are up to date as we merge things in. This means that our version controlled configs will be validated in CI.

I also checked in a generated config to run validate on. It might be nice to generate one every time in the test, but that's out of scope here since config generation requires network access of fetching datasets, and this would need to be mocked out to support.

As usual, the commits are well-ordered and split up into smaller reviewable chunks.


docs/training/task-cluster.md
```diff
@@ -102,12 +102,12 @@ tasks:
 
 ## Running only later parts of the pipeline
 
-When hacking on later parts of the pipeline it can often be useful to re-use earlier runs of the pipeline, even if those runs were done with different training parameters. To do this, we must bypass the usual caching mechanisms of Taskgraph, and force it to replace earlier tasks with ones we provide. To do this, you can run a training action as usual, but also provide `start-stage` and `previous_group_ids` parameters. For example:
```
gregtatum, line 105: This is the one part of the config that used underscore casing rather than kebab casing, and it didn't match the rest of the keys.

taskcluster/translations_taskgraph/util/dataclass_helpers.py
```diff
@@ -0,0 +1,426 @@
+"""
```
gregtatum, line 1: This section is probably the most complicated part of this PR, as it's walking the types and handling some custom logic. This top comment should hopefully explain what's going on.

gregtatum:
Since no one has looked at this yet, I'm going to mark as draft. I'm struggling a bit with actually using this in a useful way.

gregtatum:
I'm closing this and my revisit if we can upgrade to Python 3.11 and support for the Required and NotRequired in the TypedDict. I also may pull out some of the fixes here into separate PRs.

PR #1059, opened 2025-03-18
Use Standard Language Display for Languages

nordzilla:
### Description

Switches the language DisplayNames config to use `"standard"` for the language display.

This will display "Chinese (Simplfied)" instead of "Simplified Chinese", which is nicer since our sorted order is by display name.

---

#### Before

image

#### After

image


gregtatum (approved):
Thanks!

PR #1060, opened 2025-03-20
Update the task documentation

gregtatum:
This PR updates the task documentation by filling out the description fields of all of the tasks in the kind.yml files. Previously this field was mostly copying the label field and not providing any contextual information. Working on corpus continuation I was getting lost in these files trying to understand why they existed and where they fit into the pipeline. The generated documentation should be built by the normal docs building step, and we can verify after it's merged in. [I copied the generated file to this gist](https://gist.github.com/gregtatum/c633f5c611c245869038e5ab025edc0c).

I also drew out the [full updated DAG of the pipeline](https://docs.google.com/presentation/d/1HkypImI_hbA3n1ljU57ZPAzW8PuQqdv2wrXqj688KtQ/edit?slide=id.p#slide=id.p) and included it in the docs here.


gregtatum:
Looks like CI failed, I'll have a look.

taskcluster/kinds/evaluate/kind.yml
```diff
@@ -112,7 +112,9 @@ task-defaults:
 
 tasks:
     backward-{provider}-{dataset_sanitized}-{src_locale}-{trg_locale}:
-        description: backwards evaluation for {dataset_sanitized} {src_locale}-{trg_locale}
+        description: >
+            Evaluate the backwards model using a devset. This generates a
+            COMET, bleu, and chrF score.
```
eu9ene, line 117: We might add more metrics and this will become outdated

taskcluster/kinds/export/kind.yml
```diff
@@ -20,7 +20,12 @@ kind-dependencies:
 
 tasks:
     "{src_locale}-{trg_locale}":
-        description: export for {src_locale}-{trg_locale}
+        description: >
+            Export the final assets for the trained model. This includes the lexical
+            shortlist (lex.50.50.{src_locale}{trg_locale}.s2t.bin.gz), the model
+            (model.{src_locale}{trg_locale}.intgemm.alphas.bin.gz), and the
+            SentencePiece vocab (vocab.{src_locale}{trg_locale}.spm.gz). This is the
```
eu9ene, line 27: it can be a different format for split vocabs

README.md
```diff
@@ -48,6 +48,7 @@ Feel free to ask questions in our Matrix channel [#firefoxtranslations:mozilla.o
 - [Reference papers](docs/README.md#references)
 - [Model training guide](docs/training/README.md) - practical advice on how to use the pipeline
 - [High level overview post on Mozilla Hacks](https://hacks.mozilla.org/2022/06/training-efficient-neural-network-models-for-firefox-translations/)
+- [The Training Pipeline DAG](https://docs.google.com/presentation/d/1HkypImI_hbA3n1ljU57ZPAzW8PuQqdv2wrXqj688KtQ/edit?slide=id.g3421e8f521e_1_419#slide=id.g3421e8f521e_1_419)
```
eu9ene, line 51: Is it possible to download this as an image and include it in the repo?

eu9ene (approved):
lgtm overall, thanks for adding more docs! There are some minor comments. Some docs are fragile as they describe implementation details that typically change.

It slightly duplicates https://mozilla.github.io/translations/docs/training/pipeline-steps/, so maybe we can remove this one.

README.md
gregtatum: Hmm... I'm a bit hesitant to include it, as we'll need to update it in our drive, and update it in the docs, which is a bit more work. This way it points to the live document which the team has access to edit.

Issue #743, opened 2024-07-16
Support dataset desegmentation for CJK


eu9ene:

eu9ene:
@ZJaume do I understand correctly that your fixes addressed only the punctuation issues? If yes, the issue with segmented corpus might still be present. I wonder how much it can affect quality.

ZJaume:
Yes, the fixes only addressed punctuation issues. I also wonder how segmented corpora affect quality.

PR #1061, opened 2025-03-24
en-zh-bible-uedin desegmentation fix

ZJaume:
This does not add desegmentation support but fixes the only corpus I could find with that issue in Chinese, so I guess  #743 can be closed?


ZJaume:
Completely forgot about the format differences. I hope now is ok.

eu9ene (approved):

PR #1066, opened 2025-03-31
(wip) Corpus continuation

gregtatum:


gregtatum:
Successful real runs:

https://gregtatum.github.io/taskcluster-tools/src/training/?taskGroupIds=Bz4uclzcRBO2-DzC01UGHA%2CFMlwCeLOQsyGYOA2e2q4Ag%2CJfFcCrayTmu8xFRTYy7Z9g%2CPEUtQa_1Smq72d9lInzjUQ%2CBMXqL_n_R4OCCXE8s1w3Pw&showAll=false&hidden=LGELYIbhR3mFh_ydAVm4yw%2CFthyEF4lRTubGVfmW350RA

pipeline/alignments/align.py
```diff
@@ -438,14 +438,21 @@ def main() -> None:
     )
     args = parser.parse_args()
     logger.info("Starting generating alignments.")
+
+    priors_input_path = args.priors_input_path
+    if priors_input_path and not Path(priors_input_path).exists():
+        # This can happen on training continuation.
+        print("The priors were not found, they will be regenerated.")
+        priors_input_path = None
```
evgenyrp, line 448: As discussed, we should test how it works on high-resource languages later. Using priors added some kind of optimization to memory usage.

docs/training/using-pretrained-models.md
```diff
@@ -76,3 +75,95 @@ In `init` mode, the pipeline initializes model weights with the downloaded model
...
+      trg: https://example.com/backtranslations.en.zst
+      # Optional:
+      tok-src: https://example.com/backtranslations.tok-icu.ru.zst
+      tok-trg: https://example.com/backtranslations.tok-icu.en.zst
+      alignments: https://example.com/backtranslations.aln.zst
+    parallel:
+      src: https://example.com/parallel.ru.zst
+      trg: https://example.com/parallel.en.zst
+      # Optional:
+      tok-src: https://example.com/parallel.tok-icu.ru.zst
```
evgenyrp, line 145: If we have the tokenized version, the corpus becomes optional, I think.

pipeline/alignments/align.py
evgenyrp: Maybe it's worth adding a comment or todo on what we plan to do about priors in the future.

pipeline/continuation/model.py
```diff
@@ -0,0 +1,123 @@
+"""
+Continue the training pipeline with an existing model.
```
evgenyrp: Hm, it's not immediately clear how this code plays together with the old training continuation code... It seems it will replace the train-* steps in the graph by this one with downloading the models. Do we need to keep the old transform that adds mounts then?

pipeline/continuation/model.py
```diff
@@ -0,0 +1,123 @@
...
+    "model.npz",
+    "model.npz.best-bleu-detok.npz",
+    "model.npz.best-ce-mean-words.npz",
+    "final.model.npz.best-chrf.npz",
+    "model.npz.optimizer.npz",
+)
+
+potential_decoders = [
+    "model.npz.best-chrf.npz.decoder.yml",
+    "model.npz.best-bleu-detok.npz.decoder.yml",
```
evgenyrp, line 30: I don't think we need all of these. If we want to continue training with the same corpus, we need only `model.npz.yml` and I think `model.npz.optimizer.npz`. If we want to initialize to fine-tune we need `final.model.npz.best-chrf.npz`. And for "use" mode it's two `final*` files (we should double check this). It probably makes sense to copy everything for preemption but it's a different mechanism. Here, when we just want to replace model training with getting the final artifacts, we need only `final*` files.

pipeline/continuation/model.py
```diff
@@ -0,0 +1,123 @@
...
+        url = url_prefix + potential_decoder
+        logger.info(f"Checking to see if a decoder.yml exists: {potential_decoder}")
+        if location_exists(url):
+            logger.info(f"Downloading it to: {decoder_out}")
+            stream_download_to_file(url, decoder_out)
+            decoder_found = True
+            break
+    assert decoder_found
+
+    # Prefer the vocab near the model.
```
evgenyrp, line 105: That's correct. I'd say we need a validation that the vocab we use for the student is the same one as comes with the teacher, because having the same tokenizer is important for distillation.

taskcluster/kinds/continuation-model/kind.yml
```diff
@@ -0,0 +1,94 @@
...
+        provided to the config via "continuation.models.{model}" with the "use" mode. The
+        transform "translations_taskgraph.transforms.continuation" will rewrite the
+        dependencies for all tasks to rely on these existing models.
+    attributes:
+        src_locale: "{src_locale}"
+        trg_locale: "{trg_locale}"
+        stage: continuation-model
+        cache:
+            type: continuation-model
+            version: 2 # Bump the cache version to manually invalidate the previous caches
```
evgenyrp, line 28: Are caching changes part of this PR? I'd say it should be separate.

taskcluster/translations_taskgraph/transforms/cached_tasks.py
```diff
@@ -82,6 +84,9 @@ def add_cache(config, jobs):
                             digest_data.append(f"{param}:{value}")
                             break
 
+        if cache_version:
```
evgenyrp, line 87: Ok, the changes here are quite small. I see you're keeping the files and just adding another way to invalidate caches. Up you you if you want to separate this.

taskcluster/translations_taskgraph/transforms/continuation.py
```diff
@@ -0,0 +1,339 @@
...
+    """
+    When an existing corpus is available, rewriting the task graph to omit the steps
+    needed to generate that corpus.
+
+    Rewrites dependencies:
+        corpus-merge-parallel -> corpus-parallel
+        corpus-merge-mono-trg -> corpus-backtranslations
+        distillation-parallel-keep-best -> corpus-distillation
+    """
+    # import sys
```
evgenyrp, line 139: It's worth adding a comment that this is needed for debugging

taskcluster/translations_taskgraph/transforms/continuation.py
```diff
@@ -0,0 +1,339 @@
...
+                    continue
+                rewrite_dependencies(
+                    job,
+                    old_task="corpus-align-backtranslations",
+                    new_task="continuation-corpus-backtranslations",
+                )
+
+        if corpus_distillation:
+            if stage in {
+                # "distillation-parallel-src-extract-best",
```
evgenyrp: commented code

taskcluster/translations_taskgraph/transforms/continuation.py
```diff
@@ -0,0 +1,339 @@
...
+                "train-teacher-model",
```
evgenyrp, line 248: What if we want to add another step to the pipeline? It will be tricky to find these lists. Would moving them to the top of the file help?

eu9ene (approved):
I'm happy we have a way to retrain old models in any part of the pipeline now! I'm going to use it for CJK right away!

On the downside, we see that this functionality significantly raises the complexity of task graph handling, especially when we consider it together with other transforms, legacy training continuation and the train action. I don't have a good suggestion here except to brainstorm later on how we can unify all this and have fewer ways to restart things. Modifying the graph will definitely be needed for other things too (for example to remove back translations completely), so we should just make sure this part is easy to understand. 

All my comments are minor optional suggestions, so I'm not going to block this PR, but it would be nice to address at least the easy ones if they make sense to you.

docs/training/using-pretrained-models.md
gregtatum: I think you are correct, I'm adding this as a follow-up since we want to land this for your training.

https://github.com/mozilla/translations/issues/1150

pipeline/continuation/model.py
gregtatum: I clarified how this works in the comment and filed a follow-up. #1151

```
Continue the training pipeline with an existing model. This produces the `continue-model`
task, which is dynamically used (at taskgraph generation time) as a dependency rather
than a `train`

TODO(#1151) - This only support backwards models with the "use" mode. We may be able
to unify everything here.
```

pipeline/continuation/model.py
gregtatum: Ok, so I looked at this a bit, and I think the correct thing to do here is to move the `final.model.npz.best-chrf.*` to the top of the list, as only one file is output, but the others are searched. This should make it robust for old archived models. It should be harmless to include old one, and then it can help not break for re-using old models.

pipeline/continuation/model.py
gregtatum: https://github.com/mozilla/translations/issues/1152

taskcluster/translations_taskgraph/transforms/cached_tasks.py
gregtatum: Yeah I initially did it just on corpus continuation to try it out. I'd prefer to keep it just to not break things in the landing, then we can follow up on the broader changes for versioning and decide there.

taskcluster/translations_taskgraph/transforms/continuation.py
gregtatum: Yeah this was just a mistake to include, but I think it's worth keeping with a comment as you suggest.

taskcluster/translations_taskgraph/transforms/continuation.py
gregtatum: This was more of a mitigation to ensure I got it all correct, so it should be fine to add new stages. I'm updating the comment here:

```
                # Skip any jobs that should never be produced. This helps ensure
                # that if they do somehow get produced, the taskgraph will fail to
                # fully resolve.
```

taskcluster/translations_taskgraph/transforms/continuation.py
gregtatum: Thanks, I can remove that.

PR #1069, opened 2025-04-01
fix: remove unnecessary tool installs in kenlm build script

bhearsum:
cmake 4.0 was released recently and its installation in this script is causing its toolchain build to fail.

I traced this back a bit and figure out exactly why we're doing it - cmake and the other tools were already installed on this docker image when this script was added. It was probably leftover from some early iteration of my work.

### DIFF
--- a/taskcluster/scripts/toolchain/build-kenlm.sh
+++ b/taskcluster/scripts/toolchain/build-kenlm.sh
@@ -4,16 +4,7 @@ set -x
 
 KENLM_DIR=$MOZ_FETCHES_DIR/kenlm-source
 
-# TODO: I don't think we actually need the C++ stuff? just the python module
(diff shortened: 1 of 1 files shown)


taskcluster/scripts/toolchain/build-kenlm.sh
```diff
@@ -4,16 +4,7 @@ set -x
 
 KENLM_DIR=$MOZ_FETCHES_DIR/kenlm-source
 
-# TODO: I don't think we actually need the C++ stuff? just the python module
```
gregtatum, line 7: > I don't think we actually need the C++ stuff?

Narrator voice: We don't.

gregtatum (approved):
Looks good to me!

PR #1073, opened 2025-04-07
Add support for fast_align

eu9ene:
Fast refactoring that adds support for fast_align to incorporate the code I used for some old experiments.

It can be handy for shortlisting in case the quality of fast_align alignments is higher. Shotlist uses spm tokens which reduce vocabulary significantly compared to ICU tokenized alignments that we use for augmentations. The main issue with fast_align is high memory utilization when the vocabulary size is high.

### DIFF
--- a/pipeline/alignments/align.py
+++ b/pipeline/alignments/align.py
@@ -45,6 +45,11 @@ class Tokenization(Enum):
     icu = "icu"
 
 
+class AlignerType(Enum):
+    eflomal = "eflomal"
+    fast_align = "fast_align"
+
+
 def run(
     corpus_src: str,
     corpus_trg: str,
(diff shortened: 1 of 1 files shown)


gregtatum (changes requested):
This is a soft-request changes with a question. It seems like we would want to hook this up the training config so that we can actually run the code. Right now it's kind of dead code since we can't really run it, and it doesn't have tests to check it.

eu9ene:
> This is a soft-request changes with a question. It seems like we would want to hook this up the training config so that we can actually run the code. Right now it's kind of dead code since we can't really run it, and it doesn't have tests to check it.

I don't think we want to make it configurable at this point, as it's not proven it's in any way better than eflomal for shortlisting.  Likely it's not,  but it will definitely OOM for ICU tokenized corpus. I just wanted to preserve this code because we historically used fast_align, then switched to eflomal, then I experimented with fast_align again when I had some issues with eflomal. If we were to switch it, it would be in the Taskcluster kind and not in the config, as it's not something we want to configure.

eu9ene:
We decided not to commit the code that we don't need at the moment, so I'm closing the PR and leaving this branch in case we need it in the future.

PR #1075, opened 2025-04-10
Rename tasks

gregtatum:
Putting this here for a CI run.


taskcluster/config.yml
```diff
@@ -18,39 +18,40 @@ taskgraph:
...
-    - merge-translated
-    - score
-    - cefilter
+    - translate-distillation-parallel-src
+    - distillation-parallel-extract-best
+    - collect-distillation-parallel
+    - translate-distillation-mono-src
+    - collect-backtranslations
+    - merge-distillation
+    - distillation-corpus-score
```
eu9ene, line 45: Should we get rid of the word "corpus" everywhere? We basically have a monolingual and parallel corpus, and I see you renamed some of it to follow these terms.

taskcluster/config.yml
```diff
@@ -18,39 +18,40 @@ taskgraph:
...
+    - distillation-corpus-keep-best
     - alignments-teacher
```
eu9ene, line 47: There should be separate stages for "alignments-original" and "alignments-backtranslated". I think I ran into an  issue of having one stage for two steps a couple of times.

eu9ene: or maybe it should be called "alignments-parallel" and "alignments-backtranslations" following your new naming scheme

eu9ene (commented):
I have some naming suggestions. Let's discuss.

taskcluster/config.yml
gregtatum: There is some ambiguity here I was trying to resolve by using `corpus`. We have a distillation `parallel-src` branch, and a distillation `mono-src`

```
translate-distillation-mono-src
translate-distillation-parallel-src
```

However, after this step, we still have to `extract-best` on the `parallel-src`, but not on the `mono-src`.

```
distillation-parallel-extract-best
```

Maybe this could be renamed to:

```
distillation-parallel-src-extract-best
```

Then I could rename the `corpus` parts to `parallel`.

```
distillation-parallel-keep-best
distillation-parallel-score
distillation-parallel-shortlist
```

These two steps are doing very similar but different things, and at different parts of the pipeline, and it's hard to disambiguate them.

```
distillation-parallel-keep-best
distillation-parallel-src-extract-best
```

bhearsum (approved):
I don't have an opinion on the names, but I sanity checked the taskgraph bits, and they look fine to me.

taskcluster/config.yml
```diff
@@ -18,46 +18,48 @@ taskgraph:
 # The list of valid stages that can be used with `target-stage and `start-stage`.
 # These get attached to tasks in `kinds`.
 valid-stages:
-    - clean-corpus
-    - clean-mono
+    - alignments-backtranslations
+    - alignments-distillation
+    - alignments-teacher
```
eu9ene, line 23: I see you've split the alignments stage, but "alignments-teacher" looks inconsistent with other naming. "alignments-backtranslations" is also used to train the teacher. Maybe "alignments-parallel" would be more consistent? Of it, it's still ambiguous, something like "alignments-teacher-backtranslations" and "alignments-teacher-parallel".

taskcluster/config.yml
```diff
@@ -18,46 +18,48 @@ taskgraph:
...
-    - train-backwards
+    - clean-mono
+    - clean-parallel
+    - collect-backtranslations
+    - collect-distillation-mono
+    - collect-distillation-parallel
+    - distillation-parallel-keep-best
+    - distillation-parallel-score
+    - distillation-parallel-shortlist
+    - distillation-parallel-src-extract-best
```
eu9ene, line 38: Why is it called "distillation-parallel-src-extract-best"? It compares the translation to the reference, so it uses both source and target parts of the parallel corpus.

taskcluster/config.yml
```diff
@@ -18,46 +18,48 @@ taskgraph:
...
-    - all-pipeline
-    - all-pr-pipeline
+    - export
+    - finetune-student
+    - merge-cleaned-mono
+    - merge-cleaned-parallel
+    - merge-devset
+    - merge-distillation
+    - quantize-student
+    - split-backtranslations
```
eu9ene, line 53: These confuse me. Should it be like this?
```
split-backtranslations-mono-trg
split-distillation-mono-src
split-distillation-parallel-src
```

eu9ene: Also it would be consistent with the "translate*" steps

taskcluster/config.yml
```diff
@@ -18,46 +18,48 @@ taskgraph:
...
 # These get attached to tasks in `kinds`.
 valid-stages:
-    - clean-corpus
-    - clean-mono
+    - alignments-backtranslations
+    - alignments-distillation
+    - alignments-teacher
+    - all-pipeline
+    - all-pr-pipeline
+    - analyze-corpus
```
eu9ene, line 26: analyze parallel?

taskcluster/config.yml
```diff
@@ -18,46 +18,48 @@ taskgraph:
...
-    - merge-devset
-    - merge-mono
-    - train-vocab
-    - train-backwards
+    - clean-mono
+    - clean-parallel
+    - collect-backtranslations
+    - collect-distillation-mono
+    - collect-distillation-parallel
+    - distillation-parallel-keep-best
```
eu9ene, line 35: This section definitely confuses me a lot. I'm thinking, what if we replace "parallel" with "all" for stages where we have merged the mono+parallel corpus? It wouldn't be confused with the original parallel corpus then. For example,  "distillation-all-keep-best." Would it be easier to understand?

eu9ene (changes requested):
Ok, naming is hard :) I think it's getting better and better, but I went though the list again and realized that I dont't understand what some steps are and what's the order of them (and I'm supposed to be the best person to know this...). Since it will be hard to rename it in the future, let's do another iteration of discussing. I outlined some specific suggestions.

Another optional idea: what if instead of grouping the steps by type like "translate*" we group them by stage: e.g. "distillation-translate-mono-src", "backtranslations-translate-mono-trg". Would it be easier to understand what belongs where? I'm not sure.

eu9ene:
> This is the final after reviewing it with @eu9ene .
> 
> ```
> all-pipeline
> all-pr-pipeline
> export
> fetch-{item}
> dataset-{name}
> toolchain
> docker-image-{name}
> 
> backtranslations-mono-trg-chunk
> backtranslations-mono-trg-dechunk-translations
> backtranslations-mono-trg-translate
> backtranslations-train-backwards-model
> 
> build-vocab
> 
> continuation-corpus-{corpus}        # See WIP PR #1066
> continuation-model-{model}          # See WIP PR #1066
> continuation-vocab                  # See WIP PR #1066
> 
> corpus-align-backtranslations
> corpus-align-distillation
> corpus-align-parallel
> corpus-analyze-mono-{dataset}
> corpus-analyze-parallel-{dataset}
> corpus-clean-mono-{dataset}
> corpus-clean-parallel-{dataset}
> corpus-clean-parallel-bicleaner-ai-{dataset}
> corpus-clean-parallel-fetch-bicleaner-model
> corpus-merge-devset
> corpus-merge-distillation
> corpus-merge-mono
> corpus-merge-parallel 
> 
> distillation-corpus-build-shortlist
> distillation-corpus-final-filtering      # cefilter is reference-less, and can be run across all of the distillation corpus
> distillation-mono-src-chunk
> distillation-mono-src-dechunk-translations
> distillation-mono-src-translate
> distillation-parallel-src-chunk
> distillation-parallel-src-dechunk-translations
> distillation-parallel-src-translate
> distillation-parallel-src-translations-keep-best
> distillation-parallel-src-translations-score
> distillation-student-model-finetune
> distillation-student-model-quantize
> distillation-student-model-train
> 
> evaluate-backwards-model
> evaluate-quantized-student-model
> evaluate-student-model
> evaluate-teacher-ensemble
> evaluate-teacher-model
> 
> train-teacher-model
> ```

Ok, I'm approving based on the naming we discussed. Thank you for doing this refactoring!

eu9ene (approved):

gregtatum:
Trying for a full CI run with the new renames.

PR #1079, opened 2025-04-14
Add devcontainer support

gregtatum:
This migrates our local docker support to be compatible with vscode's [Dev Containers](https://code.visualstudio.com/docs/devcontainers/containers). I've been doing all of my work by first running docker, then attaching VS Code to the container. This makes it easy to do all of the python installs and cmake builds so that I know they will work. There is an official way to do this workflow through VS Code with only slight restructuring.

First, I migrated our docker scripts to use docker compose, which is frankly a nicer and more declarative API to do this kind of work. Then I moved all of the docker files into `.devcontainer` to be compatible with this feature, and added the `devcontainer.json` file. I added a few default plugins as well that will enable support for our tooling that gets checked in CI, like ruff, black, cmake, etc. We can always follow-up with adjusting this list as needed.

The existing docker commands work just the same as before, and I added a `docker-attach` to attach a terminal to the vscode invoked docker image. So the dev containers are in no way required, just a convenience for vscode users.


.devcontainer/docker-compose.yml
```diff
@@ -0,0 +1,36 @@
+name: translations
+
+services:
+  translations-base:
```
eu9ene, line 4: This will run 3 separate Docker containers. I don't think it was the goal, right? It was mostly to reuse the parent images. I'm not sure how to properly do this with docker-compose. It's usually better than Docker commands when you have several things to run at the same time.

gregtatum: Yeah as far as I can tell we would have to remove the dynamic `FROM $DOCKER_IMAGE_PARENT` from all of the `Dockerfile`s to support using docker compose. I'm not sure how that affects our production use cases of those images.

Using docker compose does have the side effect of launching those services, but they immediately exit, as they don't have any command passed to them, so they will have a blip of CPU to load them, and then the memory and disk space is never really actualized, and remains virtual according to docker. So really it's almost a null operation. It takes about 2-3 seconds to launch the dev container when I tested it, which includes initiating the editor. It takes about 1 second to run the `task docker` from the terminal

```
CONTAINER ID   IMAGE                COMMAND                  CREATED         STATUS                     PORTS      NAMES                               SIZE
9113ea84d106   translations-local   "/bin/sh -c 'echo Co…"   3 minutes ago   Up 3 minutes                          translations-translations-local-1   299MB (virtual 7.49GB)
fe3d886ee9cd   translations-test    "/bin/bash"              22 hours ago    Exited (0) 3 minutes ago              translations-translations-test-1    0B (virtual 821MB)
27f01cecabcf   translations-base    "/bin/bash"              22 hours ago    Exited (0) 3 minutes ago              translations-translations-base-1    0B (virtual 709MB)
```

Here's a screenshot where a dev container is is running:

image

gregtatum: I looked into this more, and it won't work with our current setup. It doesn't allow dynamic scripting.

PR #1086, opened 2025-04-24
Merge browsermt/marian-dev into inference/marian-fork

gregtatum:
edit: This is now ready for review. Note that the only place the commit history isn't clean is later on I decided to rename marian to marian-fork, and so that is a later commit. It was going to be too hard to rebase through that change, so I decided not to bother. The first commit is the big one with the check-in, but the rest should hopefully be reasonable to review.

---

Putting this here to get some early feedback from @nordzilla.

So to explain what's going on, the first commit is the bulk of the changes, as it's a naive commit of the `browsermt/marian-dev` fork of code into `inference/marian`. The rest of the changes are integrating it into the codebase. The idea in a review will be to look at it on a commit-by-commit basis, as it's not really feasible to review the first commit.

I then follow up with lots of integration commits of making it build with Cmake using our existing build task.

I also have commits where I attempt to trim down unneeded files in the marian repo, which are effectively dead code.

Probably as a follow-up I'd like to get this integrated into our CI so that we can run the existing tests, and create our own new tests.


nordzilla:
@gregtatum and I walked through these changes on a video call together, and I agree with the approach of moving this out of third party, trimming it down to our use cases, and fixing/improving specific areas as needed.

inference/marian/scripts/alphas/extract_stats.py
```diff
@@ -1,63 +0,0 @@
-#!/usr/bin/env python3
```
gregtatum: This one is technically used in the pipeline, but I'll restore it in a later PR where we use this code in the training pipeline. The training pipeline is still building from browsermt/marian-dev.

inference/patches/02-marian-allocation.patch
```diff
@@ -1,23 +0,0 @@
...
-
-diff --git a/inference/3rd_party/browsermt-marian-dev/src/tensors/tensor_allocator.h b/inference/3rd_party/browsermt-marian-dev/src/tensors/tensor_allocator.h
-index e3bc79f9..66f8e44d 100644
---- a/inference/3rd_party/browsermt-marian-dev/src/tensors/tensor_allocator.h
-+++ b/inference/3rd_party/browsermt-marian-dev/src/tensors/tensor_allocator.h
-@@ -13,7 +13,7 @@ class TensorAllocator {
- private:
-   const size_t CHUNK = 128;
-   const size_t MBYTE = 1024 * 1024;
--  const size_t GROW = CHUNK * MBYTE;
```
gregtatum, line 19: I'm going to audit this in my memory patch, so I'm just going to remove for now.

inference/scripts/build.py
```diff
@@ -0,0 +1,110 @@
...
+    print(f"[build] Removing uncompressed archive: {tar_path}")
+    tar_path.unlink()
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        description="Run cmake and make from the inference directory."
+    )
+    parser.add_argument("--test", action="store_true", help="Compile the test code as well")
+    parser.add_argument(
```
gregtatum, line 86: `--archive` and `--cuda_toolkit` are technically dead code paths, as they are required for the full use of `marian-fork` in the pipeline, but it seemed harmless to keep them for now.

nordzilla (approved):
I went commit by commit and everything here looks reasonable. 

I also tested `inference-test` and `inference-test-wasm` locally with clean Docker builds, and everything worked.

Next, I tested everything end-to-end and everything worked with the new WASM binary in Firefox, however, there do seem to be slight differences.

```
drwxrwxr-x 2         - nordzilla 29 May 11:28 wasm-compare
.rw-r--r-- 1 4,953,481 nordzilla 29 May 11:28 ├── new.wasm
.rw-r--r-- 1 4,956,176 nordzilla 29 May 11:27 └── old.wasm
```

The new WASM file seems to be about 2.5KB smaller than the old file, and this also changes the generated JS slightly:

```diff
diff --git a/toolkit/components/translations/bergamot-translator/bergamot-translator.js b/toolkit/components/translations/bergamot-translator/bergamot-translator.js
index e0339b61195d..5acab5546888 100644
--- a/toolkit/components/translations/bergamot-translator/bergamot-translator.js
+++ b/toolkit/components/translations/bergamot-translator/bergamot-translator.js
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */

 function loadBergamot(Module) {
-  var BERGAMOT_VERSION_FULL = "v0.5.0+b475d25";
+  var BERGAMOT_VERSION_FULL = "v0.5.0+b004fcba";
   null;

   var Module = typeof Module != "undefined" ? Module : {};
@@ -693,7 +693,7 @@ function loadBergamot(Module) {
   var tempI64;

   var ASM_CONSTS = {
-    1429732($0, $1, $2, $3, $4) {
+    1427124($0, $1, $2, $3, $4) {
       if (!Module.getOrCreateSentenceSegmenter) {
         Module.getOrCreateSentenceSegmenter = (function () {
           let segmenters = new Map();
```

---

> Question

Should we update everyone in Firefox to the latest WASM binary and generated JS?

On one hand, it's nice to have reassurance the latest changes are compatible and working end-to-end. On the other hand we'd be forcing everyone to re-download a new WASM binary for effectively no substantial change.

PR #1102, opened 2025-05-01
Fix prod config after renaming

eu9ene:


gregtatum (approved):
Thanks for the quick fix.

gregtatum:
Resolves #1103.

PR #1110, opened 2025-05-03
Refactor and fix importers

### LINKED ISSUE #1040: dataset-opus-NLLB_v1-de-en fails
https://firefox-ci-tc.services.mozilla.com/tasks/CktOe5IzSlimNstUynBbyw/runs/2

It's a big dataset (250M), so it seems it ran out of time (24h) to download.

eu9ene:
Make importers work with de-en datasets:
- get rid of bash
- fix dataset regex
- bump disk
- throw a specific exception from the downloaders
- delete the output file when a download fails
- add a test for the opposite language pair for opus

closes #1040


bhearsum (approved):

pipeline/common/downloads.py
```diff
@@ -18,19 +18,28 @@
 logger = get_logger(__file__)
```
gregtatum, line 18: If you add this path to `pyproject.toml` the `[tool.pyright]` `include` section you'll get some type checking on it that gets checked in CI. I've been slowly doing this for files that I touch. You usually need to fix a few things, but it can catch bugs and help ensure things are properly typed.

pipeline/common/downloads.py
```diff
@@ -175,7 +184,7 @@ class DownloadChunkStreamer(io.IOBase):
              gzip.GzipFile(fileobj=f)
     """
 
-    def __init__(self, url: str, total_retries=3, timeout_sec=10.0, wait_before_retry_sec=60.0):
+    def __init__(self, url: str, total_retries=3, timeout_sec=10.0, wait_before_retry_sec=1.0):
```
gregtatum, line 187: Is this an intentional to change? I know you had wanted to limit being rate limited by some of these download services.

pipeline/common/downloads.py
```diff
@@ -18,19 +18,28 @@
 logger = get_logger(__file__)
 
 
+class DownloadException(Exception):
```
gregtatum, line 21: Ah, this is a nice change here.

pipeline/data/importers.py
```diff
@@ -0,0 +1,221 @@
+import shutil
```
gregtatum, line 1: Should this be called `download_parallel.py` instead so that it matches the `download_mono.py` and is less ambiguous?

gregtatum, line 1: I generally like to have a short file-level comment explaining why this file exists, e.g.

pipeline/data/dataset_importer.py
```diff
@@ -15,8 +15,8 @@
 import os
```
gregtatum, line 15: I'm finding the naming of files a bit confusing. Do you have any ideas on how to be a bit less ambiguous?

We have `downloader` and `importer` which mean pretty much the same thing. In the tasks we have now `parallel` and `mono` when talking about datasets. Perhaps it would be nice to rename the files to match our new naming scheme and decide on `downloader` vs `importer` on terms. There's also some ambiguity around the fact that the parallel files are responsible for different things. `importers.py` downloads the parallel datasets, and the `dataset_importer.py` augments them.

gregtatum (approved):
I left a few minor suggestions, but nothing is big enough to mark it as request changes. Thanks for doing this, it will be nice having our implementation more uniform here.

pipeline/data/importers.py
```diff
@@ -0,0 +1,221 @@
...
+    url = "url"
+
+
+def opus(src: str, trg: str, dataset: str, output_prefix: Path):
+    """
+    Download a dataset from OPUS
+
+    https://opus.nlpl.eu/
+    """
+    logger.info("###### Downloading opus corpus")
```
gregtatum, line 30: I'd prefer not to have these `####` marks in Python files. I'm assuming they were useful in bash files, but here I would prefer to rely on the loggers and their `[name] log message` format to disambiguate logs. Also this log is repeated twice.

pipeline/common/downloads.py
eu9ene: No, I was testing something locally, good catch

pipeline/data/dataset_importer.py
eu9ene: Agreed. I think we can call them `parallel_importer`, `mono_importer` and `paralle_downloaders` where `importer` is something that not only downloads data but does something else required to use the data, like shuffling or augmentation. The `downloader` just downloads the required amount of data from a resource.

pipeline/common/downloads.py
eu9ene: Yeah, it's a good practice to throw specific exceptions so that the `except` code doesn't catch everything. Python has very few built-in ones. Usually, when I don't want to create a custom one, I throw `ValueError`, which is a generic user code exception. In this case, having a custom type makes sense.

PR #1115, opened 2025-05-06
Build Marian with debug info

gregtatum:
This builds a release cut of Marian, but includes the debug information. This increases binary size, and maybe some memory usage as well, but otherwise the speed of the binary should be just as fast. I have this running on my `h1-2025` branch, and it's been useful to see the full stack traces of crashes within Marian.


gregtatum:
I flagged @nordzilla in case I missed anything, since it has to do with a Marian build.

nordzilla (approved):
This looks good to me. 

This is only the build for the pipeline, correct? 

Our regular inference build is unaffected.

gregtatum:
Yeah this doesn't affect the Wasm builds or anything.

PR #1120, opened 2025-05-15
backout switching gpu workers to docker + new worker image

bhearsum:
This is a backout of #700, and a necessary taskgraph upgrade to do so. Hopefully it won't be needed, but if we continue to hit https://github.com/mozilla/translations/issues/1117 even after picking up https://github.com/taskcluster/taskcluster/pull/7723, it will hopefully be a viable way to unblock training while we work through those issues.

I won't be shocked if the initial version of this patch fails some tasks - it was a tricky backout to do after the renaming.


pyproject.toml
```diff
@@ -44,7 +44,7 @@ PyGithub="2.4.0"
 pyperclip="1.9.0"
 ruamel-yaml = "^0.18.6"
 taskcluster = "^56.0.3"
-taskcluster-taskgraph = "^14.1.1"
+taskcluster-taskgraph = "^14.2.1"
```
eu9ene, line 47: I noticed we have version `v14.2.1` pinned in `.taskcluster.yml` but here we use `^`, so it can be automatically bumped on the next `poetry lock`. Is it expected?

eu9ene (approved):

pyproject.toml
bhearsum: That's a good callout; ideally we'd keep those in sync. The one in `.taskcluster.yml` is used when taskgraph is run during decision tasks. As I understand it, the one here is used for local tests and the `taskgraph` tests that run on pushes/prs.

I'll send a follow-up once this lands to pin to the specific version in `pyproject.toml`, and add comments everywhere about keeping it in sync.

bhearsum: https://github.com/mozilla/translations/pull/1124 for this.

PR #1125, opened 2025-05-20
Fix NTREX was not being ignored in config generator

ZJaume:
Created a config for Dutch and NTREX was in the training data. This should fix it.

I think this fix must be included in any H1-2025 training that starts from a new corpus. Just adding both of you to review so you get notified.


gregtatum:
Ah, that explains why I had to comment all of those out in my configs :D

PR #1126, opened 2025-05-20
Re-use the local development container when possible

gregtatum:
We were always creating and destroying containers when attaching to them, which can be negative when working locally, as you can't have any persistent state. The old behavior can be retained by passing in `task docker -- --clean`. This allows for a more stable local development process when running docker.

### DIFF
--- a/utils/tasks/docker-run.py
+++ b/utils/tasks/docker-run.py
@@ -51,7 +51,7 @@ def container_exists(container_name: str):
         "--all",
         "--filter", f"name={container_name}",
         "--format", "{{.Names}}"
-    ] # fmt: skip
+    ]  # fmt: skip
     result = subprocess.run(
         command,
         stdout=subprocess.PIPE,
--- a/utils/tasks/docker-run.py
+++ b/utils/tasks/docker-run.py
@@ -39,10 +39,7 @@ def get_args():
         help="Provide a custom container name",
     )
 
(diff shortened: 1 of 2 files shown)


utils/tasks/docker-run.py
```diff
@@ -23,50 +26,94 @@ def get_args():
...
+                docker_command.extend(["--volume", volume])
+
+        # Run Docker with the current user's UID and GID if --run-as-user is specified
+        if args.run_as_user:
+            uid = os.getuid()
+            gid = os.getgid()
+            docker_command.extend(["--user", f"{uid}:{gid}"])
+
+        # Specify the Docker image
+        docker_command.append("translations-local")
```
nordzilla, line 108: > Question

Should this use the new `IMAGE_NAME` defined above?

nordzilla (approved):
The changes and justifications for them look reasonable to me.

nordzilla (changes requested):
Actually, I'm getting an issue. 

On your new branch, when I run:

```
task docker-run -- --run-as-user task inference-build-wasm
```

Then I get the following output:

```
task: [docker-run] utils/tasks/docker-run.py --run-as-user task inference-build-wasm
[docker-run] Attaching to the existing container translations-local-dev
```

And I'm within the docker terminal. It doesn't run my command anymore.

PR #1130, opened 2025-05-23
Sample sentences from corpus before building vocab

ZJaume:
This way only the sample for the vocab is stored decompressed in disk.

### DIFF
--- a/pipeline/train/spm-vocab.sh
+++ b/pipeline/train/spm-vocab.sh
@@ -66,10 +66,18 @@ fi
 vocab_dir=$(dirname "${vocab_src_output}")
 mkdir -p "${vocab_dir}"
 
-zstdmt -dc "${merged_corpus_src}" >"${vocab_dir}/data.src.txt"
-zstdmt -dc "${merged_corpus_trg}" >"${vocab_dir}/data.trg.txt"
-
 if [ "$vocab_split" == "true" ] || [ "$vocab_split" == "True" ]; then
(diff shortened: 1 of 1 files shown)


pipeline/train/spm-vocab.sh
```diff
@@ -66,10 +66,18 @@ fi
...
-zstdmt -dc "${merged_corpus_trg}" >"${vocab_dir}/data.trg.txt"
-
 if [ "$vocab_split" == "true" ] || [ "$vocab_split" == "True" ]; then
+  # Sample the input corpus to at most 10M sentences to avoid running out of disk space
+  # that's the amount SP uses, so it doesn't make sense to save in disk more than that
+  # after that, remove the tab separator
+  paste \
+    <(zstdmt -dc "${merged_corpus_src}") \
+    <(zstdmt -dc "${merged_corpus_trg}") \
+    | shuf -n "${sample_size}" \
```
eu9ene, line 76: Does shuf use memory or disk for shuffling?

eu9ene (approved):
Thanks for optimizing instead of just bumping disk like me :) It still might use some disk/memory depending on how shuf works.

pipeline/train/spm-vocab.sh
ZJaume: Yes, but I've checked and it seems that the `-n` option is only loading in memory the same as output size, therefore I guess we are fine. 5 or 10M sentence pairs should only be a few GB.

gregtatum (approved):

Issue #1025, opened 2025-02-04
Rerun only tasks downstream from start-stage


evgenyrp:
Currently, if we use `start_stage` and `previous_group_ids`it reuses only the tasks that are ancestors of the target-stage tasks. It causes rerunning tasks that are in a different branch of the graph.

For example, when specifying:
```
target-stage: all-pipeline
start-stage: shortlist
previous_group_ids: ["anxoOR6BQRGiRteielF1gQ"]
```

Assuming there was a change in the alignments code, it would also rerun all the tasks starting from `alignments-student-en-ko` and all the evals tasks: https://firefox-ci-tc.services.mozilla.com/tasks/groups/MDBU9q0LTAaxsKWmLLzvFw

The intention is usually to rerun only the tasks that are downstream of the start-stage task.

Related #728

eu9ene:
I couldn't figure out how to fix it as it's not clear how to debug the train action. As a workaround this at least doesn't retrigger the student training. It does rerun tons of `evaluate` tasks which wastes GPU resources:

```
existing_tasks: {
    "finetune-student-en-ko": "TAxLaqg0Q6ed-dhzu766Hw",
    "alignments-student-en-ko": "DY1ULIDMTgiuJcmdFjX3Xg",
    "train-student-en-ko": "Zdx_An18Rz2JFvMg_ICHdQ",
}
```

Issue #728, opened 2024-07-09
start_stage often reruns amost all "evaluate" tasks


evgenyrp:
I ran this one to do the export task but since "evaluate" tasks are not sequential it leads to rerunning them each time I use start_stage which wastes GPU resources. 

https://firefox-ci-tc.services.mozilla.com/tasks/aAroslw9Ru-c5cam6SvLBg

bhearsum:
@eu9ene - Any thoughts on the two options above?

eu9ene:
Another option is to run them right after training and make the next task depend on them. It makes sense because this way we don't continue the pipeline until we have good evaluation results. There was even a suggestion to add a sanity check (see https://github.com/mozilla/firefox-translations-training/issues/78). They will not be the dependencies of the "all" task then.

bhearsum:
Sure, if you want to always run them when you run a training, adjusting the `stage` entries such as https://github.com/mozilla/firefox-translations-training/blob/f7247a60a095015d39fb73065830eb9e980147e2/taskcluster/kinds/evaluate/kind.yml#L118 to be the same as their associated training step would be a good idea. Doing that would mean they always end up in the same task groups, which would let you remove them from the `all` and `all-pr` dependencies with no real downside, and should fix this.

evgenyrp:
Fixed in #1132

Issue #719, opened 2024-07-02
Improve usability of running selected tasks


evgenyrp:
While working on the big training and bug fixes I ran into many issues with scheduling specific tasks. Basically, the graph and caches can be in an arbitrary state and we still should be able to run the pipeline starting with specific stages and reusing the stages that ran before.

We currently have several tools to work with:
1. `target-stage`
2. `start_stage` and `previous_group_ids`
3. `existing_tasks`
4. pre-trained models
5. Git branches
6. Adding extra tasks to the graph

Usually, I see there's an issue when it starts scheduling tasks I don't need to schedule and I try to come up with a workaround using those tools. Also, using the current tools adds a significant mental load and is hard to use when training 10s of languages with fixes at the same time (see [this PR](https://github.com/mozilla/firefox-translations-training/pull/678)). We should rethink this approach to make it more flexible and easy to use.

Maybe introducing a concept of state similar to the data on disk in Snakemake can help here. There was an option to skip smart scheduling based on file creation dates and information about the past runs and just treat everything present on disk as completed tasks and schedule the rest.

bhearsum:
I agree that we need to step back and take a more holistic look at how to deal with this.

We don't really have a great analogue for "things on disk" in Taskcluster. The closest is the cached tasks, but obviously we've run into a number of issues with that in the past, and most of the tools listed above have been workarounds for issues we've had with it. To be quite honest, if the level of flexibility that Snakemake provided (ie: being able to manipulate it by adjusting things on disk) is needed, it may be worth considering reviving it, and perhaps adding the ability to scale into the cloud with Slurm. I seriously doubt that Taskcluster will be able to provide the same level of flexibility and simplicity at the same time.

With that all said, I think we can probably make things a lot better than they are. If the goal is "be able to override anything, anywhere", one idea is to require training configs to have all previous artifacts specified. Perhaps this would take the form of pointers to either tasks, or task artifacts that have been uploaded to GCS. Anything not present would then be scheduled.

Whatever the goals and/or use cases are, having a complete list of them written down would be very helpful. We can't design a solution until we know the extent of the problem!

bhearsum:
> With that all said, I think we can probably make things a lot better than they are. If the goal is "be able to override anything, anywhere", one idea is to require training configs to have all previous artifacts specified. Perhaps this would take the form of pointers to either tasks, or task artifacts that have been uploaded to GCS. Anything not present would then be scheduled.

[...]

PR #1132, opened 2025-05-23
Resume pipeline

### LINKED ISSUE #1129: Implement Metaflow style "resume"
It's easy to continue the workflow after failure in Metaflow:
```
python flow.py --environment pypi resume --origin-run-id 123 <step_name>
```

We can easily implement a similar mechanism in the Train action. We already have `previous_group_ids` and `start_stage`. We keep only the tasks that are ancestors of the start_stage ones, but our pipeline has multiple branches. We should find all the completed tasks and their ancestors from `previous_group_ids` and add them to `existing_tasks`.

Here's an example code that does that:

```python
 import taskcluster

  TC_MOZILLA = "https://firefox-ci-tc.services.mozilla.com"
  options = {"rootUrl": ("%s" % TC_MOZILLA)}
  queue = taskcluster.Queue(options=options)
  tasks_to_add = {}
  for group_id in previous_group_ids:
      group = queue.listTaskGroup(group_id)
      for task in group["tasks"]:
          if task["status"]["state"] != "completed":
              continue

          task_id = task["status"]["taskId"]
          task_name = task["task"]["metadata"]["name"]
          # Skip service tasks
          if (
              task_name.startswith("Action")
              or task_name.startswith("Decision")
              or task_name.startswith("PR")
          ):
              continue

          # Tasks from later previous_group_ids task groups will override previously found tasks
          tasks_to_add[task_name] = task_id

  if tasks_to_add:
      # Add ancestors of all the top level completed tasks
      for task_id, label in get_ancestors(list(tasks_to_add.values())).items():
          if (
              task_name.startswith("Action")
              or task_name.startswith("Decision")
              or task_name.startswith("PR")
          ):
              continue
          tasks_to_add[label] = task_id
``` 

It's also possible to do `python flow.py --environment pypi resume` in Metaflow without specifying the last run ID and the start step. I'm not sure if we can do this as we don't have a database with all the runs. It would be possible If we could find all the task groups with a specific experiment name using Taskcluster API.

evgenyrp:
Add Metaflow style resume mode: reuse all the completed tasks from the specified `previous-groups-ids`.

If `start-task-prefix` is specified, the tasks labelled with such a prefix and all their descendants will rerun.

It logs quite a lot more now, but since I don't know how to debug train actions locally, it's better to have more logs to easier identify problems in the future.

It's also possible now to restart only one task from a stage, for example, bicleaner-ai-nllb.

**Example train action 1:** 
https://firefox-ci-tc.services.mozilla.com/tasks/QYhpcdFUSXK5W3cQSgTsmA/runs/0/logs/public/logs/live.log 

Parameters:
```yaml
start-task-prefix: backtranslations-train-backwards-model
previous-group-ids: ["QDc7Pm4cSCy70tXhBUW-1w", "B04UrSOnT_iv0WqSw-yShg", "H3zT8CiTRKafIh4C9lba-w"]
```

It triggers exactly one task as pending as expected: trains the backwards model.

**Example train action 2:**
https://firefox-ci-tc.services.mozilla.com/tasks/JHVsJhraS8WRvslJE8jVzg

Parameters:
```yaml
previous-group-ids: ["QDc7Pm4cSCy70tXhBUW-1w", "B04UrSOnT_iv0WqSw-yShg", "H3zT8CiTRKafIh4C9lba-w"]
```

It starts from one of the later stages, filtering distilled corpus and reuses all the previous tasks including the leaf ones like "analyze" and "evaluate"

Fixes #1129 #728 #719 #1025

[skip ci]


taskcluster/translations_taskgraph/actions/train.py
```diff
@@ -58,6 +57,40 @@ def validate_pretrained_models(params):
             )
 
 
+def get_descendants(task_id, queue, res=None):
```
bhearsum, line 53: If you feel like it, upstreaming this into https://github.com/taskcluster/taskgraph/blob/ec4a5ae2a0fd35e324d9531cb896023a5465eb79/src/taskgraph/util/taskcluster.py#L479 would be nice. Totally up to you.

taskcluster/translations_taskgraph/actions/train.py
```diff
@@ -452,38 +483,77 @@ def train_action(parameters, graph_config, input, task_group_id, task_id):
...
-        if "previous_group_ids" not in input:
-            raise Exception(
-                "'previous_group_ids' is required to use 'start-stage' (otherwise we can't skip earlier tasks)"
-            )
+    previous_group_ids = input.pop("previous-group-ids", None)
+    if previous_group_ids:
+        # Resume the pipeline by reusing the completed tasks from previous task groups
+        import taskcluster
+
+        TC_MOZILLA = "https://firefox-ci-tc.services.mozilla.com"
```
bhearsum: You should be able to find this in `TASKCLUSTER_ROOT_URL` in the environment rather than hardcoding it.

taskcluster/translations_taskgraph/actions/train.py
bhearsum, line 53: Is it possible that this may be called with the same `task_id` more than once? I don't think so....but if it is, you should memoize this function to avoid repeated work.

taskcluster/translations_taskgraph/actions/train.py
```diff
@@ -452,38 +483,77 @@ def train_action(parameters, graph_config, input, task_group_id, task_id):
...
+                    label.startswith("Action")
+                    or label.startswith("Decision")
+                    or label.startswith("PR")
+                ):
+                    continue
+                tasks_to_add[label] = task_id
+
+            logger.info(f"Added ancestor tasks`: {json.dumps(tasks_to_add, indent=2)}")
+
+            # Optionally rerun the tasks with the specified prefix and their descendants by removing them from the existing tasks
```
bhearsum, line 573: I think I'm a bit unclear on why you need to fetch descendants. As far as I understand things, they will always be rerun if one of their upstreams is rerun. I suspect you could get rid of the descendant walking and this would behave exactly the same way.

Am I missing something here?

bhearsum (commented):
Hooray, I'm happy you were able to get this working! There's nothing truly blocking here, but I'd like to understand the descendant walking more before I stamp it. @gregtatum should probably take a look as well, seeing as he's the other primary user of the pipeline.

Do you want to consider adding this is a distinct/parallel option to the existing resume method, to have something to fall back on if issues are discovered in production?

taskcluster/translations_taskgraph/actions/train.py
evgenyrp: I didn't think about it this way. My goal was to produce the correct `existing_tasks` that include only the tasks that should be reused. You're right, it makes sense that if one of the intermediate steps is missing, all its dependencies in the existing tasks will be ignored and rescheduled. Do you think we should rely on this mechanism and simplify the code by removing the descendant retrieval code, or is it still better to produce the correct existing tasks as we already have the code to do this?

bhearsum: You can keep it as is as far as I'm concerned, but unless I'm missing something, there's no way that any tasks that are descendants of one of the prefixed tasks would ever be re-used?

bhearsum (approved):

taskcluster/translations_taskgraph/actions/train.py
evgenyrp: This check prevents it:

```python
       if dep_id in res:
            continue

        res[dep_id] = dep_name
        # recurse down to this child’s dependents
        get_descendants(dep_id, queue, res)
```

taskcluster/translations_taskgraph/actions/train.py
evgenyrp: If we remove the start task explicitly, yes, if we don't, everything completed will be reused.

taskcluster/translations_taskgraph/actions/train.py
bhearsum: Ah, indeed!

evgenyrp:
Greg will be out for some time, so I'm going to merge it now to start using it and address his feedback later if there is any.

taskcluster/translations_taskgraph/actions/train.py
evgenyrp: Yeah, it's a good idea. I was surprised there's `get_ancestors` method but no `get_descendants` one.

bhearsum: They're added as needed. `get_ancestors` was added when I built the earlier continuation logic :)

evgenyrp: The issue is it's in `taskgraph` package, but I'm using `taskcluster` lib to get descendants (Queue etc.), so it doesn't belong there.

PR #1135, opened 2025-05-29
Improve the memory of the Wasm inference engine

gregtatum:
I would estimate the stddev for these measurements is around +/- 5mb.

## Baseline

Inference process memory: 276mb es-en tiny non-pivot
Inference process memory: 853mb ja-ru base pivots
Inference process memory: 560mb base-memory pivots

## This patch

Inference process memory: 177mb (-99mb) es-en tiny non-pivot
Inference process memory: 530mb (-323mb) ja-ru base- pivots
Inference process memory: 362mb (-198mb) base-memory pivots

### DIFF
--- a/inference/marian-fork/src/common/binary.cpp
+++ b/inference/marian-fork/src/common/binary.cpp
@@ -176,7 +176,7 @@ void saveItems(const std::string& fileName,
     headers.push_back(Header{item.name.size() + 1,
                              (uint64_t)item.type,
                              item.shape.size(),
-                             item.bytes.size()}); // binary item size with padding, will be 256-byte-aligned
+                             item.bytes->size()}); // binary item size with padding, will be 256-byte-aligned
   }
 
   uint64_t headerSize = headers.size();
@@ -204,7 +204,7 @@ void saveItems(const std::string& fileName,
 
   // Write out all values
   for(const auto& item : items)
-    pos += out.write(item.data(), item.bytes.size()); // writes out data with padding, keeps 256-byte boundary. 
+    pos += out.write(item.data(), item.bytes->size()); // writes out data with padding, keeps 256-byte boundary. 
                                                       // Amazingly this is binary-compatible with V1 and aligned and 
                                                       // non-aligned models can be read with the same procedure.
                                                       // No version-bump required. Gets 5-8% of speed back when mmapped.
--- a/inference/marian-fork/src/common/binary.cpp
+++ b/inference/marian-fork/src/common/binary.cpp
@@ -28,34 +28,50 @@ const T* get(const void*& current, uint64_t num = 1) {
 }
 
 void loadItems(const void* current, std::vector<io::Item>& items, bool mapped) {
+  uint64_t totalBytesLoaded = 0;  // Track total bytes loaded
+
   uint64_t binaryFileVersion = *get<uint64_t>(current);
   ABORT_IF(binaryFileVersion != BINARY_FILE_VERSION,
            "Binary file versions do not match: {} (file) != {} (expected)",
            binaryFileVersion,
            BINARY_FILE_VERSION);
 
-  uint64_t numHeaders = *get<uint64_t>(current); // number of item headers that follow
-  const Header* headers = get<Header>(current, numHeaders); // read that many headers
+  totalBytesLoaded += sizeof(uint64_t);  // Account for binaryFileVersion
+
+  uint64_t numHeaders = *get<uint64_t>(current);  // number of item headers that follow
+  totalBytesLoaded += sizeof(uint64_t);           // Account for numHeaders
+
+  const Header* headers = get<Header>(current, numHeaders);  // read that many headers
+  totalBytesLoaded += sizeof(Header) * numHeaders;           // Account for headers
+
+  if(items.size() == numHeaders) {
+    // These items are already loaded.
+    return;
+  }
 
   // prepopulate items with meta data from headers
   items.resize(numHeaders);
   for(int i = 0; i < numHeaders; ++i) {
     items[i].type = (Type)headers[i].type;
     items[i].name = get<char>(current, headers[i].nameLength);
+    totalBytesLoaded += headers[i].nameLength;  // Account for item name bytes
     items[i].mapped = mapped;
   }
 
   // read in actual shape and data
   for(int i = 0; i < numHeaders; ++i) {
     uint64_t len = headers[i].shapeLength;
-    items[i].shape.resize(len); 
-    const int* arr = get<int>(current, len); // read shape
-    std::copy(arr, arr + len, items[i].shape.begin()); // copy to Item::shape 
+    items[i].shape.resize(len);
+    const int* arr = get<int>(current, len);            // read shape
+    totalBytesLoaded += len * sizeof(int);              // Account for shape bytes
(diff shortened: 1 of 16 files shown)


inference/marian-fork/src/common/binary.cpp
```diff
@@ -65,29 +81,44 @@ void loadItems(const void* current, std::vector<io::Item>& items, bool mapped) {
...
         cpu::integer::prepareAndTransposeB<Type::int16>(items[i], ptr);
       }
     } else {
-      std::copy(ptr, ptr + len, items[i].bytes.begin());
+      resize(headers[i].dataLength);
+      std::copy(ptr, ptr + headers[i].dataLength, items[i].bytes->begin());
     }
   }
+
+  LOG(info, "[memory] Model data loaded in: {}", totalBytesLoaded);
```
gregtatum, line 121: This was originally a debug log. We should work on hooking up the logs to a proper pref so that we can get them.

inference/src/translator/translation_model.cpp
```diff
@@ -100,7 +100,21 @@ void TranslationModel::loadBackend(size_t idx) {
...
   graph->forward();
+
+  // At this point the ExpressionGraph has consumed the `std::vector<marian::io::Item>`
+  // and converted them to `Tensor`s. This happens the first time that
+  // `ExpressionGraph::forward` is called. It is relatively safe to clear the `Item`s owned
+  // by the scorer since they will not practically be used again. This will free up memory
+  // in the memory-constrained environment of the browser.
+  for (auto scorer : scorerEnsemble) {
+    scorer->clearItems();
+  }
```
nordzilla, line 113: > Thought

[ScorerEnsemble](https://github.com/mozilla/translations/blob/2fe6b2b8c1ecfc0401f063f08d53744cdbde31ac/inference/src/translator/translation_model.h#L131) is an alias to `std::vector<Ptr<...>>`, and [Ptr](https://github.com/mozilla/translations/blob/2fe6b2b8c1ecfc0401f063f08d53744cdbde31ac/inference/marian-fork/src/common/definitions.h#L100-L101) is an alias to `std::shared_ptr`.

---

> Suggestion (optional)

It would be slightly more efficient to take a reference here: `auto& scorer`. 

Just plain `auto` will copy the scorer, which is safe and fine, but it will increment the `shared_ptr` count, just to decrement it again immediately. 

Taking a reference to the `shared_ptr` will just be slightly more efficient by using the same `shared_ptr` that is already in the vector.

inference/marian-fork/src/tensors/tensor_allocator.h
```diff
@@ -13,7 +13,13 @@ class TensorAllocator {
 private:
   const size_t CHUNK = 128;
   const size_t MBYTE = 1024 * 1024;
+#if defined(WASM)
+  // Drastically reduce the size of allocation leaps as Wasm is a heavily memory
+  // constrained environment.
+  const size_t GROW = MBYTE;
+#else
```
nordzilla, line 20: > Question

This is effectively the `.patch` file that you waited to re-apply in your previous PR, right, just gated on `WASM` now?

inference/marian-fork/src/common/binary.cpp
```diff
@@ -65,29 +81,44 @@ void loadItems(const void* current, std::vector<io::Item>& items, bool mapped) {
                               // If this is not set, we trigger node_initializers.cpp:186. This one just assigns the memory ptr to the tensor if set to true, but at the moment
                               // We are preparing some things on demand (the bottom portion of this code). Once we stop doing that, we can use the full mmap codepath
                               // Also when using the full mmap codepath, we need to uncomment expression_graph.h:582
-    uint64_t len = headers[i].dataLength;
-    items[i].bytes.resize(len);
```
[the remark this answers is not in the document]
nordzilla: > Thought

This is the part of the code that I don't have the best grasp on. I haven't worked with this part of the code myself.

If I'm understanding this code correctly, this would do a resize up-front regardless of the type, and then later it might be resized _again_ if it's a "Webm" type.

The new code only resizes once, exactly when it is needed, and uses the lambda function to ensure it is tracked in `totalBytesLoaded.

inference/marian-fork/src/common/binary.cpp
```diff
@@ -204,7 +204,7 @@ void saveItems(const std::string& fileName,
 
   // Write out all values
   for(const auto& item : items)
-    pos += out.write(item.data(), item.bytes.size()); // writes out data with padding, keeps 256-byte boundary. 
+    pos += out.write(item.data(), item.bytes->size()); // writes out data with padding, keeps 256-byte boundary. 
                                                       // Amazingly this is binary-compatible with V1 and aligned and 
                                                       // non-aligned models can be read with the same procedure.
                                                       // No version-bump required. Gets 5-8% of speed back when mmapped.
```
nordzilla: > nitpick

The indentation of the other comments is slightly off now.

inference/marian-fork/src/common/io_item.h
```diff
@@ -9,19 +9,72 @@ namespace marian {
...
+      mapped = other.mapped;
+      name = std::move(other.name);
+      shape = std::move(other.shape);
+      type = other.type;
+
+      other.ptr = nullptr;
+      other.mapped = false;
+    }
+    return *this;
+  }
```
nordzilla, line 71: > praise

Thank you for checking for self-assignment, and nulling out the old owner's members.

nordzilla (approved):
> Praise

Thanks for taking the time to make these changes. 

---

Everything here looks good to me.

I just have a few clarification comments and non-blocking nitpicks.

---

**Additional Review Steps Taken**

* Our [WASM tests](https://github.com/mozilla/translations/pull/1135/checks?check_run_id=43210715932) within this repository pretty extensively test a number of different inputs/outputs with our models, which helps give us confidence that this has not changed any characteristics that would affect the output of the translation.

* I hooked up these changes to my local build of Firefox and tested them manually end-to-end with no issues. 

* I enabled these changes to our Firefox CI on the Try servers, and ran a performance comparison, verifying the memory improvements on macOS, Linux, and Windows ([PerfCompare](https://perf.compare/compare-results?baseRev=04a89b0b8abbb9ca79053be031c17289d0d78642&baseRepo=try&newRev=d42a997dbb1500819121595dadeef969c2452359&newRepo=try&framework=15&sort=delta%7Cdesc)).

---

These changes reduce the memory footprint of our models at no cost to performance, and no change to the translated output.

inference/marian-fork/src/tensors/tensor_allocator.h
gregtatum: Yes, that's correct.

inference/marian-fork/src/common/binary.cpp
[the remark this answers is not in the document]
gregtatum: Yes, that's the intent of the new code. This area lit up like crazy using `dhat` for blaming allocations, so this makes it much more targeted. The lambda means you don't have to repeat code, and it should get inlined by the compiler.

inference/src/translator/translation_model.cpp
gregtatum: It'd be nice to go around and delete all of these copy constructors.

PR #1141, opened 2025-06-06
Use commas rather than newlines for CUDA_VISIBLE_DEVICES

gregtatum:


gregtatum:
I canceled the rest of the run since the bicleaner step didn't really tell me anything.

eu9ene (approved):
I don't know what's the difference but it worked in the previous format too

PR #1145, opened 2025-06-09
Use 10 training updates in CI tasks

gregtatum:
I was doing some profiling around task times, and teacher training in CI was 13 minutes, with 11 minutes being training. Seems like we just need a few updates to prove it works. I reduced the updates to a 1/5 of the original.

https://share.firefox.dev/3HLGYMg

### DIFF
--- a/pipeline/train/train.py
+++ b/pipeline/train/train.py
@@ -2,6 +2,8 @@
 Run training using Marian and OpusTrainer.
 """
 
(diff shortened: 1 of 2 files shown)


gregtatum:
It doesn't change anything. Seems like OpusTrainer doesn't respect this. Not sure how to do this easily, so closing.

Issue #994, opened 2025-01-15
Investigate using an LLM as a teacher


evgenyrp:
We should do a proof of concept for knowledge distillation from an LLM to our standard student model. The main benefit of this if it works is we won't need to deal with parallel data cleaning and training teacher models. All this can be quite challenging, especially for lower-resource languages.

This would require:
- Estimate costs for different LLMs and APIs we can use
- Run quality evaluation for those to see which model would provide the best cost/quality tradeoff
- Choose a mix of monolingual data to translate
- Run translation with an LLM 
- Train a regular student model on this data
- Try different LLMs and corpus of different sizes (for example, 10M and 50M sentences)

Folks from WMT also suggested we can try pre-training the student on parallel OPUS corpus as is and then finetune on a smaller but high-quality LLM-produced corpus to make it more cost efficient.

eu9ene:
Related to #767

evgenyrp:
Thanks for the link! I looked at their benchmarks and compared them to WMT24++ paper. At least for en-ru (the language pair I tested most), it didn't improve compared to the previous version. And it was significantly worse than Gemma 3 27B, according to my comparison based on MetricX24. Models from Unbabel, including XComet have NC license, so I can't use them. I just look at the papers and assume they used the biggest XComet/MetricX24 model for evals, the same as I used for benchmarking Gemma3 27B.

Issue #767, opened 2024-07-26
Investigate using LLMs to generate training data


marco-c:
E.g. it could be useful for generating datasets of short sentences to help with #215.

evgenyrp:
Based on #1147, we already have a pipeline to generate data with LLMs and a way to diversify the input texts. The diverse English corpus includes all kinds of things, including shorter phrases. So I'm closing this one.

PR #1147, opened 2025-06-10
Add LLM running code and docs

evgenyrp:
I'm adding this experimental but mostly reproducible code to a separate folder so that it's not lost when we want to continue experimenting with LLMs or someone else wants to try it.

Also, maybe I missed something else that we can try. Please look at it on a high level and let me know what you think.

I produced 1M and 10M parallel sentences datasets with Gemma 27b and greedy decoding and used them for fine-tuning, which led to some improvement in the COMET scores for en-ru (student pre-trained on the best 50M NLLB).  Of course not as significant as with NewsPALM for en-de that I also tried (NewsPALM was produced with significantly more compute: bigger model and using 512 candidates for MBR and QE reranking). I'm not reporting the full results of the fine-tuning experiments here as they are not finalized.

Related to #1037 #994 #767 

[skip ci]


experiments/llmaat/README.md
```diff
@@ -0,0 +1,112 @@
...
+This work follows the paper [Introducing the NewsPaLM MBR and QE Dataset:
+LLM-Generated High-Quality Parallel Data Outperforms Traditional
+Web-Crawled Data](https://arxiv.org/pdf/2408.06537).
+
+It also uses the evaluation dataset and the prompt from [WMT24++: Expanding the Language Coverage of WMT24 to 55 Languages & Dialects](https://arxiv.org/html/2502.12404v1).
+
+
+## Selecting a corpus
+
+The idea is to have a diverse monolingual dataset to translate by an LLM.
```
ZJaume, line 16: Is there a reason why this is only focused on monolingual data? If the goal is to use LLM as a teacher, I would generate distilled data at least from all parallel corpora available (or sample big corpora), then add monolingual data. Parallel data could cover certain domain areas that are not well covered by monolingual data? Also it has the advantage that if the LLM inference produces an n-best list (regardless of whether it is greedy decoding or beam decoding), the hypotheses can be filtered by chrF score like we are doing right now with distilled data (potentially removing some hallucinations).

evgenyrp: For sure, we could go with the same path we currently use for teacher distillation. The idea here was to pre-train on super clean existing parallel sentences and augment with synthetic high-quality examples. We can go with purely LLM-generated examples, similar to how we do distillation now and then we can do n-best reranking in a similar way. The limitation is inference costs. If we pretrain on the OPUS corpus, we need to generate a lot smaller fine-tuning dataset with an LLM. For example, 50M OPUS + 10M synthetic. Otherwise, we would need to generate 60M examples. We can definitely try that too.

experiments/llmaat/README.md
```diff
@@ -0,0 +1,112 @@
+# LLM as a teacher
```
gregtatum:
```suggestion
# LLM as a teacher (llmaat)
```

gregtatum: I found the `llmaat` acronym confusing.

PR #1157, opened 2025-06-20
Add url mismatch filter + fixes and cleanup

### LINKED ISSUE #569: Delete old cleaning scripts
We already have OpusCleaner defaulting to true on the production config, but after some time we should go ahead and delete the old cleaning procedures.

### LINKED ISSUE #758: Add a cleaning rule for URL names, such as Amazon.com -> Amazon.it
Similar to #736, we should discard sentences that translate the domain suffix of a website, like Amazon.com -> Amazon.it

With a regex such as `/[a-z]+\.com\b/` we could identify a URL on the English side, and ensure that it's matched on the other language.

Edit: Also, we have other English speaking countries to consider, like Amazon.co.uk

evgenyrp:
- Add a new `url_mismatch` filter to all configs
- Fix the issues with zh custom configs (add all the filters from default)
- Remove old cleaning scripts and `use-opuscleaner` setting
- Remove old experimental custom filters as we don't use them (experiments didn't show much difference)
- Add a unit test for OpusCleaner
- Update docs

Note: `opuscleaner/configs/ru-en/opus_ELRC-3075-wikipedia_health-v1.filters.json` and `pipeline/clean/opuscleaner/configs/opus_ada83-v1.filters.json` are in fact added to test the filter generation code and almost equal to the default filter.

fixes #758 
fixes #569


tests/test_clean.py
```diff
@@ -0,0 +1,111 @@
...
+
+        with open(new_config_path, "w") as f:
+            f.write(new_config)
+        return new_config_path
+    elif trg_lang == "zh":
+        return os.path.abspath(os.path.join(FIXTURES_PATH, "config.pytest.enzh.yml"))
+    return None
+
+
+def read_lines(path):
```
gregtatum, line 53: You can use the `read_lines` utility in pipeline/common/downloads.py.

tests/test_clean.py
```diff
@@ -0,0 +1,111 @@
...
+    [
+        # tests langauge-specific config
+        ("opus", "en", "zh", "NeuLab-TedTalks_v1"),
+        # tests dataset-specific config
+        ("opus", "en", "ru", "ELRC-3075-wikipedia_health_v1"),
+        # tests default config
+        ("opus", "en", "ru", "ELRC_2922_v1"),
+    ],
+)
+def test_clean_parallel(importer, src_lang, trg_lang, dataset, data_dir):
```
gregtatum, line 74: Thanks for the `run_task` tests!

pipeline/clean/opuscleaner/configs/en-zh/WikiMatrix-v1.filters.json
```diff
@@ -3,6 +3,126 @@
   "files": [
   ],
   "filters": [
+    {
```
gregtatum, line 6: Are all these config changes coming from your CJK training run?

gregtatum (approved):

pipeline/clean/opuscleaner/configs/en-zh/WikiMatrix-v1.filters.json
evgenyrp: Jaume added these filters based on his manual experiments in OpusCleaner, and we used them for training. What we missed is that the rest of the default filters should also be present. I'm fixing it here by copying them from default.config.json for CJK.

tests/test_clean.py
evgenyrp: Yes, but it requires using it with `with` an iterator ,which would make the code in the tests a lot more verbose, so I'd keep this simple function for now

Issue #248, opened 2023-11-08
Consider using a more precise language identification model to clean datasets


marco-c:
For example, paracrawl was built by using CLD2. CCMatrix by using fastText.
These models don't always work so well, for example with short sentences (see https://github.com/pemistahl/lingua-py#4-how-good-is-it).

We could consider using a more precise, even if more expensive, model. In addition, we could try to use a dictionary-lookup (in case of short sentences consisting of one or two words, check if the words are present in the dictionary for a language). We could even build a model that uses the results of another + dictionary-lookup result as a feature.

marco-c:
Some alternatives we could consider:
https://github.com/pemistahl/lingua-py
https://github.com/mbanon/fastspell (this implements the dictionary idea)

marco-c:
I added fastspell to lingua-py's benchmarks and run it for Italian, setting "it" to be similar to "es", "fr", "pt", and "en". The results are very good.
fastText:
```
>>> Accuracy on average: 88.77%

>> Detection of 1000 single words (average length: 8 chars)
Accuracy: 74.10%
Erroneously classified as English: 9.90%, Spanish: 4.70%, Portuguese: 3.20%, French: 1.70%, German: 1.30%, Unknown: 1.00%, Esperanto: 0.60%, Turkish: 0.50%, Dutch: 0.40%, Finnish: 0.40%, Latin: 0.40%, Hungarian: 0.30%, Polish: 0.30%, Romanian: 0.30%, Croatian: 0.20%, Indonesian: 0.20%, Swedish: 0.20%, Basque: 0.10%, Czech: 0.10%, Russian: 0.10%

>> Detection of 1000 word pairs (average length: 16 chars)
Accuracy: 92.40%
Erroneously classified as English: 3.30%, Portuguese: 1.20%, Spanish: 0.90%, French: 0.30%, German: 0.30%, Finnish: 0.20%, Romanian: 0.20%, Swedish: 0.20%, Turkish: 0.20%, Basque: 0.10%, Croatian: 0.10%, Dutch: 0.10%, Esperanto: 0.10%, Latin: 0.10%, Persian: 0.10%, Slovene: 0.10%, Unknown: 0.10%

>> Detection of 1000 sentences (average length: 123 chars)
Accuracy: 99.80%
Erroneously classified as English: 0.10%, French: 0.10%
```

fastSpell conservative:
```
>>> Accuracy on average: 91.97%

[...]

PR #1162, opened 2025-06-24
Fix mono clean

### LINKED ISSUE #1161: mono cleaning is broken by #1157
I was running some integration tests today and discovered [this error](https://stage.taskcluster.nonprod.cloudops.mozgcp.net/tasks/Yz4rBYjcSH6g4emF4Ka2wA/runs/1) in a mono clean task:
```
Traceback (most recent call last):
  File "/builds/worker/checkouts/vcs/pipeline/clean/tools/clean_mono.py", line 14, in <module>
    from clean_parallel import CHARS
ModuleNotFoundError: No module named 'clean_parallel'
```

The file it's trying to import was removed in https://github.com/mozilla/translations/pull/1157. Unfortunately, this file was [not part of the cache resources](https://github.com/mozilla/translations/blob/19ec4b0227f1c645bef691b3582d9db18a88c8a2/taskcluster/kinds/corpus-clean-mono/kind.yml#L32), which would've forced it to run in that PR and caught this :(. If it gets re-added, it should probably get added to that list.

evgenyrp:
fixes #1161


bhearsum (approved):
I imagine you want someone else's review on this as well, but embedding this rather than restoring the deleted file seems like an OK thing to do to me.

pipeline/clean/tools/clean_mono.py
```diff
@@ -5,13 +5,737 @@
...
 MIN_LENGTH = 2  # minimum number of words in a sentence
 MAX_LENGTH = 150  # maximum number of words in a sentence
 
 RATIO_ALPHA_WORDS = 0.4  # minimum fraction of "real" words in a sentence
 RATIO_ALPHA_CHARS = 0.5  # minimum fraction of alpha characters in a sentence
 
-from clean_parallel import CHARS
+# Copied from opuscleaner (`from opuscleaner.filters.clean_common import CHARS` doesn't work because of the conflict with "filters" module)
+# TODO: modify opus cleaner to import it here
+CHARS = {
```
gregtatum: Can we fix this the real way? I'd prefer not to create technical debt when we don't have to.

pyproject.toml
```diff
@@ -83,6 +83,7 @@ sacremoses = "0.1.1"
 hanzidentifier = "1.2.0"
 OpenCC = "1.1.9"
 PyICU = "2.8.1"
+fasttext = "*"
```
gregtatum: Is this a mistake? I don't see where this is being used or is needed. Plus the version should be locked to a specific major/minor version so we don't have accidental breakage.

gregtatum (changes requested):
I'm requesting changes based on my two comments here.

pipeline/clean/tools/clean_mono.py
evgenyrp: Sure, we can do it the right way right away. This was supposed to be a hot fix to repair CI in main.

pyproject.toml
evgenyrp: It's needed in the tests. We can for sure lock it as well.

evgenyrp:
Blocked by https://github.com/hplt-project/OpusCleaner/pull/168

pipeline/clean/tools/clean_mono.py
evgenyrp: fixed

pyproject.toml
evgenyrp: Pinned, but there are a bunch of issues with fasttext: incompatibility with numpy v2, not compiling under poetry, it is unsupported now, and the repo is archived, there's an unofficial weird release 0.9.3.  I used a prebuilt wheel here, the same as in hunspell/fastspell (we already install hunspell in taskcluster). We should have another look at all this in #248, I assume we'll switch to fastspell.

PR #1165, opened 2025-06-27
Catalan and Dutch configs from short sentences experiments

ZJaume:
These are pretty much the same as they were autogenerated.

[Link](https://gregtatum.github.io/taskcluster-tools/src/training/?dashboardName=Issues+with+short+sentences&taskGroupIds=R3IOtmXnQvyQvtf41OPEeg%2CBOgRQT7_TmedlWYqcvaGNw&showAll=false) to the dashboard.


evgenyrp:
Not critical, but we also typically upload the autogenerated versions as well, so that it's easier to compare what was changed back then.

PR #1168, opened 2025-07-01
Add config validation to validate the config whenever the train action is triggered

gregtatum:
I've been using this on the training run to actually run the validation checks. They are currently only run in the Taskcluster UI, but this makes it so that they are run in the pipeline. All configs that are checked in are run through it to validate them. We had duplicated documentation for the keys, so I centralized all the documentation in config.prod.yml.


taskcluster/translations_taskgraph/actions/train.py
```diff
@@ -77,361 +91,418 @@ def get_descendants(task_id, queue, res=None):
...
-    order=500,
-    context=[],
-    available=can_train,
-    schema=lambda graph_config: {
+def get_train_parameters(
+    parameters_obj: Parameters, training_config: dict[str, Any]
+) -> Parameters:
+    parameters: dict[str, Any] = dict(parameters_obj)
+
+    previous_group_ids = training_config.pop("previous-group-ids", None)
```
gregtatum, line 99: This is all moved code, but the diff couldn't figure that out. You should see it duplicated in the removal below.

taskcluster/translations_taskgraph/actions/train.py
```diff
@@ -475,160 +547,57 @@ def get_descendants(task_id, queue, res=None):
...
-def train_action(parameters, graph_config, input, task_group_id, task_id):
-    # TODO: Add a whack load of verification here. Things such as:
-    # - datasets all exist
-    # - locale pair exists for each dataset
-    # - stage is valid
-    # etc.
-
-    parameters = dict(parameters)
-
-    previous_group_ids = input.pop("previous-group-ids", None)
```
gregtatum, line 522: This code was moved above.

taskcluster/translations_taskgraph/actions/train.py
```diff
@@ -475,160 +547,57 @@ def get_descendants(task_id, queue, res=None):
...
                         "additionalProperties": {
                             "type": "string",
-                            # TODO: add snakepit type(s) when they are brought online
                             "enum": ["gcp-standard", "gcp-spot"],
                         },
                     },
                 },
             },
         },
-        "required": [
```
bhearsum, line 505: AIUI, removing these from `required` means that the UI won't complain if the schema is wrong. I'm assuming that this minor inconvenience is worthwhile because it allows you to get rid of the duplication?

gregtatum: The `make_schema_strict` function reverses it so that everything is required by default, and you have to mark things as optional. I was finding the previous behavior was error prone when working with validating configs.

gregtatum: Also, these days I'm not using the UI, since the trigger training has more validation and checks in it that you are doing the correct thing. This validation runs on trigger training.

tests/test_config_validation.py
```diff
@@ -0,0 +1,40 @@
...
+from jsonschema import validate
+
+
+def load_yaml(path: str):
+    with open(path, "r") as file:
+        yaml_text = file.read()
+        return yaml.safe_load(yaml_text)
+
+
+@pytest.mark.parametrize(
```
evgenyrp, line 19: Nice! We had quite a few issues with outdated configs in the past.

taskcluster/translations_taskgraph/actions/train.py
evgenyrp: Yeah, moving this function makes it hard to review this part, and it will be impossible to trace the history of the individual lines. I'd rather move it back until you think it's absolutely necessary.

evgenyrp (approved):
Approving, but I have an optional suggestion.

taskcluster/translations_taskgraph/actions/train.py
gregtatum: I looked at moving it back, and it's really hard since I'm re-using this function. I could do something a bit cleaner with individual commits to preserve things, but since we squash them it'll get lost.

PR #1169, opened 2025-07-02
Always specify the best and first number for shortlists 

gregtatum:
I read through Marian's source code to see what was going on with shortlisting, and there are two behaviors based on whether the format is binary or text. I changed our code to just be explicit on providing our hyper parameters as hard coded values so that things won't get screwed up.

I also disabled shortlisting in the eval since we aren't using it in production right now.

### DIFF
--- a/pipeline/eval/eval.py
+++ b/pipeline/eval/eval.py
@@ -187,12 +187,20 @@ def main(args_list: Optional[list[str]] = None) -> None:
         marian_extra_args = [*marian_extra_args, "--vocabs", args.vocab_src, args.vocab_trg]
 
     if args.shortlist:
-        # No arguments to the shortlist, so default ones are used
-        # this way it doesn't matter if the shortlist is binary or text
-        # because they have different arguments
-        # text shortlist args: firstNum bestNum threshold
-        # binary shortlist (has the arguments embedded) args: bool (check integrity)
-        marian_extra_args = marian_extra_args + ["--shortlist", args.shortlist]
+        # Marian has different behaviors for the CLI based on whether a shortlist is
+        # the binary format, which has the hyper parameters embedded within it, and
+        # whether or not it is a text one, which must have the hyperparameters provided.
+        # Just be explicit here with hard coded values.
+        marian_extra_args = marian_extra_args + [
+            "--shortlist",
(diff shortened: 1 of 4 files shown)


bhearsum (approved):

pipeline/eval/eval.py
```diff
@@ -187,12 +187,20 @@ def main(args_list: Optional[list[str]] = None) -> None:
...
-        marian_extra_args = marian_extra_args + ["--shortlist", args.shortlist]
+        # Marian has different behaviors for the CLI based on whether a shortlist is
+        # the binary format, which has the hyper parameters embedded within it, and
+        # whether or not it is a text one, which must have the hyperparameters provided.
+        # Just be explicit here with hard coded values.
+        marian_extra_args = marian_extra_args + [
+            "--shortlist",
+            args.shortlist,
+            # first number
+            "50",
```
evgenyrp, line 198: I didn't look at this code in Marian myself, so I'd defer this review to @ZJaume

taskcluster/kinds/evaluate-quantized/kind.yml
```diff
@@ -111,7 +111,6 @@ tasks:
                     --dataset_prefix    "$MOZ_FETCHES_DIR/{dataset_sanitized}"
                     --vocab_src          "$MOZ_FETCHES_DIR/vocab.{src_locale}.spm"
                     --vocab_trg          "$MOZ_FETCHES_DIR/vocab.{trg_locale}.spm"
-                    --shortlist         "$MOZ_FETCHES_DIR/lex.s2t.pruned"
```
evgenyrp, line 114: I think we should not disable shortlisting in the pipeline evals. It's a bug in the Firefox integration that should be fixed. We can just assume that the prod results are slightly better. Also, when properly implemented, we saw that it almost doesn't affect quality. 

Unless we make a big decision to disable shortlisting permanently. Then it should be made optional and disabled everywhere, including in the models repo evals.

evgenyrp (changes requested):

gregtatum:
I'm going to close this, as I was a bit nervous to merge without more investigation, and I didn't handle feedback.

Issue #771, opened 2024-07-29
Reduce monolingual data for da-en to investigate distillation performance


gregtatum:
An experiment for #231

da-en is one of our best models from the spring-2024 run. The teacher ensemble had a COMET score of 0.9013. The student COMET was 0.8950, with a tiny -0.0063 gap. In order to understand the role of the quantity of monolingual data this experiment will artificially reduce the amount of data to observe the effects on distillation.

### Experiment splits:

| Metric                       | Value |
| ---------------------------- | -----:|
| Teacher COMET                | 90.13 |
| Original parallel corpus     | 77,673,571 |
| Original distillation corpus | 177,346,837 |
| Monolingual corpus           | 122,958,567 |

(Edit: I changed my technique to limit the distillation corpus size, rather than strictly monolingual size, so my percentages are off from my original 100%, 75%, 50%, and 25% splits. The distillation corpus is formed from the `da` side of the parallel corpus and the `da` monolingual data.)

| Sentences     | Percent | Student COMET | Teacher Gap | Vs Baseline |
| -------------:| -------:| -------------:| -----------:| -----------:|
| `177,346,837` | `100%`  | `89.50`       | `-0.63`     | -           |
|  `92,218,925` | `52%`   | `89.44`       | `-0.69`     | `-0.06`     |
|  `61,479,283` | `35%`   | `89.31`       | `-0.82`     | `-0.19`     |
|  `30,739,641` | `17%`   | `89.44`       | `-0.69`     | `-0.06`     |
|   `1,000,000` | `0.5%`  | `85.79`       | `-3.71`     | `-3.71`     |
|      `10,000` | -       | `39.57`       | `-50.56`    | `-49.93`    |

| Sentences     | Percent | Student chrF | Vs Baseline |
| ------------- | ------- | ------------ | ----------- |
| `177,346,837` | `100%`  | `70.74`      | -           |
|  `92,218,925` |  `52%`  | `70.67`      | `-0.07`     |
|  `61,479,283` |  `35%`  | `70.75`      | `+0.01`     |
|  `30,739,641` |  `17%`  | `70.63`      | `-0.11`     |
|   `1,000,000` |  `0.5%` | `66.73`      | `-4.01`     |
|      `10,000` |  -      | `26.94`      | `-43.8`     |

### Hypothesis:

Reducing the monolingual data will decrease the COMET score, but there is a benefit as it would decrease cost by not having to synthesize as much training data for distillation.

### Results

[...]

PR #1170, opened 2025-07-02
Limit the monolingual data in the prod config

gregtatum:
Based on my numbers from #771, and the cost of this step, I reduced the default totals for monolingual data in the `h1-2025` run.

### DIFF
--- a/taskcluster/configs/config.prod.yml
+++ b/taskcluster/configs/config.prod.yml
@@ -65,14 +65,14 @@ experiment:
   # Limits the maximum sentences use in the monolingual data for both the source and target languages.
   mono-max-sentences-src:
     # More data for distillation is better for student quality.
-    total: 300_000_000
+    total: 100_000_000
     # Some datasets can be massive (like English NLLB mono has 3 trillion sentences)
     # Limit the per-dataset to make sure there is good data diversity between datasets.
-    per-dataset: 100_000_000
+    per-dataset: 70_000_000
   mono-max-sentences-trg:
     # Limit the number of back-translations
-    total: 200_000_000
-    per-dataset: 100_000_000
+    total: 100_000_000
+    per-dataset: 70_000_000
 
   # HPLT provides a document score in the dataset, and we can filter based on it.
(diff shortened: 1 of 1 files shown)


taskcluster/configs/config.prod.yml
```diff
@@ -65,14 +65,14 @@ experiment:
   # Limits the maximum sentences use in the monolingual data for both the source and target languages.
   mono-max-sentences-src:
     # More data for distillation is better for student quality.
-    total: 300_000_000
+    total: 100_000_000
```
evgenyrp, line 68: It's hard for me to decide on this one. On one hand, my intuition and experience tell me that having more data and especially more of more diverse data always wins. We might not even see the full effect of it on the limited evals datasets. Also I assume this is very language and dataset-dependent. Finally, we optimized inference quite a lot. So I don't know if it's worth it to risk reducing model quality to save marginally on costs.

On the other hand, I see in your experiment with Danish to English (which is an easy language pair btw), even just 1M sentences gave pretty good results, which is quite surprising. So it does plateau at some point, it's just hard to say for sure on what amount of data for which language pair.

I think the original Bergamot recipe aimed to squeeze every fraction of COMET/chrF from student models, that's why it was recommended to use ensembles, backward -> forward distillation etc. We're not doing a competition so cost is more important for us. 

Since retraining high-resource languages is expensive, and we want the best quality for them, I will definitely bump those values back to 300/200M, so I don't know if we want to leave it as a default.

@ZJaume what's your intuition here?

ZJaume: My intuition is reducing to 100M shouldn't hurt much, but If
> we want the best quality for them, 

I guess we should keep it at 200/300 like you say

evgenyrp: Yeah, let's keep 100 as a default, and I'll see how much data I have for each pair and adjust. Those high-resource languages are very important for us, so I'd rather show the models more examples, even though I don't have evidence that they help.

PR #1172, opened 2025-07-02
Choose the augmentation to perform based on script information

gregtatum:
This is a bit bigger work compared to some of the other PRs. The first commit is a bit off topic, but includes some fixes that were hard to rebase around, so I left them in there with this commit.

This work generalizes the CJK and Arabic script changes to use script information rather than specific language checks for what type of augmentations can be run.

The file `pipeline/data/lang_script.py` was built with a combination of CLDR information and Wikipedia sleuthing to build up a big list of potential scripts. It's a bit overkill as we'll most likely never use a lot of the scripts, but I just got a list from CLDR and worked through it.

The opustrainer configs are now built dynamically based on the script features.


gregtatum:
I'm moving this to draft because I forgot about the libicu errors I needed to sort out. I hacked around this in my release branch.

pipeline/data/lang_script.py
```diff
@@ -0,0 +1,319 @@
...
+scripts: dict[str, ScriptInfo] = {
+    "Adlm": {
+        # 𞤀𞤤𞤳𞤵𞤤𞤫 𞤁𞤢𞤲𞤣𞤢𞤴𞤯𞤫 𞤂𞤫𞤻𞤮𞤤 𞤃𞤵𞤤𞤵𞤺𞤮𞤤
+        "name": "Adlam",
+        "type": ScriptType.ALPHABETIC,
+        "bicameral": True,
+    },
+    "Arab": {
+        # العربية
+        "name": "Arabic",
```
evgenyrp, line 112: Using a data class here instead of the fixed dictionary keys would make it easier to use and less error-prone.

pipeline/data/parallel_importer.py
```diff
@@ -120,13 +106,47 @@ def add_alignments(corpus: List[str]) -> List[str]:
...
+
+    assert src_script, "The script info must exist for the src language."
+    assert trg_script, "The script info must exist for the trg language."
+
+    logger.info("src_script " + repr(src_script))
+    logger.info("trg_script " + repr(trg_script))
+
+    modifiers: list[Modifier] = [
+        RemoveEndPunctuationModifier(MIX_PROB),
+        NoiseModifier(MIX_PROB),
```
evgenyrp: The order is important. Noise should go before the placeholder, near the end, otherwise we modify the added noise lines with other modifiers.

pipeline/train/configs/opustrainer/teacher.yml
```diff
@@ -0,0 +1,94 @@
...
+    # arguments here, the alignments will be removed after this augmentation is run.
+    # pipeline/train/train.py will ensure this is the last modifier in the list.
+    - Tags: 0.005
+      custom_detok_src: "icu:{src}"
+      custom_detok_trg: "icu:{trg}"
+      # Only "augment" mode is supported for "icu" tags, which inserts random words
+      # that are aligned in the source and target sentence.
+      augment: 1
+      tag: 0 # Explicitly disable tag mode.
+
```
evgenyrp, line 73: We've lost RemoveEndPunct here

pipeline/train/configs/opustrainer/teacher.yml
evgenyrp, line 1: I think this is a lot better than the 5 copy-pasted configs we had. In the future, we can go further and generate it completely dynamically. I had to hack around it when I needed to fine-tune the student on synthetic data, so having it all in code would make it easier.

pipeline/train/train.py
```diff
@@ -257,39 +261,78 @@ def build_datasets(self) -> None:
...
+            # Bicameral scripts can have their casing augmented.
+            if src_script["bicameral"]:
+                modifiers.extend(modifier_sections["bicameral_src"])
+
+            # Phonemic languages can be misspelled.
+            if is_script_phonemic(src_script["type"]):
+                modifiers.extend(modifier_sections["phonemic_src"])
+
+            # Ensure the "Tags" modifier is last, as it removes the alignments.
+            modifiers.sort(key=lambda modifier: next(iter(modifier)) == "Tags")
```
evgenyrp: Noise should come right before Tags

utils/config_generator.py
```diff
@@ -134,7 +138,8 @@ def update_config(
     else:
         prod_config["continuation"]["models"] = {}
 
-    if is_cjk(source, target):
+    needs_bigger_vocab = {ScriptType.LOGOGRAPHIC, ScriptType.FEATURAL}
+    if src_script["type"] in needs_bigger_vocab or trg_script["type"] in needs_bigger_vocab:
```
evgenyrp: Those languages need a bigger vocab only when we use a shared vocab. If it's split-vocab we use 32k on each side as before.

utils/config_generator.py
```diff
@@ -123,7 +124,10 @@ def update_config(
     experiment["trg"] = target
     experiment["bicleaner"]["dataset-thresholds"] = {}
     # Use separate vocabs for languages with different scripts
-    experiment["spm-vocab-split"] = get_default_script(source) != get_default_script(target)
+
+    experiment["spm-vocab-split"] = not (
```
evgenyrp: Why was this changed? The logic was to use split vocabs if the languages don't share the same script. How is it related to the script being phonemic?

pipeline/train/train.py
```diff
@@ -257,39 +261,78 @@ def build_datasets(self) -> None:
...
+                )
 
-        with open(config_input, "rt", encoding="utf-8") as file:
-            config_text = file.read()
+            config_yaml.update(curriculum[teacher_mode])
 
-        logger.info(f"Applying OpusTrainer config variables: {config_input}")
-        for key, value in self.config_variables.items():
-            logger.info(f" - {key}: {value}")
+        config_text = yaml.safe_dump(config_yaml)
```
evgenyrp: We need a unit test that tests the generated configs for different cases (compares to the one we expect).

evgenyrp (changes requested):
Nice work, the approach is more general now and it's easier to extend to other languages. There are a bunch of small things that we should fix before merging.

pipeline/data/lang_script.py
gregtatum: It's using a [`TypedDict`](https://typing.python.org/en/latest/spec/typeddict.html) which has type checking, so should be equivalent to a dataclass but with slightly different syntax: 

```py
class ScriptInfo(TypedDict):
    name: str
    type: ScriptType
    bicameral: bool
```

This is enforced here:

```py
scripts: dict[str, ScriptInfo] = {
  ...
}
```

pipeline/train/train.py
gregtatum: Ah, that's a good idea, thanks. I didn't think about the testing picture when writing the code.

utils/config_generator.py
gregtatum: Let's discuss this in chat.

evgenyrp: Ok, as we discussed, let's use it for CJK-like languages only. In this case, your condition looks correct, but it's not obvious. Could you add a comment explaining what having at least one of two languages being non-phonemic means?

pipeline/data/lang_script.py
evgenyrp: ok, but why not data class? Does it provide any benefits compared to it? You use it as `script['bicameral']` later, I think `script.bicameral` would be easier.

utils/config_generator.py
evgenyrp: Actually, since we use split vocabs now for those non-phonemic languages, the vocab size always stays 32k.

utils/config_generator.py
gregtatum: I tweaked this slightly to match what we have in production now, with Chinese/Japanese being split, and with Korean having an increased vocab size.

evgenyrp: Well, the idea was to use the same method for Korean as well (split vocabs 32k), it's just the improvement was not big enough to rerun distillation, so I'm not sure if we should make an exception for it.

evgenyrp (approved):

PR #1173, opened 2025-07-02
Add types to tests/test_training.py

gregtatum:
I added types while debugging some failures.

### DIFF
--- a/tests/test_training.py
+++ b/tests/test_training.py
@@ -1,5 +1,6 @@
 import os
 import shutil
+from typing import Any, Optional
 
 import pytest
 import sentencepiece as spm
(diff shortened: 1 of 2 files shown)


tests/test_training.py
```diff
@@ -18,14 +19,17 @@
 )
 
 
-def validate_alignments(corpus_path, vocab_src_path, vocab_trg_path):
-    sp_src = spm.SentencePieceProcessor(model_file=vocab_src_path)
-    sp_trg = spm.SentencePieceProcessor(model_file=vocab_trg_path)
+def validate_alignments(corpus_path: str, vocab_src_path: str, vocab_trg_path: str) -> None:
+    # This module is not well-typed, set it to Any.
+    SentencePieceProcessor: Any = spm.SentencePieceProcessor
```
evgenyrp, line 24: I'm not familiar with this syntax. What does it do?

gregtatum: Setting a local variable like this, and declaring it as `Any` basically says "turn off type checking for this variable". The issue is that the types for the sentencepiece module are incorrect, and so attempting to use them results in errors. `spm.SentencePieceProcessor` as a type exists:

image

But the type itself is wrong, in that the type system says that it doesn't have any keyword arguments:

```py
class SentencePieceProcessor(object):
    def __init__(self):
        _sentencepiece.SentencePieceProcessor_swiginit(self, _sentencepiece.new_SentencePieceProcessor())
    __swig_destroy__ = _sentencepiece.delete_SentencePieceProcessor
```

Here you can see `__init__` takes no arguments.

You could also do `# type: ignore` for specific errors, but I don't like that style as much because it can hide real errors. I prefer just setting something untyped to an `Any` and not do type checking for that specific variable binding.

PR #1178, opened 2025-07-10
feat: run python tests in parallel with pytest-xdist

bhearsum:


bhearsum:
It looks like this isn't the easy win I'd hoped it would be. The first problem I hit was that tests shared the same `DataDir`. After fixing that, I discovered tests depend on a specifically named `DataDir`, eg: https://github.com/mozilla/translations/blob/fb511f71ae77e5579f63240b2ae7b292ae629799/tests/test_translate.py#L140.

I don't have the time to run this sort of thing down at the moment; I'll leave this open in case somebody else wants to as some point.

gregtatum:
I don't think `run_task` tests are parallelizable, as generating the taskgraph happens in the root directory of the project, so if two tasks do it at the same time they will conflict. Perhaps if you could specify the directory that those files get generated it would work.

PR #1195, opened 2025-07-17
Build the marian-fork for use in the pipeline

gregtatum:
Marking this as draft as I validate CI.

### DIFF
--- a/taskcluster/kinds/compile/kind.yml
+++ b/taskcluster/kinds/compile/kind.yml
@@ -1,0 +1,78 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+---
+
+loader: taskgraph.loader.transform:loader
+
+transforms:
+    - translations_taskgraph.transforms.worker_selection
+    - taskgraph.transforms.task_context
+    - taskgraph.transforms.run:transforms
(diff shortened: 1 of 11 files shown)


taskcluster/kinds/compile/kind.yml
```diff
@@ -0,0 +1,84 @@
...
+                type: compile
+                # Since inference/marian-fork is a large subfolder, we don't want to
+                # invalidate the prod builds for any small change in the marian-fork.
+                # Instead, rely on a version number that can be bumped when it needs
+                # to be rebuilt.
+                version: 2
+                resources:
+                    - inference/scripts/build.py
+                    # This matches recursively, but only invalidates the task in CI runs.
+                    - inference/marian-fork/src/command/*
```
bhearsum, line 38: Should this simply be `inference/marian-fork` instead? (I wonder if you're doing this to workaround an upstream bug, similar to what was recently fixed in https://github.com/taskcluster/taskgraph/pull/723?)

bhearsum (approved):
Looks fine from a taskgraph standpoint.

pipeline/quantize/extract_stats.py
evgenyrp, line 1: Maybe it's worth documenting what this file does and that it was copied from browsermt Marian

evgenyrp (approved):

taskcluster/kinds/compile/kind.yml
gregtatum: This is because of https://github.com/taskcluster/taskgraph/issues/748

I'll add a note.

Issue #1197, opened 2025-07-22
Investigate using AsyncService for the Wasm inference engine


gregtatum:
I've had concerns over memory usage when doing parallelized translations. I did an investigation in #957 that looked into memory usage using Valgrind's dhat tool. Here is the breakdown of memory usage across different components.

| Tiny mode                    |    MB |
| ---------------------------- | -----:|
| Model Size                   |  17.1 |
| Model Size Unquantized       |  41.7 |
| Total memory allocated       | 201.1 |
| Memory at peak usage         | happens during model loading before clears: 111.8 |
| Memory of initialized engine |  58.3 |
| Model stored as Tensors      |  41.7 |
| Model beam search values     |  10.5 |
| SentencePiece                |   4.4 |
| Vocab Memory                 |  We should clear this: 0.8 |

https://share.firefox.dev/4kTnpPX (no workspace)
https://share.firefox.dev/3GX84Qy

This is using the [BlockingService](https://github.com/mozilla/translations/blob/2217a7db1382e1bf20b5146591d5668cbf210352/inference/src/translator/service.h#L30) which is single threaded. There is also an [AsyncService](https://github.com/mozilla/translations/blob/2217a7db1382e1bf20b5146591d5668cbf210352/inference/src/translator/service.h#L108).

Looking into the code, it looks like the 41.7 MB "Model stored as Tensors" could be immutable memory once initialized. If this is the case, then most of the memory state would be the 10.5 MB of "Model beam search values". This would mean that for every thread it would be an additional 10.5 MB of memory. This feels pretty acceptable to me, and my intuition is that it won't explode in size with a `base-memory` model.

We should also investigate shortlisting impliciations, which is currently not being used. The implementation has some pretty bad and naive memory management that would need to be fixed. I wouldn't consider shortlisting a blocker for parallelization, as there is bigger work needed to turn shortlisting back on.

Once I get #1195 landed I'll submit a PR for my new build cli that uses the BlockingService.

PR #1198, opened 2025-07-22
Build the inference engine locally using the same codepaths as the Wasm engine.

gregtatum:
This contains work from some other PRs, so I'm marking it as draft until I get the others landed. This is a prerequisite for #1197.


gregtatum:
I should probably write some docs as well.

PR #1209, opened 2025-08-06
Trigger final evals as an action

gregtatum:
This combines the work of triggering an evaluation action and an initial prototype of an LLM evaluation.

URL that can be used in the evaluate site page:
https://firefoxci.taskcluster-artifacts.net/QRG8BTbIRJyA96SS7LIgWg/0/public/build/

Evaluate action: https://firefox-ci-tc.services.mozilla.com/tasks/CO1LewhBR_yjGg2FhxGRcA

Successful run (with truncated numbers to be fast): https://firefox-ci-tc.services.mozilla.com/tasks/groups/CO1LewhBR_yjGg2FhxGRcA

Example uploads:

```
➤ gsutil ls gs://moz-fx-translations-data--5f91-stage-translations-data/models/bs-en/spring-2024_CO1LewhBR_yjGg2FhxGRcA/final_eval/mozilla_tiny/final-eval-llm
gs://moz-fx-translations-data--5f91-stage-translations-data/models/bs-en/spring-2024_CO1LewhBR_yjGg2FhxGRcA/final_eval/mozilla_tiny/final-eval-llm/scores.json
gs://moz-fx-translations-data--5f91-stage-translations-data/models/bs-en/spring-2024_CO1LewhBR_yjGg2FhxGRcA/final_eval/mozilla_tiny/final-eval-llm/summary.json
```


gregtatum:
I triggered en-ru for tiny, base, base-memory

https://firefox-ci-tc.services.mozilla.com/tasks/JVTEz6rDSF6CPT5iqbWb3w
https://firefox-ci-tc.services.mozilla.com/tasks/UZoczXzvQ7eYMTcpWpHrow
https://firefox-ci-tc.services.mozilla.com/tasks/VrAnrNjUTZerUBlJWXFTxw

gregtatum:
Here's the dashboard live on my fork.

* [en-ru tiny](https://gregtatum.github.io/translations/llm-evals/?baseUrl=https%3A%2F%2Ffirefoxci.taskcluster-artifacts.net%2FTcGZ1Tx2QKu-Q8I8K5SU5A%2F0%2Fpublic%2Fbuild%2F)
* [en-ru base-memory](https://gregtatum.github.io/translations/llm-evals/?baseUrl=https%3A%2F%2Ffirefoxci.taskcluster-artifacts.net%2Fe_b5A3fbTXulufwjmcaY2w%2F0%2Fpublic%2Fbuild%2F)
* [en-ru base](https://gregtatum.github.io/translations/llm-evals/?baseUrl=https%3A%2F%2Ffirefoxci.taskcluster-artifacts.net%2FYIQtvNv2SK-nr73sMdg2qg%2F0%2Fpublic%2Fbuild%2F)

bhearsum (approved):

pipeline/eval/eval-batch-instructions.md
```diff
@@ -0,0 +1,111 @@
+# Identify
+
+You are a qualitative evaluator for a machine translation system ({src}-{trg}).
+
+# Input
+
+You will receive an array of translation examples. Each example is an object with:
+- src: source sentence
+- trg: model translation
+- ref: human reference translation
```
evgenyrp, line 10: We can also do referenceless evals with LLMs. The benefit is that we can run it on any text. Also, I think at this point, LLMs might even evaluate better without a reference because a human-provided reference can also be imperfect.

pipeline/eval/llm.py
```diff
@@ -0,0 +1,381 @@
...
+        if attempt == 0:
+            # Start with a stable temperature to have consistent results.
+            temperature = 0.0
+            logger.info(
+                f"Querying {config.model} with batch {batch_index} containing {len(translations)} translations."
+            )
+        else:
+            # Increase the temperature so that the results will be varied on the second
+            # attempt.
+            temperature = 0.5
```
evgenyrp, line 179: Interesting way to get rid of hallucinations

tests/test_final_eval.py
```diff
@@ -0,0 +1,29 @@
+from fixtures import DataDir
+from pathlib import Path
+import os
+import pytest
+
+ROOT_DIR = Path(__file__).parent.parent
+
+
+def test_final_eval():
+    if not os.environ.get("OPENAI_API_KEY"):
```
evgenyrp, line 10: How hard would it be to mock the OpenAI call right away to make this test functional? This would help prevent regressions in the future in case we don't have time to follow up on this.

pipeline/eval/llm.py
```diff
@@ -0,0 +1,381 @@
...
+
+def run_eval_batch_prompt(
+    client: OpenAI,
+    translations: list[dict[str, str]],
+    config: Config,
+    instructions: str,
+    batch_index: int,
+):
+    start = time.time()
+    input = translations_batch_to_text(translations)
```
evgenyrp, line 159: Did you test batching vs using a single example per call? I suspect batching is cheaper, but the quality might degrade if there's too much text. On the other hand, the scores will likely be more consistent.

taskcluster/configs/eval.yml
```diff
@@ -0,0 +1,27 @@
...
+  src: bs
+  trg: en
+  # The artifacts will be saved with the following format:
+  # gs://{upload-bucket}/models/{src}-{trg}/{model_name}_<decision>/
+  #                      final_eval/{model_service}_{model_architecture}/{task_name}/
+  service: mozilla
+  architecture: tiny
+  name: spring-2024
+corpus:
+  src: https://raw.githubusercontent.com/mozilla/firefox-translations-models/refs/heads/main/evaluation/bs-en/flores-dev.bs
```
evgenyrp, line 11: It's nice to be able to provide a custom test set, however, we could also add an option to provide just a dataset name. It's the only piece that's missing in the config to form the URL.

taskcluster/kinds/upload-final-eval/kind.yml
```diff
@@ -0,0 +1,99 @@
...
+                # the task id of FOO, which is not known until very late in
+                # the taskgraph generation process
+                # https://firefox-source-docs.mozilla.org/taskcluster/taskgraph.html#task-parameterization
+                "*.log":
+                    - {
+                        "task-reference": "logs/{src}-{trg}/{model_name}_<decision>/final_eval/{model_service}_{model_architecture}/{name}/"
+                      }
+                "*":
+                    - {
+                        "task-reference": "models/{src}-{trg}/{model_name}_<decision>/final_eval/{model_service}_{model_architecture}/{name}/"
```
evgenyrp, line 98: First observation: we don't have to use the same folder where we store models. I'd rather use a separate folder: `evals` as it's completely separate from our training uploading.

Also, I wonder if we can come up with something more discoverable. I assume `model_name` can be a version from the models repo or something else. Is `decision` an ID of the decision task? I think if we want to run evals again for the same model, it would be easier to use a timestamp so that we can later sort and take the latest results. 

How about this format: `<roo_dir>/<lang-pair>/<dataset_name>/<service>/<architecture_or_api>/<version>/<timestamp>/<evals_files>`

For example: 
`evals/en-zh/flores-test/mozilla/tiny/2.1/20250813112233/*.json`
We can also add COMET/BLEU jsons there later.


`2.1` can be a version from the models repo or any arbitrary model name if we want to evaluate some dev models.
`flores-test` can be a dataset name from the model repo, or again any other name if we want to use some external data

`evals/en-zh/flores-test/google/api/9.8.2/20250813112233/*.json`
`evals/en-zh/news123/mozilla/tiny/2.1/20250813112233/*.json`
`evals/en-zh/flores-test/mozilla/tiny/dev-retraining-experiment/20250813112233/*.json`

If we want to run final evals for the newly trained model from the training pipeline we could replace model version with a task ID like 
`evals/en-zh/flores-test/mozilla/tiny/R0kZGTH6S7Kz11wvksCrEg/20250813112233/*.json`

Then we would be able to easily list the available artifacts on any level and see what's there: `gsutil ls evals/en-zh/`

evgenyrp (commented):
@gregtatum great job! I enjoyed looking at the UI and inspecting specific examples in each bin of the histogram. It makes it a lot clearer what the model struggles with. We can later extend it with some specific categories we care about, like numbers, URLs etc.

I have some suggestions  and questions that we should discuss. Most of them are minor except the GCS structure, we should settle on something that will be easy to extend and use in the next iterations.

pipeline/eval/eval-batch-instructions.md
gregtatum: Yeah I wasn't convinced the reference was ideal on some of them. That sounds like a good idea to me. I filed it: https://github.com/mozilla/translations/issues/1219

tests/test_final_eval.py
gregtatum: It's a bit harder using run_task since we can't directly modify the test with some kind of pytest mock. I filed: https://github.com/mozilla/translations/issues/1220

pipeline/eval/llm.py
gregtatum: I did not take the time to test it. I think once it lands it'll be easier to rewrite the prompt and test the results. My intuition is that it's a trade-off between the LLM not being able to digest all of the content it's provided, vs it hallucinating bad things to report, just to report _something_. Seems worth testing in the longer term.

gregtatum: I guess you could do it with a batch size of 1, but I think it would benefit from rewriting the prompt to be much simpler.

gregtatum: https://github.com/mozilla/translations/issues/1221

taskcluster/configs/eval.yml
gregtatum: Yeah I agree, that would be a follow-up. This would need the model as well as the dataset name. This is more inline with what our models repo does.

gregtatum: https://github.com/mozilla/translations/issues/1222

taskcluster/kinds/upload-final-eval/kind.yml
gregtatum: Let me think about this, and we can discuss in chat to align on a naming strategy.

evgenyrp: It's ok to follow up on that

evgenyrp (approved):

Issue #215, opened 2023-10-04
Improve translation of short sentences


eu9ene:
We have issues with translating shorter sentences and single words. 

What might help:
- Better data cleaning to avoid removing short translations from the training datasets
- Try to find more datasets that contain short sentences

gregtatum:
Or maybe even synthesize it with the alignment data.

gregtatum:
This behavior is also visible with numbers. A good example is to do a list of numbers.

gregtatum:
Verify the fix with:

https://bugzilla.mozilla.org/show_bug.cgi?id=1888972

gregtatum:
Here is a word count distribution for the merged corpus sl-en: https://firefox-ci-tc.services.mozilla.com/tasks/groups/PPCzZRHaTT6Ys4BIhPGT5w

![word count distribution "en"](https://github.com/user-attachments/assets/12c42a05-5fdc-4963-9b36-c887984edeae)

![word count distribution "sl"](https://github.com/user-attachments/assets/d0eb3c3b-daf3-4ed5-8e67-fd1052a06a9e)

Generated via:

```
python3 pipeline/data/analyze.py --file_location https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/VK5zmxJRTLy0y0WBQ0DRJg/artifacts/public/build/corpus.en.zst --output data --dataset OPUS_corpus --language en

python3 pipeline/data/analyze.py --file_location https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/VK5zmxJRTLy0y0WBQ0DRJg/artifacts/public/build/corpus.sl.zst --output data --dataset OPUS_corpus --language sl
```

marco-c:
So we have basically 0 sentences with 1 word?

gregtatum:
I filed #878 which suggests augmenting with statistically synthesized single word translations.

gregtatum:
I filed #879 which suggests harvesting short sentences from parallel datasets.

ZJaume:
Here are some stats extracted form the merged corpus for en-ar.

![v1 ar distribution-words](https://github.com/user-attachments/assets/25ddc705-4c47-4309-947b-861a62e398ea)

![v1 en distribution-words](https://github.com/user-attachments/assets/a432627b-286d-4f2b-8ce6-734de0ae2db0)

eu9ene:
[...]

Issue #514, opened 2024-04-04
Create a dataset of short sentences for training and/or validation


marco-c:
See also #215.

PR #1215, opened 2025-08-12
Evaluate short sentences with unaligned ratio metric

ZJaume:


ZJaume:
Well I think this should be ready to merge. As we talked on Wednesday, the aligner is easily failing on any CI model because model outputs are gibberish. But I managed to make it work more or less, skipping empty sentences and returning empty alignment pairs when the aligner crashes because all the tokens in a sentences are our of vocabulary. In this case the metric will count as 1.0 (all tokens unaligned). For the CI we don't care if the metrics are crazy.

I have been able to [run the metric on real models](https://firefox-ci-tc.services.mozilla.com/tasks/groups/QzHfEpt8TSO1NeeHWxcFyA) and the unaligned ratio metric is very low as expected. However, the evaluation on short sentences has not been [able to run](https://firefox-ci-tc.services.mozilla.com/tasks/IQ7C13yYSh-N5x1SIxT1MA) because of #1234. But the dataset task runs on the CI, so I guess we can merge anyway?

pipeline/data/parallel_importer.py
```diff
@@ -25,6 +25,7 @@
 from opustrainer.modifiers.typos import TypoModifier
 from opustrainer.types import Modifier
 
+from pipeline.alignments.tokenizer import IcuTokenizer
```
ZJaume, line 28: The dependencies from here have been included in eval.txt, but I'm not sure if it's the way to do it or the kind.yml should install both alignments and eval dependencies.

pipeline/eval/eval.py
```diff
@@ -267,7 +344,11 @@ def main(args_list: Optional[list[str]] = None) -> None:
 
     if os.environ.get("COMET_SKIP"):
         comet_score = "skipped"
+        unaligned_ratio_ref = "skipped"
```
ZJaume, line 329: I put unaligned ratio together with comet in the skip section because both are model-based metrics.

utils/config_generator.py
```diff
@@ -368,6 +405,10 @@ def add_test_data(
                 devtest_datasets.append(f"sacrebleu_aug-mix_{dataset_name}")
             is_test = not is_test
 
+    print("Fetching pontoon")
```
ZJaume, line 372: Short sentences will be evaluated always

ZJaume:
This might close #215 and #514 ?

pipeline/data/parallel_importer.py
```diff
@@ -51,6 +52,44 @@ def __call__(self, batch: List[str]) -> Iterable[str]:
...
+        n = 0
+        for tok in tokenizer.tokenize(text):
+            if tok != tokenizer.SPACE_TOKEN:
+                n += 1
+        return n
+
+    def _filter_minmax_words(self, corpus: List[str]) -> Iterable[str]:
+        for line in corpus:
+            src_seg, trg_seg = line.split("\t")
+            # TODO use tokenizer
```
evgenyrp: you're already using it

pipeline/data/parallel_importer.py
```diff
@@ -51,6 +52,44 @@ def __call__(self, batch: List[str]) -> Iterable[str]:
...
+
+    def __init__(self, src_lang: str, trg_lang: str, n: int, min_words: int, max_words: int):
+        self.trg_tokenizer = IcuTokenizer(trg_lang)
+        self.src_tokenizer = IcuTokenizer(src_lang)
+        self.n = n
+        self.min_words = min_words
+        self.max_words = max_words
+
+    def __call__(self, corpus: List[str]) -> Iterable[str]:
+        return random.sample(list(self._filter_minmax_words(corpus)), self.n)
```
evgenyrp: I guess we won't deal with huge datasets for evals, so this should be fine

pipeline/data/parallel_importer.py
```diff
@@ -156,6 +197,14 @@ def augment(output_prefix: str, aug_modifier: str, src: str, trg: str):
 
     corpus = read_corpus_tsv(compressed_src, compressed_trg, uncompressed_src, uncompressed_trg)
 
+    # Special modifier aug-short just samples and finishes
+    if aug_modifier == "aug-short":
+        # Sample params are hardcoded for now
+        modifier = SampleModifier(src_lang=src, trg_lang=trg, n=400, min_words=1, max_words=2)
```
evgenyrp, line 202: It's ok to hardcode but I think our typical evals dataset size is ~1k, so maybe we can use it here as well?

evgenyrp, line 202: Since the segments are so short, 1-2 words, we should do deduplication. So in this case it would require sampling more just in case, maybe 5k, then dedupe and sample 1k.

evgenyrp: I don't know if it's useful here but we also have `datasets.sample_corpus`. It's a bit more focused on streaming to work with huge datasets like HPLT. Maybe it's worth doing the first pass using it in case we want to sample from bigger datasets in the future.

pipeline/eval/eval.py
```diff
@@ -72,6 +74,81 @@
     WANDB_AVAILABLE = False
 
 
+def tokenize_nospace(sentence, tokenizer):
```
evgenyrp: I suggest adding typing where we can

pipeline/eval/eval.py
```diff
@@ -72,6 +74,81 @@
...
+    """
+    Tokenize a sentence and return a list of tokens that have no space
+    """
+
+    def iter_tok():
+        for t in tokenizer.tokenize(sentence):
+            if t == tokenizer.SPACE_TOKEN:
```
evgenyrp: Ideally the ability to tokenize without spaces should be a property of `IcuTokenizer`

pipeline/eval/eval.py
```diff
@@ -72,6 +74,81 @@
...
+def compute_unaliged_ratio(src: str, trg: str, source_lines: List[str], target_lines: List[str]):
+    """
+    Compute ratio of unaligned tokens
+    """
+    aligner = SentenceAligner(model="bert", token_type="bpe", matching_methods="mai")
+    src_tokenizer = IcuTokenizer(src)
+    trg_tokenizer = IcuTokenizer(trg)
+    src_tokens = [tokenize_nospace(i, src_tokenizer) for i in source_lines]
+    trg_tokens = [tokenize_nospace(i, trg_tokenizer) for i in target_lines]
+    src_tokens, trg_tokens = zip(*filter_empty(src_tokens, trg_tokens))
```
evgenyrp: Why would they be empty? Is it for CI only?

pipeline/eval/eval.py
```diff
@@ -72,6 +74,81 @@
...
+                # in the sentencealigner vocab, it cannot align because there are no embeddings
+                # so, in this case we just return an empty alignment pairs
+                # which means nothing could be aligned
+                if e.args and e.args[0].startswith("Found array with 0 sample"):
+                    yield {"itermax": []}
+                else:
+                    raise e
+            except Exception as e:
+                # If it fails print sentence pair for easier debugging
+                logger.error(f"Source: {st}")
```
evgenyrp, line 114: Should we add an error message before printing to provide a context on what those errors are?

pipeline/eval/eval.py
```diff
@@ -72,6 +74,81 @@
...
+    def gen_alignments():
+        for st, tt in zip(src_tokens, trg_tokens):
+            try:
+                yield aligner.get_word_aligns(st, tt)
+            except ValueError as e:
+                # When there is a sentences containing only characters that are not present
+                # in the sentencealigner vocab, it cannot align because there are no embeddings
+                # so, in this case we just return an empty alignment pairs
+                # which means nothing could be aligned
+                if e.args and e.args[0].startswith("Found array with 0 sample"):
```
evgenyrp, line 107: Does it happen only in CI? If yes, let's add a comment about it.

taskcluster/configs/config.ci.yml
```diff
@@ -96,6 +96,7 @@ datasets:
     - sacrebleu_aug-upper_wmt19
   test:
     - flores_devtest
+    - tmx_aug-short_pontoon
```
evgenyrp, line 99: I hope it's not too slow for CI

utils/config_generator.py
```diff
@@ -77,6 +77,43 @@
     "tr", "uk", "umb", "ur", "uz", "vi", "wo", "xh", "yo", "zh", "zh", "zu"
 }  # fmt: skip
 
+pontoon_languages = { "aa", "aat", "ab", "abb", "abq", "ace", "ach", "ady", "af", "ajg", "ak",
```
evgenyrp: Since we're hardcoding pontoon lang codes in two places, it's worth moving those to a separate file, maybe somewhere in `pipeline/data/pontoon.py`.

pipeline/data/parallel_importer.py
evgenyrp: The kind for evals should install only evals dependencies

pipeline/eval/eval.py
evgenyrp: Up to you

evgenyrp (changes requested):
Overall looks great! I have a bunch of minor suggestions and one more meaningful about deduplication.

bhearsum (approved):
Looks fine from an automation standpoint.

taskcluster/configs/config.ci.yml
ZJaume: I don't know how much time was the CI doing before, but now it is not being slow, I think. The metric is a bit faster than computing COMET scores and the dataset downloading does not take much more time than others.

ZJaume: I don't know what I was looking at. Actually evaluations are much slower now. For flores it takes 1h. Let's see if I can make one pontoon evaluation successful to se how much it takes.

pipeline/eval/eval.py
ZJaume: done

taskcluster/configs/config.ci.yml
ZJaume: It was not using gpu. Now it is using it and is much faster but still takes more time than computing COMET. For example for [evaluate-quantized-flores-devtest-ru-en](https://github.com/mozilla/translations/pull/1215/checks?check_run_id=50149258852) it takes 28s to compute COMET and almost 9m for the alignments. The total time of the task is 19m, compared to the same task in #1243, which took [8m](https://github.com/mozilla/translations/pull/1243/checks?check_run_id=50087045921) (I guess the task there also got different hardware because COMET took 16s).

The evaluation task for the `pontoon-aug_short` dataset is taking 8m. Which is more acceptable IMO. So I think this conversation overlaps with the one above requesting 1k size for this dataset if we want to keep CI times low.

Taking into account that this feature has taken a lot of time to develop and is meant to monitor an issue that might already be solved, I think it does not make much sense for us to be exhaustive or to try to invest more time in it finding optimizations. We could just compute it for pontoon dataset and also keep the size at 400 sentences. Both for CI and real evaluations. What do you think?

utils/config_generator.py
ZJaume: done

pipeline/data/parallel_importer.py
ZJaume: I considered that when I started but ended discarding the idea because I wanted to do the minword filtering before sampling. The proportion of short sentences in the corpus was ranging a lot depending on the language. Sampling before, could end up with too few short sentences.

Regarding the dedup, I just applied a `set` after filtering and before sampling.

taskcluster/configs/config.ci.yml
evgenyrp: The dataset task, which also adds alignments for 1000 lines flores takes 6m: https://firefox-ci-tc.services.mozilla.com/tasks/BjKp-oXGQsOVprA8Gv0DyQ/runs/0/logs/public/logs/live.log. If it's in the same ballpark, we're good. Probably the dataset downloading might take longer.

I think not everything has to be included in CI. 8m GPU task is fine. If it takes longer like 20m, it's probably better to skip it in CI as it should be fast and cheap. 

Agreed on optimizations overall. 1000k was more about making the sample more statistically representative to follow other evals. We don't have to optimize it. If it's slow, we can remove it from CI.

pipeline/eval/eval.py
```diff
@@ -72,6 +75,64 @@
...
 
 
+def compute_unaliged_ratio(
+    src: str, trg: str, source_lines: List[str], target_lines: List[str], device: torch.device
+):
+    """
+    Compute ratio of unaligned tokens
+    """
+    aligner = SentenceAligner(
+        model="bert", token_type="bpe", matching_methods="mai", device=device
```
evgenyrp: "i" should be enough if you use only itermax as I do in the dataset task. I don't know how much overhead using other methods adds, maybe not a lot, but still.

taskcluster/configs/config.ci.yml
evgenyrp: @ZJaume do we understand why adding alignments in the dataset tasks that run on a CPU with the same method takes 1 minute? https://firefox-ci-tc.services.mozilla.com/tasks/cTMeVhoyTHuQYWTglcttJg/runs/0/logs/public/logs/live.log

pipeline/data/parallel_importer.py
```diff
@@ -51,6 +52,43 @@ def __call__(self, batch: List[str]) -> Iterable[str]:
...
+        self.min_words = min_words
+        self.max_words = max_words
+
+    def __call__(self, corpus: List[str]) -> Iterable[str]:
+        return random.sample(list(set(self._filter_minmax_words(corpus))), self.n)
+
+    @classmethod
+    def _count_words(cls, text: str, tokenizer: IcuTokenizer):
+        n = 0
+        for tok in tokenizer.tokenize(text):
```
evgenyrp, line 73: or we can just do `tokenize_nospace`

taskcluster/configs/config.ci.yml
evgenyrp: I think we can remove `tmx_aug-short_pontoon` from CI, but since the metric will be calculated for all datasets, as intended by this design, any overhead will be multiplied by 1000s of runs. So I'd make sure it works efficiently enough and there are no bugs at least.

evgenyrp (commented):
Looks good, but I have a few comments left. Let's double-check that we do everything right and that the evals tasks don't take too long now.

evgenyrp:
Some CI tasks got stuck due to #1247, the fix is on the way, see #1250. I cancelled those.

pipeline/eval/eval.py
ZJaume: good catch, let's see

PR #1217, opened 2025-08-13
New vocab `__sep__` symbol for DocMT and auxiliary symbols

ZJaume:


evgenyrp (approved):
@ZJaume Let's add those symbols for the future experiments, it won't hurt. Merging the configs is also good.

pipeline/train/spm-vocab.sh
```diff
@@ -65,6 +65,12 @@ fi
 
 vocab_dir=$(dirname "${vocab_src_output}")
 mkdir -p "${vocab_dir}"
+user_symbols="
+__source__,__target__,__done__,__start__,__end__,__sep__,
+__misc0__,__misc1__,__misc2__,__misc3__,__misc4__,
+__misc5__,__misc6__,__misc7__,__misc8__,__misc9__,
+"
```
evgenyrp, line 72: @ZJaume, one question here: did you add \n on purpose? Are there any negative implications for the current models from this? I assume the models will never encounter it in our current data.

ZJaume: This kind of multiline string in bash does not preserve endlines, it is just for code formatting reasons.

PR #1228, opened 2025-08-19
Corpora specific fixes

### LINKED ISSUE #910: More corpora specific fixes
[Here](https://github.com/ZJaume/clean/tree/master/fixes) there is a compilaton of fixes I did in the past that could be incorporated to the OpusCleaner configs.

evgenyrp:
- Delete legacy scripts
- Add OpusCleaner dataset specific configs
- Add [fixes](https://github.com/ZJaume/clean/tree/master/fixes) for Portuguese ( I also looked at en-fr for those datasets in OpusCleaner and applied them to all languages if I saw the same problems)

Fixes #910

### DIFF
--- a/pipeline/clean/merge-parallel.py
+++ b/pipeline/clean/merge-parallel.py
@@ -193,9 +193,9 @@ def get_datasets(src: str, trg: str, datasets_glob: str):
     for dataset in dataset_paths:
         path = Path(dataset)
         countbytes = True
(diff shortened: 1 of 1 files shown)


ZJaume (approved):
The cascade system seems a good idea to me, copying filters around when a corpus or language needs a specific fix is tedious and risky.

Issue #1247, opened 2025-09-11
Tasks get stuck on git clone


evgenyrp:
For example,

https://firefox-ci-tc.services.mozilla.com/tasks/FuOSjZd4RFOrRz1YoJQCrQ/runs/0
https://firefox-ci-tc.services.mozilla.com/tasks/I8Eq6frdQAa_vEwuCKSI1w/runs/0/logs/live/public/logs/live.log

@bhearsum is it an underlying Taskcluster issue?

bhearsum:
OK; installing `git-lfs` repros the problem easily over https, but not over ssh. (The latter is not an option for us on these workers at this time, though.) And I have this in the process tree:
```
bhearsum    9182  0.3  0.0 119460 24336 pts/0    S+   15:12   0:00  \_ git clone https://github.com/mozilla/firefox-translations-training htueno
bhearsum    9213  0.0  0.0   2892  1644 pts/0    S+   15:12   0:00      \_ /bin/sh -c git-lfs filter-process git-lfs filter-process
bhearsum    9214  0.0  0.0 2273672 14624 pts/0   Sl+  15:12   0:00          \_ git-lfs filter-process
bhearsum    9245  0.0  0.0  16876  3816 pts/0    S+   15:12   0:00              \_ git credential fill
```

In the immediate term, removing the LFS files from the repo and/or removing `git-lfs` from the docker image would be a workaround. It's unclear what has changed here. I suspect that github itself has changed, but can't find any evidence to support that at this time, so I'm not sure if that's actually true, and if it is, whether it's a bug or not.

PR #1250, opened 2025-09-16
Remove git lfs

### LINKED ISSUE #1249: Switch from Git LFS and local models to GCS
Git LFS is causing some issues (see #1247 ) and also makes it harder to start using the repo.

The only objects we store on it are `inference/wasm/tests/models` that we use for tests. We can download them from a GCP bucket on demand (`gs://releng-translations-dev`). Then remove Git LFS from the config and Docker images.

evgenyrp:
Move models to GCS

Fixes #1249 #1247

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,0 @@
-inference/**/*.gz filter=lfs diff=lfs merge=lfs -text
--- a/inference/scripts/test-wasm.py
+++ b/inference/scripts/test-wasm.py
@@ -5,6 +5,7 @@ import os
 import shutil
 import subprocess
 import sys
+from urllib.request import urlretrieve
 
 SCRIPTS_PATH = os.path.realpath(os.path.dirname(__file__))
 INFERENCE_PATH = os.path.dirname(SCRIPTS_PATH)
@@ -13,6 +14,9 @@ WASM_PATH = os.path.join(INFERENCE_PATH, "wasm")
 WASM_TESTS_PATH = os.path.join(WASM_PATH, "tests")
 GENERATED_PATH = os.path.join(WASM_TESTS_PATH, "generated")
 MODELS_PATH = os.path.join(WASM_TESTS_PATH, "models")
+MODELS_GCS_PATH = os.path.join(
+    "https://storage.googleapis.com/releng-translations-dev/tests/models"
+)
 WASM_ARTIFACT = os.path.join(BUILD_PATH, "bergamot-translator.wasm")
 JS_ARTIFACT = os.path.join(BUILD_PATH, "bergamot-translator.js")
 JS_ARTIFACT_HASH = os.path.join(GENERATED_PATH, "bergamot-translator.js.sha256")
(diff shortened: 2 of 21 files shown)


evgenyrp:
> For what it's worth, this was eventually an [acknowledged issue](https://www.githubstatus.com/incidents/nh22ywy9dh84) and [tasks now succeed](https://firefox-ci-tc.services.mozilla.com/tasks/T6OwoEEDQlKFpVotmZCCYw).-
> 
> I have no opinion about taking this patch or leaving things as is.

Great! I'd prefer to still get rid of git lfs as we don't really need it in this repo.

nordzilla:
I'm a bit confused here. 

At the time I implemented this, we didn't have an `en-zh` model, so for the `en-zh` tests I used a Spanish model under the hood. (The important thing to test was that spaces were omitted between sentences when reconstructing the target text).

Example:

* https://github.com/mozilla/translations/blob/dee8546cbb15bf1f3fbf6152e11bb21cdfee673a/inference/wasm/tests/test-cases/translate-html-multi-sentence-no-pivot.test.mjs#L57-L67

I'm not sure how those tests are still passing, because in theory it should be downloading the _actual_ `en-zh` model now, which wouldn't match the test case expectation.

evgenyrp:
> > I'm a bit confused here.
> > At the time I implemented this, we didn't have an `en-zh` model, so for the `en-zh` tests I used a Spanish model under the hood.
> > (The important thing to test was that spaces were omitted between sentences when reconstructing the target text.)
> > Example:
> > ```
> > * https://github.com/mozilla/translations/blob/dee8546cbb15bf1f3fbf6152e11bb21cdfee673a/inference/wasm/tests/test-cases/translate-html-multi-sentence-no-pivot.test.mjs#L57-L67
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I'm not sure how those tests are still passing, because in theory it should be downloading the _actual_ `en-zh` model now, which wouldn't match the test case expectation.
> 
> @evgenyrp
> 
> I just built and tested it locally and it's passing. How is it possible if it's downloading our real `enzh` model from GCS. It shouldn't be getting Spanish output anymore.
> 
> EDIT:
> 
> Oh, or perhaps you specifically uploaded the models that were in the tests directory to a GCS directory. That would make sense why it's matching up with the Spanish.
> 
> I'm happy to land this and file a follow-up to replace the `enzh` model in that GCS location with a real one, then also update the test expectations. I can take that work if you want.
> 
> EDIT 2:
> 
> I can't seem to find `releng-translations-dev/tests/models` in GCS. I see `translations-data-prod` and `translations-data-nonprod`, but neither of those seem to be it.

I just copied the directory to our dev bucket `releng-translations-dev`, so yes, if we want to replace it with a proper model, we can do it easily later.

Not sure what the non-prod bucket is. Maybe someone wanted to rename our dev bucket but didn't finish it. I wouldn't do this as we have links there from various places:
```
gsutil ls gs://translations-data-nonprod
BucketNotFoundException: 404 gs://translations-data-nonprod bucket does not exist.
$ gsutil ls gs://releng-translations-dev
gs://releng-translations-dev/test
gs://releng-translations-dev/data/
gs://releng-translations-dev/experiments/
gs://releng-translations-dev/logs/
gs://releng-translations-dev/models/
gs://releng-translations-dev/tests/
```

nordzilla (approved):

PR #1281, opened 2025-10-29
ELRC shutdown fix

ZJaume:
Rely only on OPUS for ELRC datasets and exclude the ones that are test sets.


nordzilla:
Changes look reasonable to me, but I'm requesting review from Evgeny explicitly, since he's much more familiar with the training data.

ZJaume:
I was trying to trigger a CI training test but it didn't work, idk why.

Issue #1261, opened 2025-09-22
Extend the model registry dashboard to show all evals


evgenyrp:
Currently, it shows only flores for the pipeline evals. It should be able to show the evals for all the datasets and metrics. https://mozilla.github.io/translations/model-registry/?searchString=&showModels=true&score=comet

Most importantly, it should show the final evals that we'll run as a part of the Evaluate Taskcluster action.
It should show all this info and make it as convenient to compare the results as we currently have here: https://github.com/mozilla/firefox-translations-models/blob/main/evaluation/comet-results.md

evgenyrp:
It will be clear what the data on GCS looks like when we get to #1259

Issue #1219, opened 2025-08-14
Support reference-less LLM evals


gregtatum:
In #1209 I'm adding LLM evaluation of our translations with explanations. We can also make it work for translations without a reference.

> @evgenyrp stated: We can also do referenceless evals with LLMs. The benefit is that we can run it on any text. Also, I think at this point, LLMs might even evaluate better without a reference because a human-provided reference can also be imperfect.

In this case you would make the `ref` property optional in the eval config, and then provide a separate prompt for this request, or update the prompt to allow for a null reference. The rest of the code should probably just work.

Issue #889, opened 2024-10-21
Consider adding NTREX-128 for evaluation


ZJaume:
Languages that do not have WMT newstest can use [NTREX](https://github.com/MicrosoftTranslator/NTREX), which is 2019 news test translated to 128 languages. This is a harder test set than Flores, which is fine to know if a model has decent quality, but it is a relatively easy test set.

evgenyrp:
Related: https://github.com/mozilla/translations/issues/1308

PR #1284, opened 2025-11-05
Migration of final evaluation

### LINKED ISSUE #764: [meta] Improve automatic quality evaluation
This is a meta issue to brainstorm ideas on how to make our final automatic quality evaluation procedure more robust and suitable for decision-making. This currently happens in [firefox-translations-models](https://github.com/mozilla/firefox-translations-models) repo but will be migrated to the main repo and W&B.

See [this doc](https://docs.google.com/document/d/17DfIV0jeHlkU6j85JGwXCTQIGfuo6S8KJXAcJvBCqNI/edit?usp=sharing).

Some ideas:
- Consider reporting absolute difference instead of %
- Consider adding more metrics recommended in the literature
- Consider adding more interpretable metrics where absolute values indicate readiness for release, for example, xCOMET
- Try not to use validation datasets where possible (remove flores-dev)
- Inspect how representative the datasets are compared to what users translate in the browser
- Use the same datasets for all languages as a final metric ( either only Flores or build a new dataset)

Recommendations from the paper [Navigating the Metrics Maze: Reconciling Score Magnitudes and Accuracies](https://arxiv.org/pdf/2401.06760):
- Use CometKiwiQE22 as the main metric. In addition to its better performance, as a quality estimation metric, it is not affected by references.
- Use at least one additional metric of a different type; e. g. BLEURT20, which is reference-based and uses a different architecture from Comet.
- For each metric delta, report estimated accuracy to help align reliability of used metrics.
- Do not use BLEU, ChrF, or spBLEU to evaluate unrelated systems.

We can add a task list here after migrating [firefox-translations-models].

### LINKED ISSUE #1222: Run inference from a model and a dataset name for the LLM evaluator.
Right now in the LLM evaluator it only takes URLs to the src, trg, and ref translations. It should be updated to handle downloading a model, and a dataset "by key", performing the inference, and then running the evaluation on that dataset. In the main pipeline we do this with separate tasks to help with scale and parallelization. Here we'll want to do that in a "fat" task, that handles all of the things, as we're not operating on huge datasets. Plus the student models work on cheaper CPUs instances.

The eval config will need to be updated to take this alternate style config. It would be nice to retain the arbitrary dataset support when doing this.

### LINKED ISSUE #1231: Add BOUQuET evaluations
BOUQuET is a new more diverse dataset of evaluations. We've been relying on flores which is a good evaluation set, but is limited in the domain. BOUQuET has been designed to not be English-centric, and it's designed to have a wide level of domains.

https://huggingface.co/datasets/facebook/bouquet
https://arxiv.org/pdf/2502.04314v2

### LINKED ISSUE #1259: Migrate final evals for bergamot and external models
Migrate the code from https://github.com/mozilla/firefox-translations-models/tree/main/evals here and reuse the pipeline evals code. They should run as a part of the final eval action.

Migrate the evals data from https://github.com/mozilla/firefox-translations-models/tree/main/evaluation to GCS

### LINKED ISSUE #1262: Fix LLM evals storage path
It should be in line with other evals. See this comment: https://github.com/mozilla/translations/pull/1209#discussion_r2274739437:

First observation: we don't have to use the same folder where we store models. I'd rather use a separate folder: evals as it's completely separate from our training uploading.

Also, I wonder if we can come up with something more discoverable. I assume model_name can be a version from the models repo or something else. Is decision an ID of the decision task? I think if we want to run evals again for the same model, it would be easier to use a timestamp so that we can later sort and take the latest results.

How about this format: <roo_dir>/<lang-pair>/<dataset_name>/<service>/<architecture_or_api>/<version>/<timestamp>/<evals_files>

For example:
evals/en-zh/flores-test/mozilla/tiny/2.1/20250813112233/*.json
We can also add COMET/BLEU jsons there later.

2.1 can be a version from the models repo or any arbitrary model name if we want to evaluate some dev models.
flores-test can be a dataset name from the model repo, or again any other name if we want to use some external data

evals/en-zh/flores-test/google/api/9.8.2/20250813112233/*.json
evals/en-zh/news123/mozilla/tiny/2.1/20250813112233/*.json
evals/en-zh/flores-test/mozilla/tiny/dev-retraining-experiment/20250813112233/*.json

If we want to run final evals for the newly trained model from the training pipeline we could replace model version with a task ID like
evals/en-zh/flores-test/mozilla/tiny/R0kZGTH6S7Kz11wvksCrEg/20250813112233/*.json

Then we would be able to easily list the available artifacts on any level and see what's there: gsutil ls evals/en-zh/

evgenyrp:
Migrate and improve the final evaluation from [firefox-translations-models.](https://github.com/mozilla/firefox-translations-models)

Example task and artifacts https://firefox-ci-tc.services.mozilla.com/tasks/PE_slRCzSAGA9MRoYQNWUg#artifacts

There will be thousands of such artifacts per task output when we run it for all languages, metrics, translators and datasets. Likely we'll group it all and run in parallel. The structure is based on the need to incrementally check and add specific metrics and translations. Also, we cannot use a proper directory structure because Taskcluster does not support recursive uploading from artifacts, so the separator is "__" instead of "/".

Out of scope for this PR: 
-  #1296 
-  #889
-  #1219 
-  #1295 
-  #1297
-  #1261
- #1298 

Testing of the uploading is blocked by: #1294 

The individual commits are not meant to be reviewed.

### AI usage disclosure

Unlike in the previous PR for the model registry UI, 99% of the code was written by hand here as I needed to express precisely what should be done.

Fixes #1259 
Fixes #1262
Fixes #1222 
Fixes #764 
Fixes #1231

### DIFF
--- a/taskcluster/kinds/upload-final-eval/kind.yml
+++ b/taskcluster/kinds/upload-final-eval/kind.yml
@@ -48,6 +48,7 @@ tasks:
         worker-type: upload-artifacts
         run-on-tasks-for: []
         worker:
+            allow-overwrites: true
             app-name: translations
             # the `development` bucket pushes to the non-production GCS bucket
             # (moz-fx-translations-data--5f91-stage-translations-data)


inference/src/app/translator_cli.cpp
evgenyrp, line 1: This is a copy-paste from the original bergamot-translator repo. I just changed the app's name to `translator-cli`.

pipeline/eval/eval.py
evgenyrp, line 1: Only refactoring here: the metrics were moved to `metrics.py`

pipeline/eval/langs.py
evgenyrp, line 1: This is not great, and some of the mappings were generated with ChatGPT, but I couldn't find a unified way of converting things and looking at what's supported where. All the tools seem to have slightly different lang ID schemes and edge cases.

pipeline/eval/llm.py
evgenyrp, line 1: This was moved to `metrics.py` with only a slight refactoring.

pipeline/eval/metricx.py
evgenyrp, line 1: This was copy-pasted from Google's repo

taskcluster/docker/eval/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+                          libboost-filesystem1.74.0 \
+                          libboost-iostreams1.74.0 \
+                          pigz \
+                          curl \
+                          wget \
+                          libicu-dev \
+    && apt-get clean
+
+COPY final_eval.txt /tmp/
+RUN pip3 install -r /tmp/final_eval.txt
```
evgenyrp, line 21: It's important to install the requirements in the Dockerfile; otherwise, it takes half an hour in the task itself, as there are too many.

evgenyrp:
I know it's a big PR, but it would be hard to split it as the bulk of the code is interrelated, and it's better to be reviewed as a whole. 

I suggest @ZJaume can rereview the meat of it in `pipeline/eval`, @bhearsum looks as usual at `taskcluster/` and @nordzilla looks at `inference/`. Everybody is welcome to look at other parts if you want, especially the docs, so let me know if I missed something or if something is unclear.

On another note, I didn't add new tests as this code doesn't impact production models. If there's a bug, we can just fix it and rerun the evals.

pipeline/eval/final_eval.py
```diff
@@ -0,0 +1,509 @@
...
+            read_path = f"{self.read_path}/{self._get_file_name(path)}"
+            tr_json = requests.get(str(read_path)).json()
+
+        return tr_json
+
+    def _get_file_name(self, name: Path) -> str:
+        return str(name).replace("/", self.SEPARATOR)
+
+
+class Secrets:
```
evgenyrp, line 316: @bhearsum you might want to have a look at how I read and set secrets

taskcluster/configs/eval.yml
```diff
@@ -1,18 +1,52 @@
...
+
+  # Bergamot models (it should match model name on GCS)
+  # For non-English centric pairs like it-de, provide two model IDs for both xx-en and en-xx
+  # Use "models: ["latest"]" to run only for the last updated model per language pair
+  # Leave empty to automatically discover all Bergamot models on GCS
+  models: ["latest"]
+#    - spring-2024_AYqN3ysXRp2EGkEqeaA5Rg
 
 taskcluster:
+  #  # The artifacts will be saved with the following format:
```
evgenyrp: Remove old comment

pipeline/eval/final_eval.py
bhearsum: Seems fine!

taskcluster/kinds/final-eval/kind.yml
```diff
@@ -64,15 +63,16 @@ tasks:
...
-          --model_service {model_service}
-          --model_architecture {model_architecture}
-          --model_name {model_name}
-          --corpus_src {corpus_src}
-          --corpus_trg {corpus_trg}
-          --corpus_ref {corpus_ref}
-          --artifacts $TASK_WORKDIR/artifacts
-          --src {src}
-          --trg {trg}
+          printf '%s\n' '{config}' > config.yml &&
```
bhearsum, line 67: This is a clever way to get the config onto disk - nice!

bhearsum (approved):

pipeline/eval/langs.py
ZJaume: I agree. For now I think it's sufficient. `iso639-lang` from PyPi can help a lot, but the conversions in some edge cases sometimes depend on the use case.

pipeline/eval/metrics.py
```diff
@@ -0,0 +1,694 @@
...
+
+
+class SacrebleuMetric(RegularMetric):
+    name = None
+
+    def __init__(self):
+        self.metric: sacrebleu.metrics.base.Metric = None
+
+    @staticmethod
+    def supports_lang(src_lang: str, trg_lang: str) -> bool:
```
ZJaume, line 64: A slight nuance here is that, even if BLEU as metric supports every language, the tokenization does not. Instead of maybe trying to list all supported languages by the default tokenization, should we instead switch to a SentencePiece based tokenization? sacreBLEU recently merged spBLEU-1K for 1K languages but it's still not released, probably `flores200` is a better choice for now.

ZJaume: Or maybe adding a new spBLEU metric.

pipeline/eval/metrics.py
ZJaume, line 1: I like the design.

pipeline/eval/eval_datasets.py
ZJaume, line 1: Is there a reason not to include NTREX here? If it's because of download, there are a few [copies in HF](https://huggingface.co/datasets/davidstap/NTREX).

(I'd like to have the evaluation on it for Catalan)

pipeline/eval/metrics.py
evgenyrp: Good point, it should use a different tokenizer for Chinese, Japanese etc. I added BLEU more for backward compatibility, as we don't plan to actively look at it, but it's probably worth implementing it properly right away

pipeline/eval/langs.py
evgenyrp: We should look at the whole thing here https://github.com/mozilla/translations/issues/1139 and not just the evals. We need to write a universal converter between all those that can support different cases of training and evaluation.

pipeline/eval/eval_datasets.py
evgenyrp: No reason, it's just the PR is already huge. It's on my list as a follow-up PR along fixing MetricX (see  https://github.com/mozilla/translations/issues/1017). It seems the official dataset is on Github, I tried to use it for zh_hant.

pipeline/eval/metrics.py
evgenyrp: I excluded CJK from BLEU-supported languages and added spBLEU as a separate metric. It's important to keep metrics and their names consistent with WMT24/25 and other papers so that we can compare the results.

inference/src/app/translator_cli.cpp
nordzilla: I removed this when I did the migration, thinking that it wasn't currently needed. 

But if it's needed now, that's great to add it back.

taskcluster/kinds/upload-final-eval/kind.yml
```diff
@@ -48,6 +48,7 @@ tasks:
         worker-type: upload-artifacts
         run-on-tasks-for: []
         worker:
+            allow-overwrites: true
```
evgenyrp, line 51: @bhearsum I added this for the final evals as well because we always overwrite *latest* files

PR #1288, opened 2025-11-18
[Experiment] Chinese Traditional

evgenyrp:
[Training dashboard](https://gregtatum.github.io/taskcluster-tools/src/training/?dashboardName=Chinese+Traditional&taskGroupIds=ISf3leOBQluIepIjSralpA&taskGroupNames=%5B%22Nov+18%22%5D)

### DIFF
--- a/configs/cjk/en-zh-hant-cjk.yml
+++ b/configs/cjk/en-zh-hant-cjk.yml
@@ -102,17 +102,20 @@ datasets:
   - opus_wikimedia/v20230407 #                              302,259 sentences
   - opus_Tanzil/v1 #                                        187,092 sentences
   - opus_TED2013/v1.1 #                                     154,579 sentences
-  - opus_News-Commentary/v16 #                              125,996 sentences
+# filtered to 0
+#  - opus_News-Commentary/v16 #                              125,996 sentences
 # cleaning fails
 #  - opus_bible-uedin/v1 #                                   124,378 sentences
 # reported by Jaume as too noisy
 #  - opus_PHP/v1 #                                            41,706 sentences
-  - opus_infopankki/v1 #                                     29,907 sentences
+# filtered to 0
+#  - opus_infopankki/v1 #                                     29,907 sentences
   - opus_TED2020/v1 #                                        16,382 sentences
   - opus_QED/v2.0a #                                         13,123 sentences
   - opus_NeuLab-TedTalks/v1 #                                 8,076 sentences
   - opus_tldr-pages/v2023-08-29 #                             4,167 sentences
-  - opus_tico-19/v2020-10-28 #                                3,071 sentences
+# filtered to 0
+#  - opus_tico-19/v2020-10-28 #                                3,071 sentences
 # failed to download/unpack
 #  - mtdata_ELRC-hrw_dataset_v1-1-eng-zho #                ~631,760 sentences (71.4 MB)
 #  - mtdata_Microsoft-ntrex-128-eng-zho_CN #       No Content-Length reported (https://github.com/MicrosoftTranslator/NTREX/archive/52b9c57c.tar.gz)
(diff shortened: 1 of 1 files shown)


configs/cjk/zh-hant-en-cjk.yml
```diff
@@ -0,0 +1,205 @@
...
+#  - mtdata_Microsoft-ntrex-128-eng_US-zho_CN #    No Content-Length reported (https://github.com/MicrosoftTranslator/NTREX/archive/52b9c57c.tar.gz)
+#  - mtdata_Microsoft-ntrex-128-eng_US-zho_TW #    No Content-Length reported (https://github.com/MicrosoftTranslator/NTREX/archive/52b9c57c.tar.gz)
+  - mtdata_Statmt-news_commentary_wmt18-13-zho-eng #    ~1,001,393 sentences (113.2 MB)
+# exclude wmt news
+#  - mtdata_Statmt-newsdev_zhen-2017-zho-eng #             ~402,756 sentences (45.5 MB)
+#  - mtdata_Statmt-newsdev_enzh-2017-eng-zho #             ~402,756 sentences (45.5 MB)
+# failed to download/unpack
+#  - mtdata_Statmt-ccaligned-1-eng-zho_CN #              ~5,669,496 sentences (640.7 MB)
+#  - mtdata_Statmt-ccaligned-1-eng-zho_TW #              ~2,407,082 sentences (272.0 MB)
+  - url_https://storage.googleapis.com/releng-translations-dev/data/en-zh_hant/hplt_v1_1.[LANG].zst
```
evgenyrp, line 133: I uploaded opus hplt v1.1 to GCS as it's not available through the API

pipeline/data/mono_importer.py
```diff
@@ -49,8 +49,8 @@ def main(args_list: Optional[list[str]] = None) -> None:
     )
     parser.add_argument("--dataset", type=str, help="The key for the dataset")
     parser.add_argument("--language", type=str, help="The BCP 47 language tag of the dataset")
-    parser.add_argument("--src", type=bool, help="Source language of a language pair")
-    parser.add_argument("--trg", type=bool, help="Target language of a language pair")
+    parser.add_argument("--src", type=str, help="Source language of a language pair")
```
evgenyrp, line 52: This bug caused everything to be always filtered and never converted to one type of Chinese. It was likely the case for Simplified as well but it's not critical as there was enough data for it.

evgenyrp:
Only 24M of parallel corpus examples left ([stats](https://firefoxci.taskcluster-artifacts.net/etyEWLwISdi4q1Fwtq0ZzA/0/public/build/corpus.stats.json), [group](https://firefox-ci-tc.services.mozilla.com/tasks/groups/W2k3K-F7SeOyO-YU8H0zkQ))

When I trained zh-en in Simplified, corpus size was 86M ([stats](https://firefoxci.taskcluster-artifacts.net/fJLrgURyTv-10joieyr0Tw/0/public/build/corpus.stats.json), [group](https://firefox-ci-tc.services.mozilla.com/tasks/groups/P0Wh6gL2SEqAe1JEWQd_bg))

@ZJaume do you have ideas why such a big difference? Can it be the new nllb fasttext model? We currently identify only zh code with it without the script. We do the conversion from Simplified to Traditional before that on data import step. I wonder if the model didn't like the converted texts.

ZJaume:
Ok, so I took a look at parallel and it's a bit difficult to tell what's happening with the cleaning without debugging log for filters. Probably it was in part because of `lid218e` model, let's see what comes out of it after the change to Openlid.

I think that maybe it is a good option to take a step back and start from parallel data that's originally written in Hant, then adding converted data and see if it improves. As far as I can tell, we are only using HPLT as "native" corpus and the rest is converted. There are only a few more in [opus](https://opus.nlpl.eu/results/en&zh-TW/corpus-result-table) but training only on that can take the conversion out of the equation of what's hurting quality. Also, a small but maybe not that small thing is that it seems we are converting everything to Traditional here

https://github.com/mozilla/translations/blob/a3406bc03c8bbcaedaef1ff467667805d942314a/pipeline/data/cjk.py#L65-L71

and there is at least one detection case that we shouldn't convert, which is `hanzidentifier.BOTH`. Then there's the `hanzidentifier.MIXED` case, but converting in that case might be right choice.

PR #1292, opened 2025-11-21
feat: allow overwrites in update db task

bhearsum:
https://github.com/mozilla-releng/scriptworker-scripts/pull/1300 is adding support for the new `allow_overwrites` option in the upload tasks. This change enables it for the `upload-db` task.

Leaving in draft for now, because it depends on https://github.com/mozilla-releng/scriptworker-scripts/pull/1300 being merged and rolled out to production.

### DIFF
--- a/taskcluster/kinds/upload-db/kind.yml
+++ b/taskcluster/kinds/upload-db/kind.yml
@@ -46,6 +46,7 @@ tasks:
         worker-type: upload-artifacts
         run-on-tasks-for: []
         worker:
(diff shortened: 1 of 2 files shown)


bhearsum:
I had to hack in a run of `upload-db` to verify this, but it worked fine in this graph: https://firefox-ci-tc.services.mozilla.com/tasks/groups/BwSbRs1sQ1eg5VjEZn4OQg (and uploads in a couple of other tasks also worked.)

PR #1317, opened 2025-12-09
Add release labels

### LINKED ISSUE #1263: Get info for the released models from GCS
Currently, https://mozilla.github.io/translations/firefox-models/ uses Remote settings to determine what's deployed, and https://github.com/mozilla/firefox-translations-models/tree/main/evaluation is used for evaluations.

It should use Remote settings and GCS. We need to figure out a way to correspond GCS models to the Remote settings ones. @gregtatum proposed to use hashes for that. Then we need a model index by hash on GCS, similar to https://github.com/mozilla/firefox-translations-models/blob/main/models/by-hash.json.
We can regenerate it in utils/model_registry.py

evgenyrp:
- add metadata writing to the export step (it includes model hash)
- I also backfilled metadata files on GCS for the existing models
- update db to store exported model metadata
- obtain the model release status by matching its hash to remote settings and store in db
- update UIs to add release status labels
- update the Firefox models UI to use the DB. Load remote setting hashes as before to see deployment results immediately on this dashboard. The user interface itself didn't change.
- Firefox Models UI now uses v2 remote settings collection (https://firefox.settings.services.mozilla.com/v1/buckets/main/collections/translations-models-v2/records)

Screenshot 2025-12-09 at 5 08 04 PM

Screenshot 2025-12-09 at 5 07 51 PM

## AI usage disclosure

The DB updater and UI parts are fully vibe-coded.

fixes #1263

### DIFF
--- a/site/firefox-models/models.mjs
+++ b/site/firefox-models/models.mjs
@@ -268,20 +268,22 @@ function setupShowAdditionalDetails() {
  */
 function getReleasedModels(models) {
   /** @type {Map<string, ModelRecord>} */
-  const langPairs = new Map();
+  const byKey = new Map();
   for (const model of models) {
-    if (getReleaseChannels(model)?.release) {
+    const channels = getReleaseChannels(model);
+    if (channels?.release) {
       const langPair = model.sourceLanguage + "-" + model.targetLanguage;
-      const existingModel = langPairs.get(langPair);
+      const key = langPair + "|" + (model.filter_expression ?? "");
+      const existingModel = byKey.get(key);
       if (
         !existingModel ||
         versionCompare(model.version, existingModel.version) > 0
       ) {
-        langPairs.set(langPair, model);
+        byKey.set(key, model);
       }
     }
   }
-  return (models = [...langPairs.values()]);
+  return [...byKey.values()];
 }
 
 /**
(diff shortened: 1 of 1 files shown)


pipeline/quantize/save_metadata.py
evgenyrp, line 1: this script was migrated from the models repo

evgenyrp:
We do have 3 UIs now but they all serve slightly different purpose so I don't know if can merge them for now. The Firefox models one shows Remote Settings directly for example. They all load the same db file though, so we can merge them later if we want.

nordzilla (changes requested):
The changes look reasonable overall. 

I don't have a strong opinion about merging the dashboards, as long as they can have different views. 

We frequently point people, even non-technical users, to https://mozilla.github.io/translations/firefox-models/ as the canonical list of models, so it's important to have a view that is simple and straightforward.

---

> RE: Firefox Models UI now uses v2 remote settings collection (https://firefox.settings.services.mozilla.com/v1/buckets/main/collections/translations-models-v2/records)

Thanks for doing this! I had intentions to update that whenever I found a moment. 

---

Unfortunately, it doesn't seem to be working yet. I'm getting an error building it locally:

image

PR #1320, opened 2025-12-12
Final evals bug fix

evgenyrp:
I fixed various small bugs while running a big backfiling of the final evals.

### DIFF
--- a/db/sql.py
+++ b/db/sql.py
@@ -15,6 +15,7 @@ from db.models import (
     Corpus,
     WordAlignedCorpus,
     Export,
+    TASK_GROUP_ID_LENGTH,
 )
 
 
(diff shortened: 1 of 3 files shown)


db/sql.py
```diff
@@ -604,6 +606,36 @@ def update_release_statuses(self, hash_to_status: dict[str, str]):
...
+        Returns the student_exported model's id if found.
+        """
+        if len(model_name) < 23:
+            return None
+
+        # Skip composite model names (contain ---) as they can't be parsed
+        if "---" in model_name:
+            return None
+
+        run_name = model_name[:-23]
```
ZJaume: I'm not aware of what suffix is being removed here. Is it ok to hardcode by lenght?

evgenyrp: This is a Taskcluster group ID + "_" which always has the same length. It's the only way to get the experiment name from a path. I can add a comment to clarify that.

PR #1325, opened 2025-12-20
Fix LLM evals

### LINKED ISSUE #1316: LLM evaluation fails to produce structured output
I've been having issues with both OpenAI 4o and 4o-mini. Sometimes it changes `}` to `]` in JSON, sometimes it adds things like ` ```json `. Quick pass on more prompt engineering doesn't help.

Example 1:
```
			"punctuation": [5, ""}
```

Example 2:

```
```json
{
```

evgenyrp:
- use OpenAI structured output API to fix JSON parsing issues
- switch to modern "responses" API
- parallelize API calls
- fix showing LLM results in the UI

plus

- fix model caching
- add Google Translate languages for reference

I'm running it for everything with the 4o-mini model. It's cheap. If we don't like it, we can rerun with another model.

#fixes #1316

### DIFF
--- a/taskcluster/configs/config.prod.yml
+++ b/taskcluster/configs/config.prod.yml
@@ -83,10 +83,9 @@ experiment:
   # See the reports for distribution of document scores per language:
   # https://github.com/hplt-project/data-analytics-tool/tree/main/reports/mono-2.0
   hplt-min-doc-score:
-    # Noiser data should be fine for distillation.
     mono-src: 7.0
-    # Only retain the best sentences for back translations.
-    mono-trg: 9.0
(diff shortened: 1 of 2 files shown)


ZJaume (approved):
LGTM. The only concern I have about this are the prompts maybe not being enough explanatory in order to have useful output from the LLM (e.g. spot more issues), the en-ru scores you attached seem pretty similar between models. But maybe this a separated task. Sergio already told me that the prompts probably could be improved, so maybe he wants to work on this.

gregtatum:
> Another issue I think is that LLMs should not use references as many of them are really bad and ambiguous

Seems reasonable to only use the internal knowledge of the LLM, especially for our smaller models.

PR #1331, opened 2025-12-24
Run MetricX in a separate process

### LINKED ISSUE #1295: MetricX does not run with the latest transformers
And if we use the old version, it breaks other things in the final evals.

evgenyrp:
- Add a metric wrapper that runs it in a subprocess with the installation of specific Python requirements in a venv. 
- Add an option to continue running even if some metrics fail on some datasets (off by default)
- Use MetricX large (770M parameters T5) with half-precision to speed things up. Running 3B xl is not feasible if we use V100s. I've been running 12B xxl on H100s for LLM evals, but we're probably ok with "large".

fixes #1295


pipeline/eval/metrics.py
```diff
@@ -81,10 +77,148 @@ def score_qe(
...
+        env = os.environ.copy()
+        pythonpath = os.getcwd()
+        if env.get("PYTHONPATH"):
+            pythonpath = f"{pythonpath}:{env['PYTHONPATH']}"
+        env["PYTHONPATH"] = pythonpath
+        self._process = subprocess.Popen(
+            [str(self._venv_path / "bin" / "python"), "-m", "pipeline.eval.metric_worker"],
+            stdin=subprocess.PIPE,
+            stdout=subprocess.PIPE,
+            stderr=None,  # Stream to parent's stderr for real-time logs
```
ZJaume: Should stderr be `subprocess.PIPE` otherwise it's lost?

PR #1349, opened 2026-01-12
Support extended language codes including zh_hant

### LINKED ISSUE #896: Add support for Chinese Traditional

### LINKED ISSUE #1139: Extend supported language codes
A large class of errors in the pipeline on the H1-2025 run is that of language tag matching.

Our config only supports ISO 639-1 language tags with the `src` and `trg` properties. In bash and python we can convert them to ISO 639-3. We don't support scripts such as `Latn`, `Hans`, and `Hant`. There are also language tags that are using an outdated form. In ICU there is a method to normalize naming tags, but we essentially need to do the reverse. What this turns into is a large list of scattered `if` checks for certain locales the require a different format of the language code.

There is a related issue with language tag specificity. For instance `no` is the Norwegian macro language, which `nb` and `nn` are the Bokmål and Nynorsk languages respectively. Training `no-en` is fine to have a macro language on the src side, but we'll need to pick either `nb` or `nn` for things like monocleaner. This would be easier done if we can specify that in the config.

```yaml
src:
  label: no # Used in task labels
  bcp_47: no-Latn # -Latn is optional here, as it can be deduced by likely subtags.
  bcp_47_fallbacks: ["nb", "nn"]
```

We can use the label tag to help disambiguate for CI, since BCP 47 can contain non-label friendly characters.

```yaml
src:
  label: zh_hans
  bcp_47: zh-Hans
```

Here we can use ICU to get the likely subtags as needed, to deduce the `Latn`.

We can also create a more generalized fallback for cases that don't strictly follow the standards, or have some odd historic choices. Those would be used literally when no matches were found. They should be checked as a last resort since they can lead to false positive matches.

```
src:
  label: sq
  bcp_47: sq
  strict_fallbacks: ["als_Latn", "als"]
```

Using BCP 47 will also start to unlock region-based configs, such as `es-MX` vs `es-ES` (Mexican Spanish vs Spain Spanish).

### LINKED ISSUE #1235: Integrate HPLT 3.0
https://hplt-project.org/datasets/v3.0

### LINKED ISSUE #1333: Split zh-hant from zh
Currently, all "zh" models land under the same language code, and the evals on the dashboards are incorrect for zh-hant. Using "zt" for it can be a good workaround to stay in 2-letter language codes until we completely rework our system.

evgenyrp:
- add support for zh_hant
- add preliminary support for other extended language codes (they will still require fixes for specific tools, but now there's a foundation for that)
- upgrade to hplt 3.0 (required for zh_hant)
- refactor language code conversions to encapsulate them in `pipeline.langs` module
- update bicleaner to use ICU in the step (required to use the new `langs` module)
- make sure to use openlid fasttext model for Chinese
- predownload fastext models with `fetch` (we have 3 of them now, and they became too big to download on demand)

I'll update the docs in a follow-up PR. I need to refresh them anyway.

fixes #1139 
fixes #896 
fixes #1333 
fixes #1235

### DIFF
--- a/tests/test_data_importer.py
+++ b/tests/test_data_importer.py
@@ -116,7 +116,6 @@ def data_dir():
 @pytest.mark.parametrize(
     "importer,src_lang,trg_lang,dataset",
     [
-        ("mtdata", "en", "ru", "Neulab-tedtalks_test-1-eng-rus"),
         ("mtdata", "en", "zh", "OPUS-gnome-v1-eng-zho_CN"),
         ("mtdata", "en", "zh_hant", "OPUS-gnome-v1-eng-zho_TW"),
         ("opus", "en", "ru", "ELRC-3075-wikipedia_health_v1"),


evgenyrp:
Successful `zh_hant-en` CI run: https://firefox-ci-tc.services.mozilla.com/tasks/groups/B4pYKA6bSx2DTiSWla_6IA

evgenyrp:
@ZJaume the main pipeline and config generator work, but I'm still testing evals for `zh_hant` code. The PR is big because, to support `zh_hant`, I had to refactor the language codes, and, as expected, it touched many parts of the pipeline. The good news is that most of the mess is now inside the `pipeline.langs` module.

It does not comprehensively support all possible codes for all tools. Rather, it lays the framework for such support. It would still be required to investigate each tool individually for each unusual code and add custom mappings.

Have a careful look, but take your time.

pipeline/clean/tools/clean_mono.py
```diff
@@ -7,36 +7,38 @@
 
 from opuscleaner.filters.clean_common import CHARS
 
+from pipeline.langs.codes import LangCode
 
 MIN_LENGTH = 2  # minimum number of words in a sentence
-MAX_LENGTH = 150  # maximum number of words in a sentence
+MAX_LENGTH = 200  # maximum number of words in a sentence
```
evgenyrp, line 13: I increased it for zh_hant, but I hope it should be fine for other languages too

pipeline/langs/codes.py
```diff
@@ -0,0 +1,375 @@
+"""
+Utilities to convert between common language code schemes.
+
+The universal language code we use in the training config is:
+    <ISO 639-1>[_<ISO-15924 script or BCB-47 region>]
```
evgenyrp: typo

pipeline/data/mono_importer.py
```diff
@@ -107,14 +109,15 @@ def main(args_list: Optional[list[str]] = None) -> None:
         return
 
     url = None
+    # iso6391 = to_iso6391(args.language)
```
evgenyrp: remove

pipeline/clean/tools/clean_mono.py
ZJaume: should be fine

pipeline/langs/maps.py
```diff
@@ -887,3 +782,433 @@
...
+    "zh_hant": "cmn_Hant",
+}
+
+ISO6393_DEFAULTS_REVERSED_MAP = {v: k for k, v in ISO6393_DEFAULTS_MAP.items()}
+
+COMMON_FALLBACKS = {
+    # Fallback to Serbian for Serbo-Croatian group (for example, for metric-x)
+    "bs": ["sr"],
+    "ca_valencia": ["ca"],
+    "cr": ["sr"],
```
ZJaume: `cr` is Cree language, where does this map comes from?

pipeline/langs/maps.py
```diff
@@ -887,3 +782,433 @@
...
+    # WMT24pp has only MX
+    "es": ["es_MX"],
```
ZJaume: Like the `cr`, should this COMMON_FALLBACKS used for real common fallbacks that are also not ambiguous? If another tool comes with default Spanish to a different variant and we put it here, there might be undesired mistakes? Should this be in the WMT24pp map?

evgenyrp: Agreed, this map should include only unambiguous common fallbacks. All weird things should go to specific maps by tool. I'll fix it.

pipeline/langs/maps.py
evgenyrp: I've mistaken it for `hr`.

pipeline/langs/maps.py
evgenyrp: Fixed, I still left Norwegian to fallback between "no" and "nb". I think this will be common for many tools.

evgenyrp:
The CI should be green, but there are some intermittent issues with mtdata dataset downloading and quantization

bhearsum (approved):
Seems fine from a taskgraph standpoint.

PR #1359, opened 2026-01-26
Make it more clear that bergamot-translator.js is generated code

tyleretchart:
Added automated message that bergamot-translator.js is automatically generated. Bugzilla https://bugzilla.mozilla.org/show_bug.cgi?id=1968038


inference/scripts/build-wasm.py
```diff
@@ -98,7 +98,9 @@ def prepare_js_artifact():
     source = (
         "\n".join(
             [
-                "/* This Source Code Form is subject to the terms of the Mozilla Public",
+                "/* NOTE: This file is generated by inference/scripts/build-wasm.py from the translations repo",
+                " * https://github.com/mozilla/translations\n\n",
+                " * This Source Code Form is subject to the terms of the Mozilla Public",
```
nordzilla: > Praise

This is great, thank you!

---

> Suggestion

Many of our current generated-code disclaimers make it explicit that you should not modify this file directly, because it was generated:

* https://searchfox.org/firefox-main/search?q=do%20not%20modify

We should probably add a similar warning to this as well. It looks like we don't have a single unified style, so whatever you feel reads the nicest is fine with me!

tyleretchart: Sounds good! I updated the PR with a message here. LMK if we need any other changes

PR #1370, opened 2026-02-05
Change opuscleaner-mode to 'customs' when generating a config

### LINKED ISSUE #1368: Make custom OpusCleaner mode the default
The config generator uses `opuscleaner-mode: default` as default, but if there are custom configs for a corpus or a language, we usually add them to be there for future trainings. A forgetful person like me, could start a retrain of some languages and think there's nothing to change since those languages were already covered and not change to `opuscleaner-mode: custom`. What do you think @evgenyrp ?

ZJaume:
closes #1368


ZJaume:
I though it could be useful to keep it in case someone need to bypass custom configs for hacking or if a language pair has very sofisticated filters and we just want to train something with just a few filters to compare. What do you think? @evgenyrp

PR #1378, opened 2026-02-12
Refresh docs

### LINKED ISSUE #1299: 404 from docs/data-and-cleaning/datasets.md
https://github.com/mozilla/translations/blob/main/docs/data-and-cleaning/datasets.md
the link "training config" points to
https://github.com/mozilla/translations/tree/main/configs/config.test.yml

### LINKED ISSUE #1365: Review and refresh docs
We've added a lot of stuff recently, so some parts of the documentation might be outdated.

We should do a comprehensive review and actualize anything related to:
- Model registry
- Evaluation
- Training guide
- Data cleaning
- CJK
- Extended language support when we land #1349 
- Generating synthetic data with LLMs and finetuning

evgenyrp:
- move all readmes to /docs to have it all in one place
- introduce infrastructure section
- Refresh docs

To test run `task serve-docs`

fixes #1365 
fixes #1299

### DIFF
--- a/docs/training/using-pretrained-models.md
+++ b/docs/training/using-pretrained-models.md
@@ -9,6 +9,8 @@ nav_order: 9
 Pretrained models are machine learning models trained previously that can be used as the starting point for your training tasks.
 Utilizing pretrained models can reduce training time and resource usage.
 
+See the recommended hyperparameters that depend on the model type in the [Model Training Guide](README.md#5-using-pretrained-models).
+
 ## Configuration Parameters
 
 To download and use models from previous training runs or external sources, use the `continuation.models` parameter in the training config. The keys in this parameter correspond to the training task `kinds` capable of using pretrained models. This is currently `train-teacher-model` and `backtranslations-train-backwards-model`. See [#515](https://github.com/mozilla/translations/issues/515) for `distillation-student-model-train` support.


docs/training/using-pretrained-models.md
ZJaume, line 1: This file is missing the additions I [mentioned](https://github.com/mozilla/translations/issues/1365#issuecomment-3877263340)

docs/infrastructure/tracking.md
```diff
@@ -147,6 +149,18 @@ When running online from Taskcluster, the resources used by the machine will be
 
 ![System charts](../assets/tracking/system_charts.png)
 
+### Configuring dashboards
```
ZJaume, line 152: It's nice to know this, but I do not have "write" permissions on the W&B workspaces, so every time I change something like this, the interface says "This changes are only visible to you" and if I refresh the page they disappear. Same happens when changing the visibility of any of the runs.

docs/training/pipeline-steps.md
ZJaume: Could also mention that we apply deduplication on the target side and takes the best pair bicleaner score when there are two or more target sentence duplicates.

docs/infrastructure/task-cluster.md
ZJaume: When the documentation talks about "group ids" I always get confused which "type" of group id it is. If it's a task group containing the Decision and Train tasks or the task group that contains the pipeline steps. It would be nice to clarify this a little bit more.

docs/infrastructure/tracking.md
evgenyrp: We'll sort it

docs/training/using-pretrained-models.md
evgenyrp: I added a reference to the section in the training config. It is likely a user would need to change those hyperparameters when setting up a training config, so I'd like to keep it there. We do have some other ones about cleaning that we document in other sections, but they have good defaults.

PR #1381, opened 2026-02-20
Decoding by sampling support

ZJaume:
Adding sampling support was just as easy as add support to propagate pipeline config `marian-args` correctly when the parameter is a vector. It turns out that when we expand `marian-args` in `kind.yml` to propagate them, if the marian argument is a vector like output sampling
```yaml
marian-args:
  decoding-backward:
    beam-size: '1'
    mini-batch-words: '5000'
    output-sampling:
      - topk
      - 10
```
the expansion turns into `--output-sampling [topk, 10]`. `translate.py` now handles this so now any marian argument that is a vector can be used. ctranslte2 translator also parses the output sampling argument and now is supported.

I have not set this decoding method as default for backtranslations because it is not clear if it's better or not. For hbs it turned out to be better for teachers but not for students, also differences were small. Maybe we'll see more differences when we start doing more low-resource languages.


pipeline/translate/translate.py
```diff
@@ -174,6 +187,8 @@ def main() -> None:
     if extra_marian_args and extra_marian_args[0] != "--":
         logger.error(" ".join(extra_marian_args))
         raise Exception("Expected the extra marian args to be after a --")
+    if get_beam_size(extra_marian_args) != "1" and "--output-sampling" in extra_marian_args:
+        raise Exception("Output sampling and beam search are incompatible, set beam to 1")
```
evgenyrp: It's usually preferred to raise a more specific exception instead of the base one, like `ValueError`.

PR #1382, opened 2026-02-20
Vocabulary improvements

ZJaume:
This enables split-digits for joint vocabs, which was missing and adds the option to change the normalization rule. Since we haven't trained anything with NFC normalization yet, the default stays the same (NFKC) as before and the user may choose a different one.

HBS branch depends on this one, so I would like to merge it and change the default if we find it successful.


pipeline/train/normalization/nfc.tsv
evgenyrp, line 1: Would it be possible to download those on demand from somewhere?  They are relatively big to store in the repo, and also, I assume they might change over time.

ZJaume: Would you agree on adding a taskcluster fetch for this? It won't autoupdate but anyways if we didn't do this PR, the default NFKC normalization wasn't going to be autoupdated because it's embedded in SentencePiece and we already have an old SP version.

ZJaume: Also the normalization for HBS is custom made for transliteration `nfkc_translit_hbs.tsv`.

ZJaume:
Evgeny requested pulling files from a remote to avoid indexing them in git and now a fetch task pulls them. However, not the custom one as it is ours.

bhearsum (approved):

PR #1387, opened 2026-03-02
Faster Bicleaner AI

ZJaume:
Benchmarks with one GPU on a interactive task (en-bs model and en-xx-large model)
| configuration | speed (rows/s) |
| -- | -- |
| base fp32 batch32 block10k (previous config) | 273 |
| base fp32 batch128 block10k | 320 |
| base fp16 batch128 block10k | 729 |
| base fp16 batch512 block10k | 746 |
| base fp16 batch512 block50k | 768 |
| large fp32 batch32 block10k (previous config) | 126 |
| large fp16 batch512 block50k | 276 |

We can squeeze a bit more speed with bigger batches but I guess we prefer to be on the safer side regarding possible OOM issues. I tested 1024 and 2048 batch sizes and does not crash, so 512 should be pretty safe.

### DIFF
--- a/taskcluster/kinds/corpus-clean-parallel-bicleaner-ai/kind.yml
+++ b/taskcluster/kinds/corpus-clean-parallel-bicleaner-ai/kind.yml
@@ -107,10 +107,7 @@ tasks:
                 - >-
                     pip install -r {bicleaner_reqs} &&
                     export PYTHONPATH=$PYTHONPATH:$VCS_PATH &&
-                    export CUDA_DIR="$MOZ_FETCHES_DIR/cuda-toolkit" &&
-                    export PATH="$PATH:~/.local/bin:$CUDA_DIR/bin" &&
-                    export XLA_FLAGS="--xla_gpu_cuda_data_dir=$CUDA_DIR" &&
-                    export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:$MOZ_FETCHES_DIR/cuda-toolkit/lib64:$MOZ_FETCHES_DIR/cudnn/lib" &&
+                    export PATH="$PATH:~/.local/bin" &&
                     $VCS_PATH/pipeline/bicleaner/bicleaner.sh
                     $MOZ_FETCHES_DIR/{dataset_sanitized}
                     $TASK_WORKDIR/artifacts/{dataset_sanitized}
@@ -121,10 +118,6 @@ tasks:
             "{provider}": corpus-clean-parallel-{provider}-{dataset_sanitized}-{src_locale}-{trg_locale}
             corpus-clean-parallel-fetch-bicleaner-model: corpus-clean-parallel-fetch-bicleaner-model-{src_locale}-{trg_locale}
         fetches:
-            fetch:
-                - cudnn
-            toolchain:
-                - cuda-toolkit
             "{provider}":
                 - artifact: "{dataset_sanitized}.{src_locale}.zst"
                   extract: false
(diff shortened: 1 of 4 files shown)


taskcluster/kinds/corpus-clean-parallel-bicleaner-ai/kind.yml
```diff
@@ -121,10 +118,6 @@ tasks:
             "{provider}": corpus-clean-parallel-{provider}-{dataset_sanitized}-{src_locale}-{trg_locale}
             corpus-clean-parallel-fetch-bicleaner-model: corpus-clean-parallel-fetch-bicleaner-model-{src_locale}-{trg_locale}
         fetches:
-            fetch:
-                - cudnn
-            toolchain:
-                - cuda-toolkit
```
evgenyrp, line 127: I think it is a common optimization to prevent a lengthy installation of CUDA at runtime, right? @bhearsum Does it not work with the new version of Bicleaner? @ZJaume

evgenyrp: A different version of cuda/cudnn can be added in fetch and toolchain

evgenyrp (commented):
Great improvement! 

Let's clarify the CUDA installation. Can we continue using the toolchain?

taskcluster/kinds/corpus-clean-parallel-bicleaner-ai/kind.yml
ZJaume: I was not able to make the new Tensorflow version work with the toolchain that Taskcluster fetches. Tensorflow now it is using the CUDA pip packages like torch has been doing for some time. I'm not sure if I was not able to make it work because Tensorflow was not compatible with our toolchain version or just because it was not looking at the LD_LIBRARY_PATH. The [instructions](https://www.tensorflow.org/install) do not mention anything about optional cuda installation, so maybe the `pip install tensorflow` (without the cuda) just installs cpu version and cannot be linked to existing CUDA installation because it does not have symlinks.

bhearsum: If it does turn out that this causes a CUDA installation, and it uses a different CUDA than we currently have in toolchains, @evgenyrp is correct that we can add another fetch+toolchain task to grab that and use it here.

evgenyrp: @ZJaume AI tells me it might be possible, but I know nothing about this specific version of Tensorflow. I guess let's double-check that there's no way to install it the old way by adding another version to the toolchain. If not, let's just do the pip install. It will impact CI costs (not sure how much). In prod Bicleaner takes days anyway, so it doesn't make a difference.

ZJaume: I've been trying to make it work with LD_LIBRARY_PATH with the same CUDA version that tensorflow pip installs and no success. I think the tensorflow package without cuda lacks the symlinks. TBH, I would prefer to do the hermetic install with `[and-cuda]` that should always work and it is much easier to upgrade tensorflow. If there is concern about lengthy installations, using `uv` can be a solution. I tried this in an interactive task
```
pip install uv
uv pip install --system -r checkouts/vcs/pipeline/bicleaner/requirements/bicleaner-ai.txt
```
and it took 1m10s. For comparison, a previous pip install for this task without including nvidia packages, was already taking 5m.

ZJaume: PS: we could also use uv for every task to save a substantial time in the CI. It's much faster downloading and writing install files to disk and it's quite stable.

evgenyrp: Ok, let's just do the pip install for now and file a ticket for switching to uv with "optimization" label.

PR #1394, opened 2026-03-17
Use uv as installer

ZJaume:
Replacing pip by uv as installer for the tasks. These are the wall clock times for some of the tasks that i found significant gains:

| task | pip | uv |
| ------ | --- | --- |
| corpus-clean-parallel-bicleaner-ai-opus-ELRC-3075-wikipedia_health_v1-ru-en | 5m 44s | 3m 59s |
| corpus-clean-parallel-opus-ELRC-3075-wikipedia_health_v1-ru-en | 3m 42s | 2m 39s |
| corpus-clean-mono-news-crawl-en-news_2007-mono-trg | 3m 51s | 2m 30s |
| dataset-opus-ELRC-3075-wikipedia_health_v1-ru-en | 8m 9s | 4m 41s |
| evaluate-teacher-ensemble-flores-devtest-ru-en | 20m 22s | 9m 8s |
| evaluate-teacher-flores-devtest-ru-en-1 | 14m | 10m 51s |
| test-lint-black | 1m9s | 35s |
| backtranslations-train-backwards-model | 7m 36s | 6m 13s |

for the rest of the task kinds there are not much gains, but it is quite noticeable on `corpus-clean-*`, `dataset-*` and `evaluate-*` which we have many, so I guess it's worth the change. I will squash it before merging so we can easily revert it if something goes wrong.

### DIFF
--- a/taskcluster/docker/base/Dockerfile
+++ b/taskcluster/docker/base/Dockerfile
@@ -43,6 +43,8 @@ RUN apt-get update -qq \
     # Clean up the apt cache to reduce image size.
     && apt-get clean
 
+COPY --from=ghcr.io/astral-sh/uv:0.10.9 /uv /uvx /bin/
+
 RUN locale-gen "$LANG"
 
 RUN pip install zstandard
--- a/taskcluster/docker/base/Dockerfile
+++ b/taskcluster/docker/base/Dockerfile
@@ -57,9 +57,11 @@ ENV IS_DOCKER=1
 ENV SHELL=/bin/bash \
     HOME=/builds/worker \
     PATH="/builds/worker/.local/bin:$PATH" \
-    TERM=hterm-256color
+    TERM=hterm-256color \
+    UV_CACHE_DIR=/builds/worker/.task-cache/uv
 
 VOLUME /builds/worker/checkouts
 VOLUME /builds/worker/.task-cache/pip
+VOLUME /builds/worker/.task-cache/uv
 
 USER root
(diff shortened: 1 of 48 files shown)


taskcluster/docker/base/Dockerfile
```diff
@@ -55,9 +62,11 @@ ENV IS_DOCKER=1
 ENV SHELL=/bin/bash \
     HOME=/builds/worker \
     PATH="/builds/worker/.local/bin:$PATH" \
-    TERM=hterm-256color
+    TERM=hterm-256color \
+    UV_CACHE_DIR=/builds/worker/.task-cache/uv
```
bhearsum, line 66: In addition to this, you should update the `use-caches` entries in the kinds to add `uv`. Otherwise the caches won't be shared between tasks (this directory will be empty whenever a new task starts).

bhearsum (changes requested):
Seems like a no brainer to me for the most part!

evgenyrp:
We also do the installation of packages for MetricX separately [here](https://github.com/mozilla/translations/blob/7eb971749e0cd3b87322f6b2dab423e40b69d5af/pipeline/eval/metrics.py#L113) and it usually takes a while. It's worth switching to uv as well.

taskcluster/docker/base/Dockerfile
ZJaume: done

ZJaume:
> We also do the installation of packages for MetricX separately [here](https://github.com/mozilla/translations/blob/7eb971749e0cd3b87322f6b2dab423e40b69d5af/pipeline/eval/metrics.py#L113) and it usually takes a while. It's worth switching to uv as well.

This is also now managed by uv as well. I tried running locally for one language and it seems to work.

bhearsum (approved):

PR #1399, opened 2026-03-24
Add NTREX dataset downloader support

### LINKED ISSUE #889: Consider adding NTREX-128 for evaluation

ZJaume:
Since the latest trainings have shown teacher models training less than they should, I thought NTREX as an additional validation could be a useful resource to achieve longer trainings in the upcoming languages.

closes #889


ZJaume:
This change adds part of the dataset to our git tree and also as public in taskcluster, but it is relatevely old and it was already indexable from their repo, so I guess it's already leaked. However it is CC-BY-SA, so I guess we have to add the attribution and license somewhere, but I do not know where.

bhearsum (approved):

PR #1422, opened 2026-05-13
Add --file support to utils/run_model.py

### LINKED ISSUE #523: Add more sources support for utils/run_model.py
It would be nice to have:

* File path `--file`
* URL `--url`
* Remote settings collection `--remote-settings en-ca --version 1.0a1`

For remote settings we would need to use the fork of Marian to support the quantization I believe.

Geetanjali-Kulkarni:
## Summary

Adds support for translating input from a text file using a new `--file` argument in `utils/run_model.py`.

## Changes

* Added `--file` argparse option
* Updated `translate_over_websocket()` to process lines from a file
* Preserved existing interactive translation behavior

## Example

```bash
task run-model -- --file input.txt
```

fixes #523

### DIFF
--- a/utils/run_model.py
+++ b/utils/run_model.py
@@ -139,22 +139,42 @@ def connect_to_ws(port: int) -> WebSocket:
     return ws
 
 
-def translate_over_websocket(port: int):
+def translate_over_websocket(port: int, input_file: Optional[str] = None):
     """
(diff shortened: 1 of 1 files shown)


evgenyrp:
Let's discuss in #523 if this is what was actually meant by adding --file.

Also you can run linter with `task lint-fix`. See https://mozilla.github.io/translations/docs/contributing/development/

PR #1437, opened 2026-06-25
Trying to build and export embedding from translation model using Marian

tliumozilla:
current state; PR for easier diff inspection


gregtatum:
Can you provide more context into this work? It's unclear if it's meant to share context into work as a draft, or if you are intending to land and ship this code.

If the former you'll want to put this in as a draft.

If the latter, you'll need to work with either me or @nordzilla on a plan to ship this to Firefox, which can be a bit involved as it involves coordinating work in this repo and in Firefox. It would be good to explain this work more if you're actually wanting to ship it.

PR #1449, opened 2026-08-04
IndicTrans2 as a teacher

ZJaume:
Opening this just to show the progress. Bear in mind that this branch also has commits from dev_next26. Only the last 3 are for IndicTrans2


ZJaume:
Closed in favor of #1460

PR #1460, opened 2026-09-03
IndicTrans2 integration as teacher to distil from

ZJaume:
This adds support for continuation from a pretrained teacher model to generate distillation data and train student. Then adds support for external pretrained model IndicTrans2.


ZJaume:
This is now ready to review. The commit history is a bit of a mess, since I was not able to test locally a lot of things, but I guess it is okay because it only adds one (or maybe two, continuation and a new decoder) feature.

How it has been tested:
 - a new test case in test_continuation.py (however it only checks the graphs
 - a new test case in test_ctranslate2.py to test IndicTrans2 decoder, which is disabled because it needs HF credentials to download the model, but it passes locally.
 - a test [training](https://firefox-ci-tc.services.mozilla.com/tasks/groups/fLN_k3DoQB-Sawgti17JXg) for ur-en using 100k sentences runs successfully skipping teacher training and using IndicTrans to generate distillation data.
 
The implementation of the taskcluster continuation logic has been done with a lot of trial and error and help of LLMs, my understanding of how the whole thing works is partial, so maybe there are better ways to do it.

pipeline/translate/translate_ctranslate2.py
```diff
@@ -114,92 +116,202 @@ def parse_sampling(self):
...
+        # Check if we are inside a taskcluster task
+        import os
+
+        if os.environ.get("TASK_ID"):
+            from pipeline.common.secrets import Secrets
+
+            secrets = Secrets()
+            secrets.prepare_key_hf()
+
+        self.maxi_batch_size = 10000
```
evgenyrp: Does this setting depend on a GPU type? If yes, it should be in the config ideally (don't we already have it for marian decoding? we could reuse)

pipeline/translate/translate_ctranslate2.py
```diff
@@ -114,92 +116,202 @@ def parse_sampling(self):
...
+        self.beam_size = beam_size
+        self.model = IndicTrans2Inference(
+            src_locale,
+            trg_locale,
+            device=device if device == "cpu" else "cuda",
+            device_index=device_index if device == "gpu" else 0,
```
evgenyrp, line 240: Did you check that it utilizes all 4 GPUs while running inference with this code?

docs/training/using-pretrained-models.md
```diff
@@ -20,9 +20,8 @@ continuation:
   models:
     # Continue training a teacher model.
     teacher:
-      urls:
-        # Replace the following {task_id} with the "train-teacher-model" task id.
-        - https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/{task_id}/artifacts/public/build
+      # Replace the following {task_id} with the "train-teacher-model" task id.
+      url: https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/{task_id}/artifacts/public/build
```
evgenyrp, line 24: Let's make sure we have 'url' instead of 'urls' everywhere, like in configs and in the model registry templates: https://mozilla.github.io/translations/model-registry/?searchString=&showModels=true&modelName=next26-beam4&modelLangpair=en-be&modelModelName=teacher_2&metric=comet

evgenyrp (commented):
I left a few questions. I think one of the main directions to fix in the future is our continuation mechanism and removing the legacy code. Ideally, it should be one path for everything that we can easily test and understand.

See all the related issues:

https://github.com/mozilla/translations/issues/1151
https://github.com/mozilla/translations/issues/1289
https://github.com/mozilla/translations/issues/1214
https://github.com/mozilla/translations/pull/1218/changes

If we work on integrating LLM-based distillation/finetuning later on, we should straighten all this out.

bhearsum (approved):

pipeline/translate/translate_ctranslate2.py
ZJaume: Yes it does.

pipeline/translate/translate_ctranslate2.py
ZJaume: Well, it kind of depends yes, but it was here mostly because it was an specific option to IndicTrans2 for two reasons. Ctranslate2 uses a real translate_iterable so maxi_batch is not used and maxi_batch here means a different thing, for marian it is the "number of batches to be preloaded" here it is "the number of sentences to be preloaded". However you are right and we should be able to configure it, so I've just moved it to decoder config and now it can be parametrized in training config.

docs/training/using-pretrained-models.md
ZJaume: Although it is not 100% clear to me which parts of the code are part of the new continuation logic and which parts are part of the old one, the registry example configs provide `continuation: train-teacher:` examples which is depending on `taskcluster/translations_taskgraph/transforms/training_continuation.py:get_model_mounts()` function that is part of the old logic, so I think the examples in the model-registry should be addressed when we fully migrate to the new continuation logic. Currently indictrans is accessed with `continuation: teacher:` entries, which are separated.

evgenyrp (approved):

PR #1464, opened 2026-09-10
Add Uyghur to the supported langauges

evgenyrp:


pipeline/langs/all_autogenerated.json
evgenyrp, line 1: I think this was not regenerated after some recent changes in datasets

ZJaume: true, and probably will need to be generated again when I merge my next26 branch

ZJaume (changes requested):
Uygur has Bouquet support, but it is not in the evals. Could you add it to the list of Bouquet languages and run evals for it?

Also, a couple of dataset-newscrawl tasks failed, I'm not sure if it's temporary. Last week there was a change in their website and I had to change the html regex parser in the config generator, but this failure does not seem related to the html page. I cannot rerun them manually (lack of permissions)

ZJaume (approved):

